PR #276, opened 2013-12-19
Allow optional :all parameter for RDoc::Options#visibility= as synonym for :private

zzak:
This feature moves RDoc::Options#visibility= out to a separate method
and takes a synonymous argument of :all for :private.

Since :private is not an obvious name for documenting all methods of
this visibility, we provide this optional synonym. Also, this is similar
to how we can pass "--all" when parsing command-line options.

### DIFF
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -325,7 +325,7 @@ class RDoc::Options
   # other visibilities may be overridden on a per-method basis with the :doc:
   # directive.
 
-  attr_accessor :visibility
+  attr_reader :visibility
 
   def initialize # :nodoc:
     init_ivars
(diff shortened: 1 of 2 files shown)


windwiny:
@zzak   @drbrain 

Why not set visibility default to private? I think this is better than protected,

Most private method is important for end-users , like `Module#extended`, `Module#included`, `Class#inherited`, etc..
if a method only used inner function, it may add `':nodoc:'`

http://ruby-doc.org/ website doc show private methods,
http://rubyinstaller.org/downloads/ chm files not show private methods.

zzak:
@windwiny I'm not sure the best default, this feature was only designed to implement the option parameter, not decide defaults.

In any case, it should be easy for the developer to decide which kind of visibility they want in their generated docs. We can't possibly predict what everyone wants for visibility IMO

PR #353, opened 2015-06-18
Added position fixed to the nav bar

Hoijof:
The webpage is unusable if every time you scroll down you lose the nav bar. Just fixed it to its position and fixed also the footer to the bottom left of the page.


drbrain:
This makes it impossible to scroll to the bottom of the `nav` element if it is longer than the screen is tall.

It also causes the validator badges to overlap the content if the content.

rdoc pull 353

I don't think it will work correctly on mobile devices, either.

PR #369, opened 2015-09-08
Only display post-install message when necessary

amcaplan:
Considering that most installs are on modern Rubies (i.e., 1.9.3+), may as well not confuse people with a post-install message unless it's actually necessary.

This arose out of personally seeing this post-install message frequently for the past year despite the fact that I always use 2.0 and above.


drbrain:
This won't work as that code is executed at specification build time, not at install time.

RubyGems doesn't support this feature so I must reject this issue.

amcaplan:
Ha, shows how much I know about RubyGems.  Thanks for the quick response!

PR #371, opened 2015-11-10
minor modification to the data can be generated from other scopes

rplaurindo:
Without this modification as "monkey patching" you can not generate, through a "generator" of a "gem" personnel, "rdoc" in the app that install this "gem".

### DIFF
--- a/lib/rdoc/task.rb
+++ b/lib/rdoc/task.rb
@@ -240,13 +240,13 @@ class RDoc::Task < Rake::TaskLib
     ].flatten.compact
 
     task rdoc_task_name => [rdoc_target]
+    args = option_list + @rdoc_files
     file rdoc_target => rdoc_target_deps do
       @before_running_rdoc.call if @before_running_rdoc
-      args = option_list + @rdoc_files
 
       $stderr.puts "rdoc #{args.join ' '}" if Rake.application.options.trace
-      RDoc::RDoc.new.document args
     end
+    RDoc::RDoc.new.document args
 
     self
   end


lib/rdoc/task.rb
```diff
@@ -240,13 +240,13 @@ def define
...
     task rdoc_task_name => [rdoc_target]
+    args = option_list + @rdoc_files
     file rdoc_target => rdoc_target_deps do
       @before_running_rdoc.call if @before_running_rdoc
-      args = option_list + @rdoc_files
 
       $stderr.puts "rdoc #{args.join ' '}" if Rake.application.options.trace
-      RDoc::RDoc.new.document args
     end
+    RDoc::RDoc.new.document args
```
rplaurindo: This line could not be performed when it is inside the block passed as parameter to the "file" method. If really there is need of stay inside the "file" method, then there is something wrong in this method.

lib/rdoc/task.rb
```diff
@@ -240,13 +240,13 @@ def define
     ].flatten.compact
 
     task rdoc_task_name => [rdoc_target]
+    args = option_list + @rdoc_files
     file rdoc_target => rdoc_target_deps do
       @before_running_rdoc.call if @before_running_rdoc
-      args = option_list + @rdoc_files
```
rplaurindo: This line could not be performed when it is inside the block passed as parameter to the "file" method. If really there is need of stay inside the "file" method, then there is something wrong in this method.

lib/rdoc/task.rb
```diff
@@ -240,13 +240,13 @@ def define
...
 
       $stderr.puts "rdoc #{args.join ' '}" if Rake.application.options.trace
-      RDoc::RDoc.new.document args
```
rplaurindo: This line could not be performed when it is inside the block passed as parameter to the "file" method. If really there is necessity of staying inside the "file" method, then there is something wrong in this method.

PR #396, opened 2016-03-15
Remove hoe from development

zzak:
I want to remove hoe to make it easier for testing plugins and third-party extensions against master.

Removing hoe will simplify the Rakefile and reduced the number of dependencies we have to maintain.

This also allows testing extensions and plugins easier, but also contributors will have a much more familiar workflow to contribute; which IMO is the biggest benefit and worth the complexity of moving parser generation onto our Rakefile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 /lib/rdoc/markdown.rb
 /pkg
 /tmp
+Gemfile.lock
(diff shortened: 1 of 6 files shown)


hsbt:
@zzak I switched bundler's task from hoe on rake repository. see https://github.com/ruby/rake/pull/122

PR #397, opened 2016-03-15
Performance improvement

mame:
This PR reduces MRI's `make` time from 70 sec. to 45 sec. on my machine.


hsbt:
I confirmed to reduce elapsed time from 78 sec to 57 sec with my env.

hsbt:
Thank you!

zzak:
Sorry for no reply! I meant to comment sooner.

Thank you for understanding and patch mame-san!!

PR #407, opened 2016-05-21
Fix travis failure

hsbt:
`rake travis` asserts generated files  to list of Manifest.txt. It always fails. I skipped these asserts and invoke only tests.


hsbt:
It works.

zzak:
Yay! Thank you for getting the build green <3 <3

zzak:
As an aside, I also think it's better that CI executes the same tests/environment as a developer would.

:+1: to removing hoe-travis

PR #412, opened 2016-07-09
Removed strict dependency for json, it break to install ruby trunk(2.4)

hsbt:
Ruby 2.4 can not build json-1.8.3. because it uses `rb_cFixnum`. this pull request allows to install json 2.0 later with rdoc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ notifications:
   email:
   - mail@zzak.io
 rvm:
-  - 1.8.7
   - 1.9.3
   - 2.0.0
   - 2.1.10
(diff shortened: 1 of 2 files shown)


hsbt:
json 2.0.0 later does not allow to install `> Ruby 2.0.0` . We should remove json dependency and drop to support Ruby 1.8. I think.

zzak:
@hsbt Got it, sounds good to me!

hsbt:
I update this for rdoc 5.0.0

zzak:
@hsbt I think this needs a rebase since I merged #396

hsbt:
done.

hsbt:
I created another pull request. see https://github.com/rdoc/rdoc/pull/419

PR #416, opened 2016-08-04
Fix multiple comments rendering

valich:
In classes and modules there is some strange code which leads to problems when comments are added from different places. This leads to some docs like

```
# #<RDoc::Comment:0x007fba8317a5a8>
# ---
# #<RDoc::Comment:0x007fba83a67968>
# ---
# #<RDoc::Comment:0x007fba84106120>
```

In these commits I address this issue as well as the cosmetics problem when empty comments are appended to the same class/module.

I am not sure whether I have to file an issue in the tracker or a couple of tests in the repo will suffice.

### DIFF
--- a/test/test_rdoc_class_module.rb
+++ b/test/test_rdoc_class_module.rb
@@ -52,6 +52,24 @@ class TestRDocClassModule < XrefTestCase
     assert_equal "comment 1\n---\ncomment 2\n---\n* comment 3", cm.comment
   end
 
+  def test_add_empty_comment_elsewhere
+    tl1 = @store.add_file 'one.rb'
+    tl2 = @store.add_file 'two.rb'
+    tl3 = @store.add_file 'three.rb'
+
+    cm = RDoc::ClassModule.new 'Klass'
+
+    cm.add_comment comment('comment 1'), tl1
+    assert_equal 'comment 1', cm.comment.text
+
+    cm.add_comment comment(''), tl2
+    assert cm.comment.is_a? RDoc::Comment
+    assert_equal 'comment 1', cm.comment.text
+
+    cm.add_comment comment("* comment 3"), tl3
+    assert_equal "comment 1\n---\n* comment 3", cm.comment
+  end
+
   def test_add_comment_duplicate
     tl1 = @store.add_file 'one.rb'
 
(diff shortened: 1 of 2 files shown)


hsbt:
@valich Can you show me the target code?

I couldn't imagine real situation for this pull request.

test/test_rdoc_class_module.rb
```diff
@@ -35,6 +35,41 @@ def test_add_comment_comment
...
+    tl2 = @store.add_file 'two.rb'
+    tl3 = @store.add_file 'three.rb'
+
+    cm = RDoc::ClassModule.new 'Klass'
+
+    cm.add_comment comment('comment 1'), tl1
+    assert_equal 'comment 1', cm.comment.text
+
+    cm.add_comment comment(''), tl2
+    assert cm.comment.is_a? RDoc::Comment
```
aycabta: Please use below:
```ruby
assert_kind_of RDoc::Comment, cm.comment
```

aycabta:
I understand that documents are collected up when different files have documents for the same class/module, like:

```ruby
# foo.rb

# comment 1
class Klass; end
```

```ruby
# bar.rb

# comment 2
class Klass; end
```

In this case, the documents for `Klass` becomes like below for HTML and `ri`:

```
comment 1

comment 2
```

But it's not like:

```
#<RDoc::Comment:0x0000000000000>
---
#<RDoc::Comment:0x0000000000000>
```

So I couldn't imagine real situation too. Please show source code to reproduce what you assume to become strange documents.

aycabta:
OK, I understand what did you want to do...

The `rdoc` command generates HTML for classes from [`RDoc::ClassModule#comment_location`](https://github.com/ruby/rdoc/blob/7255613233084122ef07aba8251bdadb0290dabc/lib/rdoc/class_module.rb#L117), and the `ri` command uses `RDoc::ClassModule#comment` as `Array` of `RDoc::Markup::Document` from marshaled comments as binary data, and [`test_scan_duplicate_module`](https://github.com/ruby/rdoc/commit/e033904abe51ece5f7857b3767054bbd08f65eaa) uses `comment_location` for test.

So, I think `RDoc::ClassModule#comment` as `String` isn't used, and this changes makes no sense for RDoc.

hsbt:
We will close this.

@valich If you have any issue, Please comment with your usecase scenario.

Issue #354, opened 2015-06-28
errors with <visibility> def ... (ruby >= 2.1)


thyresias:
With the code below:
* method X::Y#foo is not documented
* class X::Z is documented as Z, with superclass Y (instead of X::Z < X::Y)
```ruby
module X
  class Y
    protected def foo
    end
  end
  class Z < Y
    def foo
    end
  end
end
```

zzak:
Going to close this in favor of #355, since I might be able to get @senny to help out with this one ;)

PR #435, opened 2016-12-28
Parse ruby 2.1 `<visibility> def` method definition

### LINKED ISSUE #355: <visibility> def hides following methods
xref #354

Just noticed that [this `private def` definition](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/relation/query_methods.rb#L607) removed all following methods from the docs.

amatsuda:
This patch adds an ability to parse ruby 2.1 style `private def foo()` method definition.

I suppose test/test_rdoc_context.rb is not the right place to put a test code for this, and also test/xref_data.rb may not be the right place to put the test data.
I'm willing to fix them if anyone can tell me where to put them.

fixes #355 #354

### DIFF
--- a/lib/rdoc/context.rb
+++ b/lib/rdoc/context.rb
@@ -99,6 +99,11 @@ class RDoc::Context < RDoc::CodeObject
   attr_accessor :visibility
 
   ##
+  # Current visibility of this line
+
+  attr_writer :current_line_visibility
+
+  ##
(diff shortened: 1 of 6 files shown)


hsbt:
I felt this tests is correctly. but I didn't know right place to put them too.

zzak:
:+1: Thank you @amatsuda and @hsbt!

PR #448, opened 2017-04-18
Improve ri --help and --version options handling

stomar:
* Fix `ri -h`, which **does not work currently**.

  Instead of showing the help banner, ri apparently ignores the `-h' option (without error message for an unknown option) and starts in interactive mode.
  (Also reported at https://bugs.ruby-lang.org/issues/10838.)

* Make `--help` and `--version` options appear in the --help output

* Various small improvements of the `ri --help` output


drbrain:
My name is Eric Hodel and I approve this pull request (pending CI).

PR #449, opened 2017-04-19
Remove unused use_cache default option for ri

stomar:
A tiny fix,
the corresponding ri --[no-]use-cache option was removed
long ago (Nov 2009), with 390a0707273528c5ae6c38b0b21ea081df06d27b.


hsbt:
👍

PR #451, opened 2017-04-27
Prevent negative IDs in output of #inspect.

voxik:
This prevents output of negative object IDs:

https://bugs.ruby-lang.org/issues/13397

I have not tested this closely, but:

1) I hope the test would catch the possible issues.
2) It should not by user visible change ...


hsbt:
We are moving to ripper from ruby_lex. Can you reconfirm this issue and send another patch?

PR #458, opened 2017-07-09
Add test for require dynamic string

aycabta:
The require dynamic string like below shouldn't be parsed:

``` ruby
prefix = 'path'
require "#{prefix}/a_library"
```

### DIFF
--- a/test/test_rdoc_parser_ruby.rb
+++ b/test/test_rdoc_parser_ruby.rb
@@ -2390,6 +2390,20 @@ end
     assert_equal :private, date_time_now.visibility, date_time_now.full_name
   end
 
+  def test_parse_require_dynamic_string
+    content = <<-RUBY
+prefix = 'path'
+require "\#{prefix}/a_library"
+require 'test'
+RUBY
+
+    util_parser content
+
+    @parser.parse_statements @top_level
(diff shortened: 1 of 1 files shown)


hsbt:
I wonder why it is not parsed.

But it's existence behavior. I'm ok to merge this.

aycabta:
I think the dyanmic string expansion with variable is heavy to parse for RDoc, because RDoc just parses Ruby code but doesn't execute. I guess what this is the reason of the behavior.

Memo: [In the first RDoc commit of Ruby trunk at revision 5073](https://bugs.ruby-lang.org/projects/ruby-trunk/repository/diff?utf8=%E2%9C%93&rev=5073&rev_to=5072), it [uses `eval` for string in arguments](https://bugs.ruby-lang.org/projects/ruby-trunk/repository/revisions/5073/entry/lib/rdoc/parsers/parse_rb.rb#L2486). And [at `eb53cb6` of Git repository](https://github.com/rdoc/rdoc/commit/eb53cb6d28216fa58ddc344a976482d8a8ed1eae#diff-865b3bc60914a21343df7207e2083f74R1377), it skips dynamic string in the same method.

PR #462, opened 2017-07-17
Move collecting up tokens for a symbol to RubyLex

aycabta:
`RDoc::Markup::ToHtml` uses `RDoc::RubyLex.tokenize` directly and calls `RDoc::TokenStream.to_html` with returned tokens.

On the one hand, `RDoc::Parser::RubyTools#get_tk` collects up `TkSYMBEG` and trailing tokens inside the symbol with tokens by `RDoc::RubyLex`.

I moved logic what collects up tokens for a symbol from `RDoc::Parser::RubyTools#get_tk` to `RDoc::RubyLex`. So `RDoc::TokenStream.to_html` becomes to handle symbols as value.

Before, each symbols are separeted `TkSYMBEG` and trailing tokens:

separeted symbol

``` html
<pre class="ruby">
<span class="ruby-identifier">a</span> = :<span class="ruby-identifier">hoge</span>
</pre>
```

I think what `:` and `hoge` with `ruby-identifier` attribute are separeted is not correct because symbol is one literal and one value. It's not identifier.

After, each symbols are treated as literal with `ruby-value` attribute.

collected up symbol

``` html
<pre class="ruby">
<span class="ruby-identifier">a</span> = <span class="ruby-value">:hoge</span>
</pre>
```

### DIFF
--- a/lib/rdoc/parser/ruby_tools.rb
+++ b/lib/rdoc/parser/ruby_tools.rb
@@ -33,26 +33,6 @@ module RDoc::Parser::RubyTools
 
     tk = nil if TkEND_OF_SCRIPT === tk
 
-    if TkSYMBEG === tk then
-      set_token_position tk.line_no, tk.char_no
-
-      case tk1 = get_tk
-      when TkId, TkOp, TkSTRING, TkDSTRING, TkSTAR, TkAMPER then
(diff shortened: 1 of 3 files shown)


hsbt:
@zzak Are you done to review this pull-request?

We should review to change code without test code before merge this.

PR #464, opened 2017-07-21
Add hash rocket '=>' token

aycabta:
You can check this feature as sample file below:

```ruby
# sample code:
#
#     { :a => :b }
#
def meth
end
```

The original RDoc generates:

original hash rocket highlighting

```html
<pre class="ruby">
{ <span class="ruby-value">:a</span> =<span class="ruby-operator">&gt;</span> <span class="ruby-value">:b</span> }
</pre>
```

This separated between `=` and `>` as different token types.

RDoc with this Pull Request generates:

with this Pull Request highlighting

```html
<pre class="ruby">
{ <span class="ruby-value">:a</span> <span class="ruby-operator">=&gt;</span> <span class="ruby-value">:b</span> }
</pre>
```

This provides `=>` as one token.

### DIFF
--- a/lib/rdoc/ruby_lex.rb
+++ b/lib/rdoc/ruby_lex.rb
@@ -483,7 +483,7 @@ class RDoc::RubyLex
                   "=", "==", "===",
                   "=~", "<=>",
                   "<", "<=",
-                  ">", ">=", ">>") do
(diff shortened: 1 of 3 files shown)


hsbt:
It seems reasonable.

PR #465, opened 2017-07-21
Fix seek position

aycabta:
When `TkSYMBOL` generated, the seek position moves to incorrect position. This Pull Request fixes it.


hsbt:
I understood to need this. Thanks.

PR #466, opened 2017-07-23
Handle "[]" and "[]=" after "." as method

aycabta:
`[]` after `.` is handled as operator. But, `[]` inside `Array.[]` is method because it is after `.`. This Pull Request fixes it.


hsbt:
I confirmed expected the behavior of tests.

PR #467, opened 2017-07-23
Handle JSON-style hash key as symbol

aycabta:
RDoc handles the JSON-style hash key as identifier and operator `:`, like below:

separated identifer and :

```html
<pre class="ruby">
{ <span class="ruby-identifier">hiee</span><span class="ruby-operator">:</span> <span class="ruby-value">:aaaa</span> }
{ <span class="ruby-identifier">class</span><span class="ruby-operator">:</span> <span class="ruby-value">:aaaa</span> }
</pre>
```

I think the hash key is a symbol. This Pull Request fixes it like below:

JSON-style hash key is a symbol

```html
<pre class="ruby">
{ <span class="ruby-value">hiee:</span> <span class="ruby-value">:aaaa</span> }
{ <span class="ruby-value">class:</span> <span class="ruby-value">:aaaa</span> }
</pre>
```

### DIFF
--- a/lib/rdoc/ruby_lex.rb
+++ b/lib/rdoc/ruby_lex.rb
@@ -948,7 +948,7 @@ class RDoc::RubyLex
                   @indent_stack.push token_c
                 end
               else
-                if peek(0) == ':'
+                if peek(0) == ':' and !peek_match?(/^::/)
                   token.concat getc
                   token_c = TkSYMBOL
                 else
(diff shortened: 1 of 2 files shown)


hsbt:
I understand this pull request and expected the behavior.

Original test of ```{ class:"foo" }``` tokenize was strange expectation for me. Your fixed expectation is reasonable.

PR #468, opened 2017-07-24
Handle TkDREGEXP as regexp in HTML

aycabta:
Dynamic regexp token is not highlighted for HTML. The `TkDREGEXP` token is put in above `TkNode` in case-when statement, because `TkDREGEXP` is a kind of `TkNode`.


hsbt:
👍

PR #469, opened 2017-07-24
Handle "#{}" as dynamic string

aycabta:
Before this Pull Request, RDoc detects dynamic string with only `#` inside string literal. So `#` is handled as dynamic string. This Pull Request fixes the behavior. When `#{` is contained in a string, it is handled as dynamic string.


aycabta:
@hsbt Now, I have many commits for new Pull Requests, but I can't publish because those depend on some other Pull Requests what are already made. So let's handle all Pull Requests in existence at next chance.

hsbt:
I got an explanation for this changes from @aycabta . I understood this.

PR #470, opened 2017-07-24
Handle redefinable operators as identifier

aycabta:
For example, operators is highlighting as a operator, but redefineable operators as method is highlighting as a method in GitHub:

```ruby
def sample_method
end

def +(a, b) # this method name is the same color of above
  a + b # this operator is defferent color from method
end
```

This Pull Request is fixes the behavior.

When `@lex_state` is `:EXPR_FNAME` or `:EXPR_DOT`, the token is method name. So This Pull Request handles redefineable operators as `TkId` (identifier) in the case.

### DIFF
--- a/lib/rdoc/ruby_lex.rb
+++ b/lib/rdoc/ruby_lex.rb
@@ -479,11 +479,7 @@ class RDoc::RubyLex
       Token(TkNL)
     end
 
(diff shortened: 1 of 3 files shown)


hsbt:
Thanks for an additional unit test. It is reasonable change.

PR #471, opened 2017-07-26
Stop dropping continuing backslash

aycabta:
In string or regexp literal, `\` is a escape character and continuing `\` is dropped like below:

```ruby
/\\\\/ # => /\\/
```

This Pull Request stops the dropping:

```ruby
/\\\\/ # => /\\\\/
```


hsbt:
I'm not sure why it only fix L1238. But this pull request fixed realworld usecase.

I'm ok to merge this.

PR #473, opened 2017-07-28
Fix for "\n" after backslash escape

aycabta:
The "\n" after backslash escape is not correctly now, so the code below is broken in HTML.

```ruby
a = 1 if first_flag_var and \
  this_is_flag_var
```

broken HTML

This Pull Request fixes it.

fixed HTML


amatsuda:
Need not to add a test for this?

aycabta:
OK, I added a test for this.

aycabta:
OK, I added a test for this.

hsbt:
confirmed. Thanks!

PR #474, opened 2017-08-02
Fix escape in string

aycabta:


aycabta:
This depends on #469.

hsbt:
Thanks to add feature and unite tests.

PR #477, opened 2017-08-03
Fix handling hash rocket after symbol

aycabta:
The method what creates `TkIDENTIFIER`, `identify_identifier`, it takes continuing `=` for method name as symbol because `RDoc::RubyLex` handles some symbols via `TkSYMBEG` and `TkIDENTIFIER`. But it is sometimes broken for hash literal, like below:

```ruby
{ :method_name => :value } # keys is :method_name, success
{ :method_name= => :value } # key is :method_name=, success
{ :method_name=> :value } # key is :method_name, failure
```

This Pull Request fixes the behaviour.


hsbt:
Test case seems to good.

PR #478, opened 2017-08-03
Fix percent string literal with operator

aycabta:
For example, `%*string*` is changed to `%Q*string*`. It's equivalent string literal to be sure, but this is not good for source code quotation in HTML document.

```ruby
def example
  a_string = %*string*
  the_same_of_above = %Q*string*
end
```

The code above is highlighted below:

```html
<pre>
<span class="ruby-keyword">def</span> <span class="ruby-identifier">example</span>
  <span class="ruby-identifier">a_string</span> = <span class="ruby-string">%Q*string*</span>
  <span class="ruby-identifier">the_same_of_above</span> = <span class="ruby-string">%Q*string*</span>
<span class="ruby-keyword">end</span>
</pre>
```

%*string* becomes %Q*string*

This Pull Request fixes it:

```html
<pre>
<span class="ruby-keyword">def</span> <span class="ruby-identifier">example</span>
  <span class="ruby-identifier">a_string</span> = <span class="ruby-string">%*string*</span>
  <span class="ruby-identifier">the_same_of_above</span> = <span class="ruby-string">%Q*string*</span>
<span class="ruby-keyword">end</span>
</pre>
```

%*string* keeps its

It's the same of the original source code.

### DIFF
--- a/lib/rdoc/ruby_lex.rb
+++ b/lib/rdoc/ruby_lex.rb
@@ -1211,7 +1211,7 @@ class RDoc::RubyLex
     str = if ltype == quoted and %w[" ' /].include? ltype then
             ltype.dup
           else
-            "%#{type or PERCENT_LTYPE.key ltype}#{PERCENT_PAREN_REV[quoted]||quoted}"
+            "%#{type}#{PERCENT_PAREN_REV[quoted]||quoted}"
           end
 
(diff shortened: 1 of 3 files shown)


hsbt:
I do not understand why existence behavior happens...

PR #480, opened 2017-08-04
Fix complex condition in for/while/until

aycabta:
The `parse_statements method` calls `skip_optional_do_after_expression` in case of `for`/`while`/`until` tokens. The `skip_optional_do_after_expression` uses `get_end_token` for checking end of loop condition expressions, but it's not correct, because the end is detected by only `do` token, newline and semi-colon token.

For example,

```ruby
class Foo
  def blah()
    for i in (k+1)...n do
    end
    for i in (k+1)...n
    end
  end
end
```

This code is highlighted below:

broken

```html
<pre>
  <span class="ruby-keyword">def</span> <span class="ruby-identifier">blah</span>()
    <span class="ruby-keyword">for</span> <span class="ruby-identifier">i</span> <span class="ruby-keyword">in</span> (<span class="ruby-identifier">k</span><span class="ruby-operator">+</span><span class="ruby-value">1</span>)<span class="ruby-operator">...</span><span class="ruby-identifier">n</span> <span class="ruby-keyword">do</span>
    <span class="ruby-keyword">end</span>
    <span class="ruby-keyword">for</span> <span class="ruby-identifier">i</span> <span class="ruby-keyword">in</span> (<span class="ruby-identifier">k</span><span class="ruby-operator">+</span><span class="ruby-value">1</span>)<span class="ruby-operator">...</span><span class="ruby-identifier">n</span>
    <span class="ruby-keyword">end</span>
  <span class="ruby-keyword">end</span>
<span class="ruby-keyword">end</span>
</pre>
```

I think this is broken. This Pull Request fixes it:

correct

```html
<pre>
<span class="ruby-keyword">def</span> <span class="ruby-identifier">blah</span>()
  <span class="ruby-keyword">for</span> <span class="ruby-identifier">i</span> <span class="ruby-keyword">in</span> (<span class="ruby-identifier">k</span><span class="ruby-operator">+</span><span class="ruby-value">1</span>)<span class="ruby-operator">...</span><span class="ruby-identifier">n</span> <span class="ruby-keyword">do</span>
  <span class="ruby-keyword">end</span>
  <span class="ruby-keyword">for</span> <span class="ruby-identifier">i</span> <span class="ruby-keyword">in</span> (<span class="ruby-identifier">k</span><span class="ruby-operator">+</span><span class="ruby-value">1</span>)<span class="ruby-operator">...</span><span class="ruby-identifier">n</span>
  <span class="ruby-keyword">end</span>
<span class="ruby-keyword">end</span>
</pre>
```

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -2022,7 +2022,6 @@ class RDoc::Parser::Ruby < RDoc::Parser
   def skip_optional_do_after_expression
     skip_tkspace false
     tk = get_tk
-    end_token = get_end_token tk
 
     b_nest = 0
     nest = 0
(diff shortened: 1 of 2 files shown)


hsbt:
Confirmed. Thanks.

PR #481, opened 2017-08-08
Fix n-ary detecting condition for rational and imaginary

aycabta:
The `0i` as imaginary and `0r` as rational are confusable with n-ary notation, such as `0x1F` as hexadecimal, `0o37` and `037` as octal, `0b011111` as binary and others.


hsbt:
Understood. Thanks.

PR #484, opened 2017-08-09
Fix postfix if after escaped newline

aycabta:
In the `identify_identifier` method, if `@lex_state != :EXPR_BEG && @lex_state != :EXPR_FNAME`, `TkIF` becomes `TkIF_MOD`. So when `"\n"` comes after escaped backslash, doesn't set `:EXPR_BEG` to `@lex_state` with this commit. This commit fixes correctly handling `TkIF_MOD` after escaped newline.

### DIFF
--- a/test/test_rdoc_ruby_lex.rb
+++ b/test/test_rdoc_ruby_lex.rb
@@ -428,6 +428,44 @@ U
     assert_equal expected, tokens
   end
 
+  def test_class_tokenize_postfix_if_after_escaped_newline
+    tokens = RDoc::RubyLex.tokenize <<'RUBY', nil
+def a
+  1 if true
+  1 \
+    if true
+end
+RUBY
+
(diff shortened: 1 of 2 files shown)


test/test_rdoc_ruby_lex.rb
```diff
@@ -591,96 +591,39 @@ def test_class_tokenize_string
     assert_equal expected, tokens
   end
 
-  def test_class_tokenize_string_with_escape
+  def test_class_tokenize_postfix_if_after_escaped_newline
```
hsbt: `test_class_tokenize_string_with_escape` is needless test case? It seems regression.

aycabta:
I fixed merging what was handled in a rush yesterday.

hsbt:
I confirmed test-case and your patch.

PR #485, opened 2017-08-09
Fix postfix if/unless/while/until after heredoc

aycabta:
The postfix `if`/`unless`/`while`/`until` are detected by what `@lex_state` is `:EXPR_END`. So sets `:EXPR_END` on `@lex_state` when `TkHEREDOCBEG` is created.


hsbt:
I confirmed to assertion for `Foo#blah` markup.

PR #486, opened 2017-08-09
Fix escape behavior in single quote string literal

aycabta:
Remove addtion unnecessary `\` after escape character `\`.


aycabta:
This Pull Request depends on https://github.com/ruby/rdoc/pull/474.

hsbt:
I confirmed to correct test case.

PR #487, opened 2017-08-11
Support safe navigation operator

aycabta:
The safe navigation operator is an operator from the name, but it's a variety of dot token for method calling. So `TkSAFENAV` class inherited `Token` class instead of `TkOp` class.


aycabta:
I'm happy for your working! :relaxed: :star2:

PR #488, opened 2017-08-11
Fix creation of symbol token

aycabta:
Symbol tokens are created form `TkIDENTIFIER` with `":"`, but there are several other cases. Subclasses of `TkId`can become symbol token with `":"`, for example reserved keywords can become symbol;

```ruby
{ undef: :undef }
```

 This Pull Request fixes it.


hsbt:
A test case is good to me.

PR #489, opened 2017-08-11
Separate "foo!=bar" to "foo", "!=" and "bar"

aycabta:


hsbt:
I heard why `foo?` was assigned `TkFID from @aycabta. But I could find a real reason for `TkFid`.

But I confirmed to assign `!=` to `TkNEQ`.

PR #490, opened 2017-08-11
Fix opassign

aycabta:
The operator assignment, `%=`, it's rounded to `=`. It's just bug. This Pull Request fixes it.


hsbt:
Confirmed. But I wonder why `TkOPASGN` does not need to `=` character.

PR #491, opened 2017-08-12
Fix for symbol detection when nested method

aycabta:
When nested method as identifer detected, `@lex_state` is `:EXPR_MID` and it becomes `:EXPR_END`. However, in `:` token processing, `TkSYMBEG` is generated only when `@lex_state` is not `:EXPR_END`. So symbol for nested method becomes `TkCOLON` as distinct from `TkSYMBEG`.

This Pull Request fixes the behavior.

Before:

before

The symbol, `:name`, it consists `TkCOLON` and `TkIDENTIFIER`.

After(fixed):

after(fixed)

The `:name` is a symbol.

### DIFF
--- a/test/test_rdoc_ruby_lex.rb
+++ b/test/test_rdoc_ruby_lex.rb
@@ -520,6 +520,21 @@ U
     assert_equal expected, tokens
   end
 
+  def test_class_tokenize_symbol_for_nested_method
+    tokens = RDoc::RubyLex.tokenize 'return untrace_var :name', nil
+
+    expected = [
+      @TK::TkRETURN    .new( 0, 1,  0, "return"),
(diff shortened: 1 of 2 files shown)


test/test_rdoc_ruby_lex.rb
```diff
@@ -789,28 +789,16 @@ def test_class_tokenize_symbol
     assert_equal expected, tokens
   end
 
-  def test_class_tokenize_particular_kind_of_symbols
```
hsbt: Is it a regression?

`est_class_tokenize_particular_kind_of_symbols` is still working test case.

aycabta:
I fixed merging yesterday. Haste makes waste...

hsbt:
Thanks. Looks reasonable.

PR #492, opened 2017-08-13
Support escaped character literal

aycabta:
Like `?\s`.


hsbt:
I heard why change original test case named `test_class_tokenize_character_literal`. It test case is better to use single character.

And additional test case is good to me.

PR #493, opened 2017-08-13
Fix symbol handle in args of yield/super/defined?

aycabta:
In `:` token processing, `TkSYMBEG` is generated only when `@lex_state` is not `:EXPR_END`. But `yield`/`super`/`defined?` is `:EXPR_END`. So if first arg in those is `:`, it becomes `TkCOLON` as distinct from `TkSYMBEG`.

This Pull Request fixes it.


aycabta:
Thanks :heartpulse: a :heartbeat: lot!

PR #496, opened 2017-08-18
Fix postfix :nodoc:

aycabta:
In `parse_statements`, when `TkEND === tk and nest == 0`, checks postfix `:nodoc:` for class even where inside methods. So if a method has postfix `:nodoc:`, it raises strange behavior.

This Pull Request fixes it.


hsbt:
Thanks. It's useful fixes.

PR #497, opened 2017-08-18
Fix regexp literal in oneliner

aycabta:
Regexp literal is detected by what `@lex_state` is `:EXPR_BEG` or `:EXPR_MID`. But when regexp literal is after method name in oneliner, `@lex_state` is `:EXPR_END` and the regexp literal becomes broken tokens.

This Pull Request fixes it.

### DIFF
--- a/test/test_rdoc_parser_ruby.rb
+++ b/test/test_rdoc_parser_ruby.rb
@@ -2499,6 +2499,28 @@ EXPTECTED
     assert_equal markup_code, expected
   end
 
+  def test_parse_statements_method_oneliner_with_regexp
+    util_parser <<RUBY
+class Foo
+  def blah() /bar/ end
+end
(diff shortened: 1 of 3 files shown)


test/test_rdoc_parser_ruby.rb
```diff
@@ -2499,22 +2499,15 @@ def blah()
     assert_equal markup_code, expected
   end
 
-  def test_parse_statements_postfix_if_after_heredocbeg
```
hsbt: Is it needless test case? It seems regression with merging master branch.

aycabta:
I fixed it now, I mistook merging yesteday. More haste, less speed.

hsbt:
I heared this fix resolve `ruby-regexp` parsing.

PR #498, opened 2017-08-22
Support quoted symbol for JSON-style hash literal

aycabta:
Like below:

{ 'first': 1, "second": 2 }


hsbt:
Test case seems good. Thanks.

PR #500, opened 2017-08-25
Fix % literal with the same sign of original

aycabta:
In `identify_string method`, `%` literal is detected by literal type and quote sign. So `%r/regexp/` is treated as `/regexp/`, because literal type is regexp and quote sign is `/`, it's the same. This commit fix it by checking `type` argument what is given only when `%` literal.


hsbt:
Confirmed. Thanks.

PR #502, opened 2017-08-26
Context#each_section yields only #display? items

aycabta:
`Context#each_section` yields all constants and attributes, but RDoc treats only displayed items for documentation. The Darkfish template, `lib/rdoc/generator/template/darkfish/class.rhtml`, it [retrieves items only what `#display?` is true inside `#each_section`](https://github.com/ruby/rdoc/blob/a92b02ee7e717f5d291afbf549d02230fbd492db/lib/rdoc/generator/template/darkfish/class.rhtml#L29-L30). It's strange behavior. I think `#each_section` should return only displayed items.

This Pull Request fixes it.

### DIFF
--- a/lib/rdoc/context.rb
+++ b/lib/rdoc/context.rb
@@ -762,7 +762,7 @@ class RDoc::Context < RDoc::CodeObject
     attributes.default = []
 
     sort_sections.each do |section|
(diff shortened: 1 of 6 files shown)


hsbt:
Confirmed. `displayed?` filter should run before template rendering.

PR #503, opened 2017-08-29
homepage in the gemspec points at non-existent page

amatsuda:
The homepage is also located under ruby organization now.


hsbt:
Thanks!

PR #505, opened 2017-08-29
Do not require rdoc from rdoc.gemspec

rhenium:
Use require_relative instead.

When running under a Bundler environment (e.g., 'bundle exec rake -T'),
RubyGems first activates the RDoc installed globally through the require
statement in rdoc.gemspec. This is problematic since Bundler then tries
to activate the gemspec defined in rdoc.gemspec. They don't match and
this causes an error like this:

	bundler: failed to load command: rake (/tmp/rdoc/.bundle/gems/ruby/2.5.0/bin/rake)
	Gem::LoadError: You have already activated rdoc 5.1.0, but your Gemfile requires rdoc 6.0.0.beta1. Since rdoc is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports rdoc as a default gem.
	  /opt/ruby/trunk/lib/ruby/gems/2.5.0/gems/bundler-1.15.4/lib/bundler/runtime.rb:317:in `check_for_activated_spec!'
	  /opt/ruby/trunk/lib/ruby/gems/2.5.0/gems/bundler-1.15.4/lib/bundler/runtime.rb:32:in `block in setup'
	  /opt/ruby/trunk/lib/ruby/2.5.0/forwardable.rb:229:in `each'
	  /opt/ruby/trunk/lib/ruby/2.5.0/forwardable.rb:229:in `each'
	  /opt/ruby/trunk/lib/ruby/gems/2.5.0/gems/bundler-1.15.4/lib/bundler/runtime.rb:27:in `map'
	  /opt/ruby/trunk/lib/ruby/gems/2.5.0/gems/bundler-1.15.4/lib/bundler/runtime.rb:27:in `setup'
	  /opt/ruby/trunk/lib/ruby/gems/2.5.0/gems/bundler-1.15.4/lib/bundler.rb:101:in `setup'
	  /opt/ruby/trunk/lib/ruby/gems/2.5.0/gems/bundler-1.15.4/lib/bundler/setup.rb:9:in `<top (required)>'
	  /opt/ruby/trunk/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:60:in `require'
	  /opt/ruby/trunk/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:60:in `require'

### DIFF
--- a/rdoc.gemspec
+++ b/rdoc.gemspec
@@ -52,4 +52,5 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
   s.add_development_dependency("racc", "> 1.4.10")
   s.add_development_dependency("kpeg")
   s.add_development_dependency("minitest", "~> 4")
+  s.add_development_dependency("json")
 end
--- a/rdoc.gemspec
+++ b/rdoc.gemspec
@@ -1,6 +1,4 @@
-# -*- encoding: utf-8 -*-
-$:.unshift File.expand_path("../lib", __FILE__)
-require 'rdoc'
+require_relative "lib/rdoc"
 
 Gem::Specification.new do |s|
   s.name = "rdoc"


aycabta:
I worried about RDoc documentation step in CRuby building with this patch, but I verified what it works fine.

rdoc.gemspec
```diff
@@ -1,6 +1,4 @@
-# -*- encoding: utf-8 -*-
-$:.unshift File.expand_path("../lib", __FILE__)
-require 'rdoc'
+require_relative "lib/rdoc"
```
aycabta: Almost `require` in RDoc are with single-quote. The command below is for listing `require` with double-quote:

```bash
$ git grep '^require "'           
bin/console:require "bundler/setup"                        
bin/console:require "rdoc"                                 
bin/console:require "irb"                                  
lib/rdoc/ruby_lex.rb:require "e2mmap"                      
lib/rdoc/ruby_lex.rb:require "irb/slex"                    
lib/rdoc/ruby_lex.rb:require "stringio"                    
test/test_rdoc_parser_ruby.rb:require "\#{prefix}/a_library"
```

The `bin/console` is auto-generated code as you know, the `lib/rdoc/ruby_lex.rb` is forked source code from IRB and last one is dynamic string. So I think what you should use single-quote for this `require`.

rhenium: I chose double-quotes following the existing string literals in rdoc.gemspec. Do you want me to replace them all?

drbrain:
I now think double quotes for most strings is best, but I know it breaks tradition with old rdoc code.

rdoc.gemspec
aycabta: I explained incompletely. The double quotes for `require` are stemmed from historical background. I just wanted to say about most `require` are used with single quote. But it's not important if I really think about it.

aycabta:
OK, I understand this problem and I sometimes encountered the same situation with Bunlder. This is important fixing.

The string literal type is minor important in the now.

PR #506, opened 2017-08-29
Support constant visibility

rhenium:
Parse Module#private_constant and Module#public_constant calls and set
the visibility for the constants appropriately.

Also let RDoc::Context#remove_invisible filter constants as well as
methods and attributes, so that private constants won't be rendered.


aycabta:
I understand what you want to do, and looks good for me.

PR #507, opened 2017-08-30
Supports multi-line constant body with escaped newline

aycabta:
In `parse_constant_body` method, `TkNL` is used as terminator of constant body, but `TkBACKSLASH` can be escape character for `TkNL` in Ruby syntax.

This Pull Request fixes it.


aycabta:
This is simple fix.

PR #508, opened 2017-08-30
Use IO.console.winsize instead of IO.console_size for JRuby 9k

aycabta:


aycabta:
I think this is simple to understand and easily trouble.

PR #512, opened 2017-09-06
Use Ripper for lexical analyser

aycabta:
I replaced legacy lexical analyser from IRB with Ripper what is standard library of Ruby. It's important to support new syntax.

I explain what this code can parse Ruby code with checking diff for all Ruby documents and all Rails documents.

# Preparing

## Checkout each branches

```bash
mkdir test-for-rdoc
cd test-for-rdoc
git clone -b master --single-branch git@github.com:ruby/rdoc.git original-rdoc
cd original-rdoc
git checkout 0c5eb20 # This is HEAD of master as of now
bundle
cd ..
git clone -b use-ripper --single-branch git@github.com:aycabta/rdoc.git ripper-rdoc
cd ripper-rdoc
git checkout 23f12c1 # This is HEAD of use-ripper branch as of now
bundle
cd ..
git clone -b trunk --single-branch git@github.com:ruby/ruby.git
cd ruby
git checkout d68a6b3
cd ..
git clone -b master --single-branch git@github.com:rails/rails.git
cd rails
git checkout 665ac7c
cd ..
```

## Generate CRuby documents

```bash
ruby -Ioriginal-rdoc/lib original-rdoc/exe/rdoc -o ruby-doc-original ruby
ruby -Iripper-rdoc/lib ripper-rdoc/exe/rdoc -https://github.com/zzak/sdoc/pull/112o ruby-doc-ripper ruby
```

## Generate Rails documents

```bash
cd rials
# Tha patch below for removing monkey patch (https://github.com/zzak/sdoc/pull/112) and,
# using suitable variable (https://github.com/zzak/sdoc/pull/113) with
# suitable behavior of RDoc (https://github.com/ruby/rdoc/pull/502)
curl -fsSL https://gist.githubusercontent.com/aycabta/1d4eac210d3ede1b91a520804499a457/raw/ca90a8ce8fe3089a1d45629a01550012b2be46e1/sdoc.diff | patch -p1 
bundle update
RUBYOPT=-I../original-rdoc/lib bundle exec rake rdoc
mv doc/rdoc ../rails-doc-original
RUBYOPT=-I../ripper-rdoc/lib bundle exec rake rdoc
mv doc/rdoc ../rails-doc-ripper
cd ..
```

### Ruby

```bash
diff -ru --exclude "*.js" --exclude "*.gz" --exclude "*.css" --exclude "*.png" --exclude "*.gif" --exclude "*.ttf" rails-doc-original rails-doc-ripper | less
```

### Rails

```bash
[...]

PR #515, opened 2017-09-10
Use "bundle exec" in .travis.yml for JRuby test

aycabta:
JRuby test succeeded on Travis CI finally and I made sure that [`use-ripper` branch](https://github.com/ruby/rdoc/pull/512) with these commits too.

The `test_accept_verbatim_nl_after_backslash` is failed on JRuby because Ripper on JRuby fails to lexical analyse around escaped newline. I filed a issue for it: jruby/jruby#4787. The `skip` should be removed after the issue is resolved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ rvm:
 env:
   global:
     NOBENCHMARK=1
-script: rake
+script: bundle exec rake
 matrix:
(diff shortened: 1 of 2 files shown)


hsbt:
I'm ok to skip `test_accept_verbatim_nl_after_backslash`.

PR #521, opened 2017-09-21
Prevent negative IDs in output of #inspect.

voxik:
This prevents output of negative object IDs:

https://bugs.ruby-lang.org/issues/13397

I have not tested this closely, but:

1. I hope the test would catch the possible issues.
2. It should not by user visible change ...

Please note that this is updated version of PR #451. The only change is dropped lib/rdoc/ruby_lex.rb patch, since it vas replaced by lib/rdoc/ruby_token.rb, where the inspects looks to be correct.

### DIFF
--- a/lib/rdoc/alias.rb
+++ b/lib/rdoc/alias.rb
@@ -75,8 +75,8 @@ class RDoc::Alias < RDoc::CodeObject
 
   def inspect # :nodoc:
     parent_name = parent ? parent.name : '(unknown)'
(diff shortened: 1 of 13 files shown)


hsbt:
We will merge after closing issue on ruby core tracker.

ioquatix:
@hsbt should we merge this now?

hsbt:
https://bugs.ruby-lang.org/issues/13397#note-10

PR #532, opened 2017-09-29
Fix broken links

tnir:
cf. 20af2f6


aycabta:
OK, I checked what all URLs are correct.

Issue #45, opened 2011-06-10
Distance recommendations for ri 


erikh:
Could work similar to rubygems' levenshtein matcher, e.g.:

```
 $ ri Hash.deltee
 WHY HELLO THERE PILGRIM: DID YOU MEAN Hash#delete ?
```

PR #533, opened 2017-10-02
Support did_you_mean

aycabta:
Ruby 2.3 or later, did_you_mean gem is adopted as standard library. The did_you_mean gem can be used as a library via `DidYouMean::SpellChecker` at 1.0.3 or later, so checks that the class is defined and uses it. And then, this feature will become enabled if the gem 1.0.3 or later is installed.

This closes https://github.com/ruby/rdoc/issues/45.

### DIFF
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -939,7 +939,7 @@ or the PAGER environment variable.
         else
           false
         end
-      rescue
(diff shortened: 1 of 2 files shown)


yuki24:
looks good to me 👍🏻 thanks!

hsbt:
Great! It seems interesting and useful feature at RDoc 6.

Issue #433, opened 2016-12-10
:arg:, :args: doesn't work.


tonytonyjan:
[The example from rdoc document](http://docs.seattlerb.org/rdoc/RDoc/Markup.html#class-RDoc::Markup-label-Method+arguments) doesn't work:

```ruby
##
#  :args: a, b

def some_method(*a)
end
```

But it works inline:

```ruby
def some_method(*a) # :args: a, b
end
```

My rdoc version is 5.0.

aycabta:
OK, it's reproduced if RDoc 6.0.0.beta2. I'll check this bug later.

PR #541, opened 2017-10-11
Support :args: directive before method

aycabta:
The `:args:` directive is supported only by after `def` keyword line like below:

```ruby
  def method_with_args_comment(*array) # :args: a, b, c
  end
  # => documented as method_with_args_comment(a, b, c)
```

This commit changes to support like below:

```ruby
  ##
  # :args: a, b, c
  def meth(*array)
  end
  # => documented as meth(a, b, c)
```

This fixes https://github.com/ruby/rdoc/issues/433.


hsbt:
New tests is fine to me.

Issue #180, opened 2013-02-05
rdoc only parses one definition per file


zzak:
I was trying to move lib/rubygems.rb module `Gem` definition to the top module definition, it's defined twice in the file. Only adding the docs to the second definition will get parsed by rdoc.

Here's a contrived example:

``` ruby
# omg
module Foo
end

# wtf
module Foo
end
```

If you generate the documentation for this file, you will only see the "wtf".

![double module](http://screencloud.net/img/screenshots/4e9dbffe9d456dd7f165fc2902859b7e.png)

PR #542, opened 2017-10-14
Handle all definitions per file in Ruby

aycabta:
`RDoc::Parser::C` cann't discern each documents per file for one class, so `RDoc::ClassModule` adopts the latest comment per file, however `RDoc::Parser::Ruby` can take each documents per file for one class. This commit changes the behavior only when `RDoc::ClassModule` is for `RDoc::Parser::C`. Of course this works around `RDoc::Parser::C`.

And I checked all CRuby's documents and 2 documents change with this patch because the class/module has re-definition line without document after first definition line with document. Those are [`ERB`](https://github.com/ruby/ruby/blob/25aec0b8352428871b97de41f7da9abff1ab7d90/lib/erb.rb#L261) (re-definitions without document: [1](https://github.com/ruby/ruby/blob/25aec0b8352428871b97de41f7da9abff1ab7d90/lib/erb.rb#L272), [2](https://github.com/ruby/ruby/blob/25aec0b8352428871b97de41f7da9abff1ab7d90/lib/erb.rb#L751), [3](https://github.com/ruby/ruby/blob/25aec0b8352428871b97de41f7da9abff1ab7d90/lib/erb.rb#L960), [4](https://github.com/ruby/ruby/blob/25aec0b8352428871b97de41f7da9abff1ab7d90/lib/erb.rb#L1008)) and [`DRb::DRbProtocol`](https://github.com/ruby/ruby/blob/25aec0b8352428871b97de41f7da9abff1ab7d90/lib/drb/drb.rb#L721) ([re-definition without document](https://github.com/ruby/ruby/blob/25aec0b8352428871b97de41f7da9abff1ab7d90/lib/drb/drb.rb#L1018)). Those documents are lost now, and with this patch, those become to be treated correctly.

ref. https://github.com/ruby/rdoc/commit/ecae6cb52319d6791950fd9a1dc1e2d9ad518b19, https://github.com/ruby/rdoc/commit/d84ada248f8f3c2fffa5f81ce60dfc9a7bc9b8c6, https://github.com/ruby/rdoc/commit/e033904abe51ece5f7857b3767054bbd08f65eaa

This closes https://github.com/ruby/rdoc/issues/180.

### DIFF
--- a/lib/rdoc/class_module.rb
+++ b/lib/rdoc/class_module.rb
@@ -136,7 +136,9 @@ class RDoc::ClassModule < RDoc::Context
                 normalize_comment comment
               end
 
-    @comment_location.delete_if { |(_, l)| l == location }
+    if location.parser == RDoc::Parser::C
+      @comment_location.delete_if { |(_, l)| l == location }
+    end
 
     @comment_location << [comment, location]
 
--- a/test/test_rdoc_class_module.rb
+++ b/test/test_rdoc_class_module.rb
@@ -42,7 +42,8 @@ class TestRDocClassModule < XrefTestCase
     cm.add_comment '# comment 1', tl1
     cm.add_comment '# comment 2', tl1
 
-    assert_equal [['comment 2', tl1]], cm.comment_location
+    assert_equal [['comment 1', tl1],
+                  ['comment 2', tl1]], cm.comment_location
   end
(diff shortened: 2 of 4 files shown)


hsbt:
I understood detail of this fixes.

PR #546, opened 2017-10-19
Don't use lex_state emulation on Ruby 2.5 or later

aycabta:
`RDoc::RipperStateLex` has `InnerStateLex` that is emulation layer about `lex_state` of parse.y, but after Ruby 2.5 or later, `Ripper` gives `lex_state` value as `state` method inside `Ripper::Filter`'s callback. On this Pull Request, uses the new feature of `Ripper` on Ruby 2.5 or later.

### DIFF
--- a/lib/rdoc/parser/ripper_state_lex.rb
+++ b/lib/rdoc/parser/ripper_state_lex.rb
@@ -1,6 +1,9 @@
 require 'ripper'
 
 class RDoc::RipperStateLex
+  # TODO: Remove this constants after Ruby 2.4 EOL
+  RIPPER_HAS_LEX_STATE = (RUBY_VERSION >= '2.5.0')
+
   EXPR_NONE = 0
   EXPR_BEG = 1
   EXPR_END = 2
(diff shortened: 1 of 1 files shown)


lib/rdoc/parser/ripper_state_lex.rb
```diff
@@ -1,6 +1,9 @@
 require 'ripper'
 
 class RDoc::RipperStateLex
+  # TODO: Remove this constants after Ruby 2.4 EOL
+  RIPPER_HAS_LEX_STATE = (RUBY_VERSION >= '2.5.0')
```
hsbt: Should we use `respond_to?` instead of `RUBY_VERSION` ?

hsbt: or `defined?`

aycabta:
OK, I'll use `Ripper::Filter.method_defined?(:state)`.

lib/rdoc/parser/ripper_state_lex.rb
```diff
@@ -1,6 +1,9 @@
 require 'ripper'
 
 class RDoc::RipperStateLex
+  # TODO: Remove this constants after Ruby 2.4 EOL
+  RIPPER_HAS_LEX_STATE = Ripper::Filter.method_defined?(:state)
```
hsbt: 👍

PR #549, opened 2017-11-08
Fix tag in table_of_contents.html

aycabta:
`RDoc::Markup::AttributeManager#convert_html` that is called from `RDoc::Markup::AttributeManager#flow` replaces HTML tags of received `str` with `NULL` characters in HTML generation for each files, and the `NULL` characters are removed in `RDoc::Markup::AttributeManager#split_into_flow`. But the `String` objects are reused for table_of_contents.html generation, so the tags lose at that moment.

This Pull Request fixes it with using duplicated `String` object.


aycabta:
This depends on https://github.com/ruby/rdoc/pull/550 for CI passes tests.

PR #553, opened 2017-11-10
 Remove unnecessary condition of RDoc::TokenStream.to_html

aycabta:
`RDoc::TokenStream.to_html` checks that `:on_op` and `EXPR_ARG` token comes for treating it as identifier but the case never come because `RDoc::RipperStateLex#get_op_tk` set `:on_ident` and `EXPR_ARG` when `:on_op` as identifier before `RDoc::TokenStream.to_html`.


hsbt:
I confirmed to the same result with master branch applied to patch additional test of this branch.

Issue #357, opened 2015-07-31
rdoc + darkfish + dry-run aborts


rovf:
The command

```
rdoc --dry-run --root=lib --format=darkfish --op=rdocs
```

aborts with 

```
 No such file or directory @ rb_sysopen - /home/FISRONA/gitwrk/vp5/rdocs/js/search_index.js
```

unless directory rdocs already contains files from a previous generation. It seems that --dry-run doesnt anticipate, that these intermediate directories will be generated in time.

Using rdoc 4.2.0 (Cygwin)

zzak:
Ahh, interesting.. although I cannot reproduce it, I wonder if this is related to #334 based on the file mentioned in your backtrace.

Could you still reproduce it with a full backtrace? Thanks!

rovf:
Run it again with --debug (I guess this is sufficient for getting full backtraced?):

Compressing generated JSON index
Reading the JSON index file from /home/FISRONA/gitwrk/playground/rdocs/js/search_index.js
Errno::ENOENT: No such file or directory @ rb_sysopen - /home/FISRONA/gitwrk/playground/rdocs/js/search_index.js
  /usr/share/gems/gems/rdoc-4.2.0/lib/rdoc/generator/json_index.rb:167:in `read'
  /usr/share/gems/gems/rdoc-4.2.0/lib/rdoc/generator/json_index.rb:167:in `read'
  /usr/share/gems/gems/rdoc-4.2.0/lib/rdoc/generator/json_index.rb:167:in `generate_gzipped'
  /usr/share/gems/gems/rdoc-4.2.0/lib/rdoc/generator/darkfish.rb:249:in `generate'
  /usr/share/gems/gems/rdoc-4.2.0/lib/rdoc/rdoc.rb:530:in `block in generate'
  /usr/share/gems/gems/rdoc-4.2.0/lib/rdoc/rdoc.rb:525:in `chdir'
  /usr/share/gems/gems/rdoc-4.2.0/lib/rdoc/rdoc.rb:525:in `generate'
  /usr/share/gems/gems/rdoc-4.2.0/lib/rdoc/rdoc.rb:508:in `document'
  /usr/share/gems/gems/rdoc-4.2.0/bin/rdoc:20:in `<top (required)>'
  /usr/bin/rdoc:23:in `load'
  /usr/bin/rdoc:23:in `<main>'
No such file or directory @ rb_sysopen - /home/FISRONA/gitwrk/playground/rdocs/js/search_index.js
[...]

PR #556, opened 2017-11-11
Fix --dry-run option

aycabta:
This fixes https://github.com/ruby/rdoc/issues/357.

### DIFF
--- a/test/test_rdoc_rdoc.rb
+++ b/test/test_rdoc_rdoc.rb
@@ -51,11 +51,21 @@ class TestRDocRDoc < RDoc::TestCase
 
     rdoc = RDoc::RDoc.new
 
+    out = nil
     temp_dir do
-      capture_io do
+      out, = capture_io do
         rdoc.document options
       end
+
+      refute File.directory? 'doc'
+      assert_equal rdoc, rdoc.store.rdoc
     end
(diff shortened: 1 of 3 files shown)


test/test_rdoc_rdoc.rb
```diff
@@ -40,6 +40,24 @@ def test_document # functional test
...
+    options.setup_generator 'darkfish'
+    options.main_page = 'MAIN_PAGE.rdoc'
+    options.root      = Pathname File.expand_path('..', __FILE__)
+    options.title     = 'title'
+    options.dry_run = true
+
+    rdoc = RDoc::RDoc.new
+
+    temp_dir do
+      capture_io do
```
hsbt: Can we confirm any assertions? We should assert that has no error or not generate rdoc files.

test/test_rdoc_rdoc.rb
```diff
@@ -40,6 +40,34 @@ def test_document # functional test
...
+
+    rdoc = RDoc::RDoc.new
+
+    out = nil
+    temp_dir do
+      out, = capture_io do
+        rdoc.document options
+      end
+
+      refute File.directory? 'doc'
```
hsbt: 👍

PR #560, opened 2017-11-15
Add lib/rdoc/markdown/literals.rb to .gitignore

aycabta:
The file is generated by KPeg in Rakefile.


hsbt:
Good catch, I forgot this file is auto-generation file.

PR #569, opened 2017-12-14
Do not store current timestamps in gz headers

bmwiedemann:
to enable reproducible builds of rdoc

Normally, 0 would be the preferred value to indicate "no date"
but that value is handled differently in Zlib::GzipWriter
to put in the current time

### DIFF
--- a/lib/rdoc/generator/json_index.rb
+++ b/lib/rdoc/generator/json_index.rb
@@ -175,7 +175,7 @@ class RDoc::Generator::JsonIndex
     debug_msg "Writing gzipped search index to %s" % outfile
 
     Zlib::GzipWriter.open(outfile) do |gz|
-      gz.mtime = File.mtime(search_index_file)
+      gz.mtime = 1 # make output reproducible
       gz.orig_name = search_index_file.basename.to_s
       gz.write search_index
       gz.close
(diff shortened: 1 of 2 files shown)


lib/rdoc/generator/json_index.rb
```diff
@@ -175,7 +175,7 @@ def generate_gzipped
     debug_msg "Writing gzipped search index to %s" % outfile
 
     Zlib::GzipWriter.open(outfile) do |gz|
-      gz.mtime = File.mtime(search_index_file)
+      gz.mtime = 1 # make output reproducible
```
aycabta: I think the `mtime` should be updated because the `search_index.js.gz` is generated dynamically. Please use SOURCE_DATE_EPOCH.

lib/rdoc/generator/json_index.rb
```diff
@@ -193,7 +193,7 @@ def generate_gzipped
         debug_msg "Writing gzipped file to %s" % outfile
 
         Zlib::GzipWriter.open(outfile) do |gz|
-          gz.mtime = File.mtime(dest)
+          gz.mtime = 1 # make output reproducible
```
aycabta: Please fix `FileUtils.install` with `:preserve` option in `RDoc::Generator::JsonIndex#generate` because the `mtime` is based on original `.js` file.

For your information, `FileUtils.install` is based on "install(1)" command.

PR #570, opened 2017-12-14
created.rid: use SOURCE_DATE_EPOCH

bmwiedemann:
use SOURCE_DATE_EPOCH instead of current time in created.rid top line
to enable reproducible builds of ruby docs

See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

### DIFF
--- a/lib/rdoc/rdoc.rb
+++ b/lib/rdoc/rdoc.rb
@@ -232,6 +232,9 @@ option)
 
   def update_output_dir(op_dir, time, last = {})
     return if @options.dry_run or not @options.update_output_dir
+    unless ENV['SOURCE_DATE_EPOCH'].nil?
+      time = Time.at(ENV['SOURCE_DATE_EPOCH'].to_i).gmtime
+    end
 
     open output_flag_file(op_dir), "w" do |f|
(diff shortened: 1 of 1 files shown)


aycabta:
OK, I understand what you want to do. It's very great for my RDoc development.

aycabta:
I added a test for this at https://github.com/ruby/rdoc/commit/819461271cf1287096b98834ec8d09da2c872a84.

aycabta:
Commented: https://bugs.ruby-lang.org/issues/13627#change-69519

utkarsh2102:
Hm, I think this doesn't seem to work.
Reproducibility of ruby-dbus is affected because of rdoc: 
[ruby-dbus.txt](https://github.com/ruby/rdoc/files/4666319/ruby-dbus.txt)

Any clue what's wrong? :)

bmwiedemann:
@utkarsh2102  Looks like a different issue with mtimes in data.tar.xz - shouldnt those be normalized by dpkg via `SOURCE_DATE_EPOCH` ?

Issue #571, opened 2017-12-22
Raise NoMethodError when installing Active Record with Rdoc 6.0.0 


y-yagi:
## RDoc version

6.0.0 

## Steps to reproduce 

Install Active Record with document generating.

```
$ gem install activerecord
Successfully installed activerecord-5.1.4
Parsing documentation for activerecord-5.1.4

RDoc::Parser::Ruby failure around line 1 of
lib/rails/generators/active_record/migration/templates/create_table_migration.rb

Before reporting this, could you check that the file you're documenting
has proper syntax:

  /home/yaginuma/.rbenv/versions/2.3.1/bin/ruby -c lib/rails/generators/active_record/migration/templates/create_table_migration.rb

RDoc is not a full Ruby parser and will fail when fed invalid ruby programs.

The internal error was:

	(NoMethodError) undefined method `[]' for nil:NilClass

ERROR:  While executing gem ... (NoMethodError)
    undefined method `[]' for nil:NilClass 
```

## Cause 

There are template files of Thor gem that extension is `rb` in Active Record. It seems that it is an error trying to parse it. 

Since we have changed the extension of templates files in https://github.com/rails/rails/commit/8dd76a7a6ff1bb7105beabb8f834ca54ab1e5fc2, this problem does not happen in Rails 5.2.0.
But old versions have this problem.

aycabta:
@y-yagi OK, I think it's critical. I'm fixing it. Please wait for it.

aycabta:
@y-yagi I got the picture. This issue is not RDoc's. I want to fix activerecord gem 5.1.4 because template files of 5.2 [are renamed](https://github.com/rails/rails/commit/8dd76a7a6ff1bb7105beabb8f834ca54ab1e5fc2). How can I send a patch to 5.1.x?

matthewd:
A change in a future 5.1.x release won't fix the problem for people who need to install older versions.

Those gems can be installed when using RDoc 5, but have the above error when RDoc 6 is present. Isn't that a regression?

aycabta:
RDoc 5.x uses legacy Ruby lexical analyzer from IRB what is written 20 years ago. It has so many bugs. RDoc 5 tries to parse template files of Rails and finishes with no errors but **it is a complex bug of legacy lexical analyzer and parser**.

RDoc 6 uses Ripper. It fails to parse template files of Rails because it's not correct as Ruby syntax.

I think ActiveRecord 5.1.x needs the patch below.

```patch
[...]

PR #573, opened 2017-12-23
Bypass ERB's Ruby syntax error

aycabta:
The legacy lexical analyzer and parser of RDoc 5 process ERB file with no errors because of complex bug. RDoc 6 what uses Ripper causes errors when processes ERB file. It was just bug fix but RDoc lost
compatibility. For example, when installs ActiveRecord 5.1.4 by `gem install activerecord -v 5.1.4`, it uses RDoc internal, so the RDoc crashes if it's RDoc 6.0.0.

This commit provides compatibility behavior with RDoc 5 or older. When an exception occurs inside Ruby parser, confirms that the source code contains `"<%"` and `"%>"` to detect ERB file. Skips it if RDoc guesses that the file is ERB by the logic.

This closes https://github.com/ruby/rdoc/issues/571.

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -178,6 +178,7 @@ class RDoc::Parser::Ruby < RDoc::Parser
     @size = 0
     @token_listeners = nil
     @scanner = RDoc::RipperStateLex.parse(content)
+    @content = content
     @scanner_point = 0
     @prev_seek = nil
     @markup = @options.markup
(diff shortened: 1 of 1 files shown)


aycabta:
@hsbt After merging this, please bring RDoc 6.0.1 to Ruby 2.5...

hsbt:
I confirmed to install `activerecord`

```
~/D/g/r/rdoc (aycabta-bypass-erb) > gem i activerecord
Fetching: activerecord-5.1.4.gem (100%)
Successfully installed activerecord-5.1.4
Parsing documentation for activerecord-5.1.4
RDoc detects ERB file. Skips it for compatibility:
lib/rails/generators/active_record/migration/templates/create_table_migration.rb
RDoc detects ERB file. Skips it for compatibility:
lib/rails/generators/active_record/migration/templates/migration.rb
Installing ri documentation for activerecord-5.1.4
Done installing documentation for activerecord after 5 seconds
1 gem installed
```

PR #574, opened 2017-12-24
Remove needless json dependency

hsbt:
It provides on all support versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,10 @@ rvm:
   - 2.3.5
   - 2.4.2
   - ruby-head
-  - jruby-9.1.15.0
 env:
   global:
     NOBENCHMARK=1
 script: rake
+matrix:
+  allow_failures:
+    - rvm: jruby-9.1.15.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,4 @@ rvm:
 env:
   global:
     NOBENCHMARK=1
-script: bundle exec rake
+script: rake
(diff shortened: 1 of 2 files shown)


aycabta:
I don't understand why CI fails...I'm surveying the reason.

hsbt:
Yes, It's very strange. I investigate too.

rhenium:
I added that in commit 82e352edc5b4a9d8ee9448515aecfb898e80d7cf so I could use `bundle exec rake test` to run rdoc's test suite. Later, as of commit 63b455a69a35a3fc49aa022cb89bb8be94482591, .travis.yml also started to use `bundle exec`.

As in the commit message, I needed json as a development dependency (just) because RubyGems' rubygems/test_case (which test/test_rdoc_rubygems_hook.rb does `require`) wants to activate the json gem explicitly. I didn't know why rubygems/test_case does that in the first place, though.

hsbt:
@rhenium Thanks for your clarification!

aycabta:
The https://github.com/ruby/rdoc/commit/08b23dd5d5732f7d683cabd4e86af72ae2dbf182 added `RDoc::RubygemsHook` that is ported from [lib/rubygems/rdoc.rb of RubyGems](https://github.com/rubygems/rubygems/blob/0749715e4bd9e7f0fb631a352ddc649574da91c1/lib/rubygems/rdoc.rb#L33-L36) (ref. https://github.com/rubygems/rubygems/commit/33060f7e21cf8b287afb9974148def548606675a). It depends on `Gem::UserInteraction` of RubyGems. As @rhenium said, RubyGems needs `json` gem explicitly. Therefore, RDoc depends on `json` gem via RubyGems.

I don't know the story so far about the porting.

hsbt:
JRuby test failure was caused by `autoload`. We should fix it another pull request.

PR #580, opened 2018-01-02
Fix method line number

### LINKED ISSUE #425: Line numbers are wrong if magic comments are used
If you use a magic comment, the line numbers are wrong as they treat the first non-magic comment line as line 1.  This probably needs fixing in `lib/rdoc/encoding.rb`.

Example:

```
$ cat a.rb && rdoc24 a.rb > /dev/null 2>&1 && fgrep line doc/Object.html
def a; end
            <pre><span class="ruby-comment"># File a.rb, line 1</span>
$ cat a2.rb && rdoc24 a2.rb > /dev/null 2>&1 && fgrep line doc/Object.html
# encoding: utf-8
def a; end
            <pre><span class="ruby-comment"># File a2.rb, line 1</span>
$ cat a3.rb && rdoc24 a3.rb > /dev/null 2>&1 && fgrep line doc/Object.html
# encoding: utf-8
# frozen-string-literal: true
def a; end
            <pre><span class="ruby-comment"># File a3.rb, line 1</span>
$ cat a4.rb && rdoc24 a4.rb > /dev/null 2>&1 && fgrep line doc/Object.html
# encoding: utf-8
# frozen-string-literal: true

def a; end
            <pre><span class="ruby-comment"># File a4.rb, line 2</span>
```

aycabta:
`RDoc::Encoding.read_file` removes magic comments, after that, Ruby parser processes the content. So method line number is reduced by the amount of the magic comment lines.

This commit fixes it by that changes the behavior of `RDoc::Encoding.read_file` to not remove magic comments.

This closes #425.

...But please wait for merging, this branch outputs strange documents. I'll add some commits to this branch later.

### DIFF
--- a/lib/rdoc/encoding.rb
+++ b/lib/rdoc/encoding.rb
@@ -7,6 +7,18 @@
 
 module RDoc::Encoding
 
+  HEADER_REGEXP = /^
+    (?:
+      \A\#!.*\n
+      |
+      ^\#\s+frozen[-_]string[-_]literal[=:].+\n
+      |
+      ^\#[^\n]+\b(?:en)?coding[=:]\s*(?<name>[^\s;]+).*\n
+      |
+      <\?xml[^?]*encoding=(?<quote>["'])(?<name>.*?)\k<quote>.*\n
+    )+
+  /xi # :nodoc:
+
   ##
   # Reads the contents of +filename+ and handles any encoding directives in
   # the file.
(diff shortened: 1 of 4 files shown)


aycabta:
I added https://github.com/ruby/rdoc/pull/580/commits/8b3b2567cf8983eac525449c94a7b955d3c1bc7e, it removes magic comments from documents and keeps correct line number. I checked all CRuby documents between with this branch and without this branch. There are only fixed line numbers differences (you can check it by [my check tool](https://gist.github.com/aycabta/ecac4400f86985d4511e3d78cf3a85cf)).

PR #581, opened 2018-01-04
Fix --width option of ri

### LINKED ISSUE #450: ri ignores -w (--width) option
Instead, the output line width is hard coded to 78 without any possibility to change, in  `lib/rdoc/markup/to_rdoc.rb`, see https://github.com/rdoc/rdoc/blob/master/lib/rdoc/markup/to_rdoc.rb#L49.

aycabta:
This closes #450.


stomar:
Thanks for fixing this!

PR #592, opened 2018-01-25
Fix `#tokens_to_s` method for token streams

robin850:
Hello,

This pull request is a tiny follow-up to #512 to fix the `tokens_to_s` method for `RDoc::TokenStream`. As the token stream isn't represented by an array of token classes anymore but rather with an array of hashes, the token's text is accessible through a key and not a method.

Have a nice day !


aycabta:
Good catch, thank you!

PR #593, opened 2018-02-02
Fixed broken test with autoload on JRuby

hsbt:
@aycabta I fixed broken tests with JRuby.


aycabta:
OK, it's good! Thank you!

PR #595, opened 2018-02-02
Remove rdoc/test_case from release gem

hsbt:
@aycabta 

`rdoc/test_case` is a needless file for the production environment. I move its file to `test/minitest_helper`. 

How do you think this?


aycabta:
This is good for me because I sometimes feel the same problem.

PR #602, opened 2018-02-22
Remove date field from gemspec.

hsbt:
It's automatically assigned when pushing to rubygems.org.

I learn it from https://github.com/ruby/rake/pull/253


aycabta:
I understand and it seems good.

PR #608, opened 2018-03-19
rb_global_function reworked (it should produce private and singleton)

ViugiNick:
By the source code of Ruby it is clear that `rb_define_global_function` is a `rb_define_module_function` for rb_mKernel, which corresponds to rb_define_private_method and rb_define_singleton_method. 

The main reason for creating this request was that global methods have public visibility.

rb_define_module_function(VALUE module, const char *name, VALUE (*func)(ANYARGS), int argc)


aycabta:
All singleton methods of `Kernel` can be accessed from global namespaces, and private methods mean secret methods for Ruby. So I think that current implementation is correct.

PR #610, opened 2018-03-20
Expand temp dir used by File.realpath. 

hsbt:
It was failed on macOS.


aycabta:
I don't use macOS now, but I think that this fix doesn't affect other platforms.

PR #611, opened 2018-03-23
Add support for 'Raises' lines in TomDoc parser

nbeyer:
* For the special case of 'initalize', handle 'Raises' as the
  indicator of the Returns section.
* Stop treating Raises lines as multine line text and let them be
  their own paragraphs

According to my interpretation of the TomDoc spec (http://tomdoc.org/), 'Raises' lines are part of the Returns section. The current RDoc TomDoc parser doesn't recognize these lines and treats them as just continuations of an preceding 'Returns' line. Additionally, constructors (#initialize methods) can be documented without a 'Returns' line, but if there's a 'Raises' line it would mean there is a Returns section. This pull request addresses both of these quirks.

### DIFF
--- a/lib/rdoc/tom_doc.rb
+++ b/lib/rdoc/tom_doc.rb
@@ -180,12 +180,19 @@ class RDoc::TomDoc < RDoc::Markup::Parser
 
       case type
       when :TEXT then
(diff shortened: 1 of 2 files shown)


nbeyer:
The commit message should be fixed at 54f6e5b.

PR #612, opened 2018-03-23
Cleanup minitest deprecation warnings about assert_nil

nbeyer:
Quick pull request that replaces all 'assert_equal nil, blah' forms with 'assert_nil blah', per the deprecation warnings in Minitest.


hsbt:
Thanks!

PR #634, opened 2018-06-27
add test to display_name

masakazutakewaka:
add test to `display_name`

this PR improves the coverage of the file `test/test_rdoc_ri_driver.rb` from 69.5% to 69.7%


aycabta:
It makes sense to add the case of `RDoc::RI::Driver#display_name`. Thank you.

PR #650, opened 2018-09-07
Added Rubocop task after file generation.

hsbt:
I always faced removing trailing-whitespace when I merged rdoc into ruby core repository.

https://github.com/ruby/ruby/commit/6a052fcd046db10ad1d04fcfc078444703bb575f

I added rubocop for removing trailing-whitespace and added rubocop task on Rakefile.


hsbt:
I know your concern. But I use `DisabledByDefault: true`. So, We only use nit-pick style cop.

PR #658, opened 2018-10-05
Prefer require_relative

marcandre:
Small but free speed gain by using `require_relative` instead of `require`

### DIFF
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -4,7 +4,7 @@
 require 'erb'
 require 'fileutils'
 require 'pathname'
-require 'rdoc/generator/markup'
+require_relative 'markup'
 
 ##
 # Darkfish RDoc HTML Generator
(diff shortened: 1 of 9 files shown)


aycabta:
@marcandre I'm considering this but I need a performance report. I'll reject any optimizations without a performance report.

aycabta:
And please add the report to the commit comment.

Example: https://github.com/ruby/rdoc/pull/640/commits/818ee979e02151885b4ae769b9f87b8f6b5f1802

marcandre:
@aycabta I'm considering your request but I need a single example or argument whereby `require_relative` would not be faster, safer and more expressive than `require`. I'll reject your request without that.

nobu:
I agree that `require_relative` will be faster (never slower at least) than `require`.
And curious how it get faster depending on the length of `$:`.

ioquatix:
We don't need performance report for this. It's simply better to prefer `require_relative`. So, if you update the PR, I will merge it.

marcandre:
@ioquatix PR rebased

hsbt:
It's totally maintainers who are @aycabta 's convenience.

marcandre:
Ping @aycabta

ioquatix (approved):

PR #660, opened 2018-10-27
Mark up code inside link text as <code>

nobu:
I want links to a class/module/method name to be `<code>`.
For instance, `Array#union` should be become `<a href="Array.html#method-i-union"><code>Array#union</code></a>`  instead of `<a href="Array.html#method-i-union">Array</a>`.


aycabta:
It makes sence!

PR #673, opened 2018-12-02
Use test-unit instead of minitest

nobu:
Minitest 6 will err `assert_equal` with `nil`, as seattlerb/minitest#779, even if the first argument is not a literal.


hsbt:
I have no strong opinion to switch test-unit from minitest. When we accept this patch, We need to change `minitest_helpr.rb` to `test_helper.rb`.

hsbt:
@nobu Can you rebase this? It's time to merge this.

nobu:
`test_helper.rb` will be a subject of tests.
I renamed it as just `helper.rb`.

Issue #681, opened 2018-12-26
Missing file from specifications causing build failures


poloka:
With release of 6.1.0, started getting errors during my site generation.  Cannot seem to load a file `rdoc/markup/regexp_handling` which looks to be introduced with the update
https://github.com/ruby/rdoc/compare/v6.0.4..v6.1.0

Troubleshooted by locking in 6.0.4 and I can create my site doc.

Upon investigating, discovered the regexp_handling.rb was not in the list of loaded files within the specification.  Will submit a PR.

```
Generating Darkfish format into /Users/poloka/temp/foo_service/target/doc...
rake aborted!
LoadError: cannot load such file -- rdoc/markup/regexp_handling
/Users/poloka/.rvm/gems/ruby-2.4.1@foo_service/gems/activesupport-5.2.2/lib/active_support/dependencies.rb:291:in `require'
/Users/poloka/.rvm/gems/ruby-2.4.1@foo_service/gems/activesupport-5.2.2/lib/active_support/dependencies.rb:291:in `block in require'
/Users/poloka/.rvm/gems/ruby-2.4.1@foo_service/gems/activesupport-5.2.2/lib/active_support/dependencies.rb:257:in `load_dependency'
/Users/poloka/.rvm/gems/ruby-2.4.1@foo_service/gems/activesupport-5.2.2/lib/active_support/dependencies.rb:291:in `require'
/Users/poloka/.rvm/gems/ruby-2.4.1@foo_service/gems/rdoc-6.1.0/lib/rdoc/markup/attribute_manager.rb:319:in `split_into_flow'
/Users/poloka/.rvm/gems/ruby-2.4.1@foo_service/gems/rdoc-6.1.0/lib/rdoc/markup/attribute_manager.rb:261:in `flow'
/Users/poloka/.rvm/gems/ruby-2.4.1@foo_service/gems/rdoc-6.1.0/lib/rdoc/markup/to_html.rb:400:in `to_html'
/Users/poloka/.rvm/gems/ruby-2.4.1@foo_service/gems/rdoc-6.1.0/lib/rdoc/markup/to_html.rb:189:in `accept_paragraph'
/Users/poloka/.rvm/gems/ruby-2.4.1@foo_service/gems/rdoc-6.1.0/lib/rdoc/markup/paragraph.rb:11:in `accept'
[...]

PR #682, opened 2018-12-26
Correction to include regexp_handling in list of loaded files

poloka:
With the update to 6.1.0, a new file was added `regexp_handling.rb` but it was not added to the specification for loading upon including gem into project.


ChapterHouse:
+1

aycabta:
Thank you, this problem(https://github.com/ruby/rdoc/issues/681) is my mistake. I'll release 6.1.1 right now.

aycabta:
Released!
https://rubygems.org/gems/rdoc

Thanks a lot to you, I realized the problem with your report and was able to quickly measures.

poloka:
Awesome!  Thanks @aycabta for the quick turnaround.

PR #695, opened 2019-02-16
Add frozen string literal support

deivid-rodriguez:
I created this but when I was forking the gem, I noticed that it's already fixed on https://github.com/ruby/rdoc/tree/fix-for-frozen-string-literal. And the test there is probably better than re-running everything :man_shrugging:.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -12,4 +12,4 @@ jobs:
         gem install bundler --no-document
         bundle install
     - name: Run test
-      run: rake
+      run: rake && RUBYOPT=--enable-frozen_string_literal rake
--- a/lib/rdoc/erb_partial.rb
+++ b/lib/rdoc/erb_partial.rb
@@ -12,7 +12,7 @@ class RDoc::ERBPartial < ERB
   def set_eoutvar compiler, eoutvar = '_erbout'
     super
 
-    compiler.pre_cmd = ["#{eoutvar} ||= ''"]
+    compiler.pre_cmd = ["#{eoutvar} ||= +''"]
   end
 
 end


yuki24:
If it's already fixed, I think we could always run the entire test suite with `RUBYOPT=--enable-frozen_string_literal` specified.

What do you think? @hsbt @aycabta

deivid-rodriguez:
It's fixed but only on a branch that has not been incorporated to master. Maybe it has other problems? :man_shrugging:. Something to be noted is that this patch does not pass CI with string literals enabled under ruby 2.3 for some reason.

esparta:
It's an interesting problem here. Checking the logs in Travis CI, the problem is not `rdoc` itself but `ERB` who, apparently is not frozen-string complaint in ruby 2.3:

Test file:

```ruby
# frozen_string_literal: true

require 'erb'

x = 42

template = ERB.new <<-EOF
  The value of x is: <%= x %>
EOF
puts template.result(binding)

```

Ruby 2.3:
```bash
$ ruby --version
ruby 2.3.8p459 (2018-10-18 revision 65136) [x86_64-linux]

$ ruby erb_testing.rb
  The value of x is: 42
$ RUBYOPT=--enable-frozen_string_literal ruby erb_testing.rb
(erb):1:in `concat': can't modify frozen String (RuntimeError)
        from (erb):1:in `<main>'
        from /usr/local/lib/ruby/2.3.0/erb.rb:864:in `eval'
        from /usr/local/lib/ruby/2.3.0/erb.rb:864:in `result'
        from erb_testing.rb:9:in `<main>'
```

Meaning while in Ruby 2.4 and later:

```bash
$ ruby --version
ruby 2.4.6p354 (2019-04-01 revision 67394) [x86_64-linux]
$ ruby erb_testing.rb 
  The value of x is: 42
root@65e3c0240197:/src# RUBYOPT=--enable-frozen_string_literal ruby erb_testing.rb
  The value of x is: 42
```

I was not able to test the same in the minimal version supported(ruby 2.2) because it has no way to enable frozen string:

```
$ ruby --version
ruby 2.2.10p489 (2018-03-28 revision 63023) [x86_64-linux]
$ ruby erb_testing.rb 
  The value of x is: 42
$ RUBYOPT=--enable-frozen_string_literal ruby erb_testing.rb
ruby: warning: unknown argument for --enable: `frozen_string_literal'
  The value of x is: 42
```

IMO, having `RUBYOPT=--enable-frozen_string_literal` in ruby 2.3 is not a real use case because it would break not only `rdoc` but (probably) must of the apps running in that version.

deivid-rodriguez:
Since 2.3 support has been dropped, I rebased this PR and it's now green.

ioquatix:
LGTM.

PR #714, opened 2019-04-30
appveyor.yml - correct case in paths, use trunk

MSP-Greg:
This can cause very odd issues, seemingly coming out of nowhere.  I had the mistake in some repos also...

2nd commit uses trunk.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,13 @@
 ---
+image: Visual Studio 2017
+
+clone_depth: 50
+
 init:
-  # Set UTF-8 to use Unicode property because the defualt encoding on AppVeyor is Encoding::IBM437
+  # Set UTF-8 to use Unicode property because the default encoding on AppVeyor is Encoding::IBM437
   - set RUBYOPT=-EUTF-8
   # To avoid duplicated executables in PATH, see https://github.com/ruby/spec/pull/468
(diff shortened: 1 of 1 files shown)


MSP-Greg:
Test times should be much faster, as Appveyor does not set a default clone depth, so the `git clone` was grabbing everything.  I set it to match Travis, which is 50.

Sorry about the extra tests, this is passing in my fork (multiple times), but 2.6 isn't passing here.  I also got interrupted when I wanted to merge two commits, didn't, and pushed...

I'll raise an issue with Appveyor about 2.6.

MSP-Greg:
Done.  Forgive me, might I inquire why?

aycabta:
Unfortunately, AppVeyor doesn't support 2.6 yet...

MSP-Greg:
@aycabta

Appveyor does support 2.6.  It is an Appveyor issue, but the problem is only with older (?) VS images.  The image can be set in the UI, or in the yaml.  I added a newer image to the yaml, and it works fine (2.6.1)

See
https://ci.appveyor.com/project/ruby/rdoc/builds/24219700/job/fc8725pb2fc6hv2m
and
https://ci.appveyor.com/project/ruby/rdoc/builds/24219700/job/u95t56kk7aksf5y1#L98

They will updating 2.4, 2.5, & 2.6 sometime in the near future.

JFYI, you may (from time to time) find CI Ruby version summaries I run on CRON sessions helpful:

https://travis-ci.org/MSP-Greg/travis-ruby
https://ci.appveyor.com/project/MSP-Greg/appveyor-ruby

https://travis-ci.org/MSP-Greg/travis-ruby

PR #723, opened 2019-06-09
Add --load-ri-dir to generate other format

aycabta:


zenspider:
I've resolved conflicts locally and pushed. Nothing more.

zenspider:
This is currently failing because your tests use `capture_io` but test-unit provides no such method. You might want to steal a copy from minitest and put it in test_helper or something.

PR #726, opened 2019-06-16
Skip test option

sriedel:
A surprising amount of gems include their test suites in the gem distribution (out of 185 gems in my current gem set, 78 include a first-level 'spec' or 'test' directory. This can increase the number of files to process by quite a bit.

This proposed change will not gather files to process in under directories called 'spec' or 'test', unless the --no-skipping-tests command line option is supplied.

This feature is similar to --exclude, however will work regardless if the current directory includes a .rdoc_options file. 

Alternatively, a feature could be introduced that loads and parses an .rdoc_options file from the users home directory before such a file is looked for and parsed from the current directory.

### DIFF
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -339,6 +339,10 @@ class RDoc::Options
 
   attr_reader :visibility
 
+  ##
+  # Indicates if files of test suites should be skipped
+  attr_reader :skip_tests
+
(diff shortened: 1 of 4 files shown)


sriedel:
Hmm, not really sure whats up with the failing rake task. Could you give me some pointers as to what may be going wrong? The rake command works fine locally, even after a 'rake clean'.

PR #747, opened 2019-10-25
Make `(#methodname)` a link

nobu:
https://twitter.com/schneems/status/1187074925611376642
> The links at the top of base64.rb are broken in the class docs. I think there needs to be a space after the `(` but before the `#` so `(#encode64` would become `( #encode64`.


schneems:
Thank you nobu!

aycabta:
Great! Works perfectly!

PR #755, opened 2020-01-01
Update link to GFM specification

pocke:
This pull request will update the GitHub Flavored Markdown specification link.
The old link looks expired. It redirects to https://help.github.com/en/github/writing-on-github, but the destination is not the specification. It looks "how to use GitHub".

---

The original idea is https://github.com/rurema/doctree/pull/1474 (Japanese issue), by @sho-h. Thanks.


aycabta:
Thank you, @pocke and @sho-h!

PR #764, opened 2020-04-14
Process crossref before tidylink

aycabta:
The crossref must be linked before tidylink because `Klass.method[:sym]` will be processed as a tidylink first and will be broken.


natematykiewicz:
Does this fix the "options" link on https://ruby-doc.org/stdlib-2.7.0/libdoc/net/ftp/rdoc/Net/FTP.html#method-c-new on the "ssl" box of the `::new` method?

aycabta:
No, I'm trying to fix the problem. It's complex but I think I can provide a solution.

Issue #759, opened 2020-03-31
Reduce the number of useless new lines in html doc


benoittgt:
Hello

With `rdoc` 6.2.1 and the ruby code:
```
class Shape
  def initialize(polyline)
  end
end
```

I get an html with a lot of extra lines and trailing whitespace.

<img width="640" alt="Capture d’écran 2020-03-31 à 10 26 05" src="https://user-images.githubusercontent.com/8417720/78004187-0fe22e00-733a-11ea-999d-492337d4d282.png">

Is it something that can be fixed? 

I can look at it

benoittgt:
I looked a it a little bit. It's probably related to ERB because [the template](https://github.com/ruby/rdoc/blob/6347dd01f90ebe6813d1fd0d9c76eedb996c3994/lib/rdoc/generator/template/darkfish/class.rhtml) have empty lines but is not responsible of all those extra lines. 

I am on Ruby 2.7 and the [trim mode](https://github.com/ruby/rdoc/blob/0ead78616bbddfbdc4d21a3776df16abc25ab811/lib/rdoc/generator/darkfish.rb#L782) seems ok.

I will keep digging.

benoittgt:
It is more related to an ERB behavior. I don't see an easy way at the moment to do not create empty lines and don't break indentation.

```ruby
require 'erb'

expected = %{
<ul>
  <li>1</li>
  <li>2</li>
  <li>3</li>
</ul>
}

template_unindented = %{
<ul>
<% list = [1,2,3] %>
<% for item in list %>
<% if item %>
<li><%= item %></li>
<% end %>
<% end %>
</ul>
 } # space is needed for correct formating

template_indented = %{
<ul>
  <% list = [1,2,3] %>
  <% for item in list %>
  <% if item %>
  <li><%= item %></li>
  <% end %>
  <% end %>
</ul>
}

templates = [
  {
    name: 'template_unindented',
    trim_mode: '<>',
    content: template_unindented
  },
  {
    name: 'template_indented',
    trim_mode: '<>',
    content: template_indented
  },
  {
    name: 'template_indented',
    trim_mode: '%>',
    content: template_indented
  },
  {
    name: 'template_indented',
    trim_mode: '-',
    content: template_indented
  }
]

templates.each do |template|
  puts "🎲  With trim_mode: #{template[:trim_mode]}, for #{template[:name]}\n\n"
  pp ERB.new(template[:content], trim_mode: template[:trim_mode]).result
  puts
end

[...]

PR #775, opened 2020-07-02
Remove empty lines from html file by using ERB trim_mode flag

benoittgt:
Close: https://github.com/ruby/rdoc/issues/759

As mentioned in https://github.com/ruby/rdoc/issues/759, I saw that `rdoc` is generating extra empty lines. This PR try to reduce the amount of empty lines to have a cleaner html structure. 

The impact of this PR can be seen on `rspec-core` gem documentation : https://github.com/benoittgt/rdoc-rspec-core-no-empty-lines/commit/6b57098b906e55583d81fc59ce2331a6c73896a7

First line of partial templates do not need to have `-%>` at the end, because it breaks the indentation when they are used in parent template.

### DIFF
--- a/lib/rdoc/generator/template/darkfish/_head.rhtml
+++ b/lib/rdoc/generator/template/darkfish/_head.rhtml
@@ -15,9 +15,8 @@
 
 <link href="<%= asset_rel_prefix %>/css/fonts.css" rel="stylesheet">
 <link href="<%= asset_rel_prefix %>/css/rdoc.css" rel="stylesheet">
-<% if @options.template_stylesheets.flatten.any? then %>
-<%   @options.template_stylesheets.flatten.each do |stylesheet| %>
+<%- if @options.template_stylesheets.flatten.any? then -%>
+<%-   @options.template_stylesheets.flatten.each do |stylesheet| -%>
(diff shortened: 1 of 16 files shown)


kbrock:
fwiw/ in erb, I tend to not bother with the `<%-` if it is at the beginning of the line with no indentation. But I tend to always have `-%>` at the end of the line. Noticed you're not using skip at the end of the line. reason for this?

benoittgt:
Hello @kbrock 

Thanks for your kind words. You are completly correct for the `<%-` when it is the begining of the line. I will change that. 
What do you mean by "skip"? :)

Is it related to a comment I make:
> First line of partial templates do not need to have -%> at the end, because it breaks the indentation when they are used in parent template.

aycabta:
Great works! I created https://github.com/ruby/rdoc/pull/760 for this problem after https://github.com/ruby/rdoc/issues/759 is issued, and you did investigations and created this Pull Request. I'll release a new version of RDoc later.

kbrock:
@benoittgt adding `-%>` ignores the trailing `"\n"`
So it is easier to not add too much white space

glad this got merged

PR #776, opened 2020-07-07
Fix failing test by adding gettext as a development dependency

dorianmariefr:
The failing test was:

```
========================================================================================
rdoc/test/rdoc/test_rdoc_i18n_locale.rb:35:in `rescue in test_load_existent_po'
Omission: gettext gem is not found [test_load_existent_po(TestRDocI18nLocale)]
========================================================================================
```


kadirselcuk (approved):

aycabta:
Good catch!

PR #785, opened 2021-01-08
Update Rdoc.css sidebar panel.

metacritical:
Updates css so the sidebar look like a panel instead of looking like chopped edges.

Minor css added to Panel : 
Screenshot 2021- 01-08 at 10 27 55 PM

normally it looks like : 
Screenshot 2021-01-08 at 10 31 07 PM


aycabta:
Good catch!

PR #795, opened 2021-03-11
Consider comments within DSL

mtsmfm:
`ActiveSupport::Concern` has `included` DSL and we may define singleton methods in it.

https://github.com/rails/rails/blob/168ddaa08a63cd956bb7c3ba10be1a7ae36d4ee2/activerecord/lib/active_record/core.rb#L9-L20

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -1867,11 +1867,11 @@ class RDoc::Parser::Ruby < RDoc::Parser
 
         when 'end' then
           nest -= 1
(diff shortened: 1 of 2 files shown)


zzak:
@aycabta Any ideas? :bow:

aycabta:
This patch has a critical bug. An "end" that does not belong to the class does not change the visibility but it overrides it. I issued https://github.com/ruby/rdoc/pull/819, please check it out.

mtsmfm:
@aycabta Thank you for pointing out. I close this PR.

PR #810, opened 2021-05-14
Follow-up rubygems

nobu:
Now latest Rubygems uses test-unit assertions instead of minitest.


olleolleolle (approved):
This test alias addition looks fine. I guess it exists with those names in some other part (of Ruby).

PR #813, opened 2021-05-17
lib/rdoc/rdoc.rb: Allow RDoc::Options and Symbol in .rdoc_options

nobu:
As `Psych.load` calls `safe_load` now, need to permit these classes explicitly to load.

### DIFF
--- a/lib/rdoc.rb
+++ b/lib/rdoc.rb
@@ -118,6 +118,8 @@ module RDoc
     ensure
       require 'yaml'
     end
+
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
Good repair!

(I guess having an uppercase instance variable is better than a non-frozen constant.)

aycabta:
This problem is fixed by https://github.com/ruby/rdoc/commit/ebe185c8775b2afe844eb3da6fa78adaa79e29a4.

PR #815, opened 2021-06-18
Rwrite test-case for rubygems_hook without Gem::TestCase

hsbt:


test/rdoc/test_rdoc_rubygems_hook.rb
```diff
@@ -1,30 +1,30 @@
...
       s.rdoc_options = %w[--main MyTitle]
       s.extra_rdoc_files = %w[README]
     end
+    @tempdir = Dir.mktmpdir("test_rubygems_", File.expand_path("tmp"))
 
-    write_file File.join(@tempdir, 'lib', 'a.rb')
-    write_file File.join(@tempdir, 'README')
+    @a.instance_variable_set(:@doc_dir, File.join(@tempdir, "doc"))
+    @a.instance_variable_set(:@gem_dir, File.join(@tempdir, "a-2"))
+    @a.instance_variable_set(:@full_gem_path, File.join(@tempdir, "a-2"))
```
nobu, line 24:
```suggestion
    @a.loaded_from = File.join(@tempdir, 'a-2', 'a-2.gemspec')

```

test/rdoc/test_rdoc_rubygems_hook.rb
```diff
@@ -33,8 +33,10 @@ def setup
     rescue Gem::DocumentError => e
       skip e.message
     end
+  end
```
nobu, line 36:
```suggestion
    @old_ui = Gem::DefaultUserInteraction.ui
    Gem::DefaultUserInteraction.ui = Gem::SilentUI.new
  end
```

test/rdoc/test_rdoc_rubygems_hook.rb
```diff
@@ -33,8 +33,10 @@ def setup
...
 
-    Gem.configuration[:rdoc] = nil
+  def teardown
```
nobu, line 38:
```suggestion
  def teardown
    Gem::DefaultUserInteraction.ui = @old_ui
```

nobu (commented):
Suppress outputs from the hook.

test/rdoc/test_rdoc_rubygems_hook.rb
```diff
@@ -1,40 +1,46 @@
...
+    FileUtils.touch   File.join(@tempdir, 'a-2', 'lib', 'a.rb')
+    FileUtils.touch   File.join(@tempdir, 'a-2', 'README')
 
     @hook = RDoc::RubygemsHook.new @a
 
     begin
       RDoc::RubygemsHook.load_rdoc
     rescue Gem::DocumentError => e
-      skip e.message
+      pend e.message
```
olleolleolle, line 35: This is an interesting change, would `pend` be interchangeable with `omit` here?

olleolleolle: Ah, I was late to the conversation, sorry about that.

hsbt: I'm not sure what differences `pend` and `omit`.

@kou Do you have any opinion about them?

kou: `omit` is suitable for this case.
`omit` for omitting the test because the test isn't needed to be ran.
`pend` for omitting the test temporary. The test should be fixed later.

kou: Generally, `pend` is used with a block.

If the given block is failed, it's expected for `pend`. The test is marked as "pending".

If the given block is not failed, it's unexpected for `pend`. The test is marked as "failed".

```ruby
require "test-unit"

class T < Test::Unit::TestCase
  test "pend: pass" do
    pend "XXX" do
      assert_equal("AAA", "AAA")
    end
  end

  test "pend: failure" do
    pend "XXX" do
      assert_equal("AAA", "BBB")
    end
  end
end
```

hsbt: @kou Thanks! I understood them.

PR #819, opened 2021-06-29
Support ActiveSupport::Concern.included

aycabta:


zzak:
@aycabta What do you think about `prepended` and `class_methods`?

e.g.

```
module SomeConcern
  extend ActiveSupport::Concern

  included do
  end

  prepended do
  end

  class_methods do
  end
end
```

PR #821, opened 2021-06-30
Relative loading for easier development

BuonOmo:
This patch makes sure we only load relative code. Hence when coding or
testing rdoc, we'll be sure to always be using the correct code.

Discussion started at https://github.com/ruby/rdoc/pull/817.

---

There is one place I did not change `require` for `require_relative`, that is under `exe/` because I don't know how that is handled by rubygem, and guessed it could be copied. Moreover, there are no related tests.

---

I almost went and created a helper `autoload_relative` method, but I guess I'll let that [for ruby core at some point](https://bugs.ruby-lang.org/issues/15330).

---

The second commit acts as a _test_ for that new behavior, it may feel intrusive. In that case I'll revert that bit

### DIFF
--- a/test/rdoc/test_rdoc_ri_driver.rb
+++ b/test/rdoc/test_rdoc_ri_driver.rb
@@ -32,8 +32,8 @@ class TestRDocRIDriver < RDoc::TestCase
   end
 
   def teardown
-    ENV['RI'] = @orig_ri
-    FileUtils.rm_rf @tmpdir
+    defined?(@orig_ri) and ENV['RI'] = @orig_ri
+    defined?(@tmpdir) and FileUtils.rm_rf @tmpdir
 
     super
   end
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,6 @@
-$:.unshift File.expand_path 'lib'
-require 'rdoc/task'
+# frozen_string_literal: true
+
+require_relative 'lib/rdoc/task'
 require 'bundler/gem_tasks'
 require 'rake/testtask'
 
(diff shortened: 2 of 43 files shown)


test/rdoc/test_rdoc_ri_driver.rb
```diff
@@ -32,8 +32,8 @@ def setup
   end
 
   def teardown
-    ENV['RI'] = @orig_ri
-    FileUtils.rm_rf @tmpdir
+    defined?(@orig_ri) and ENV['RI'] = @orig_ri
```
BuonOmo, line 35: when my tests were failing, the teardown here was failing as well, hence the guard.

nobu: Where your tests failed, `mkdir_p`?

BuonOmo: @nobu I don't remember nor I simply reproduce. However I still think keeping that is a good idea, if you ever fail at some point in the setup, you don't want to fail as well here.

Moreover the `RDoc::TestCase` class already uses that pattern:

```
  ##
  # Abstract test-case teardown

  def teardown
    ENV["HOME"] = @orig_home if defined?(@orig_home)

    super
  end
```

If you find that it is superfluous I can still remove this bit.

nobu: I see, just curious.

nobu:
Needs nobu/rdoc@02fa391c5a4dd7b1b59ee1d81d9dfc2632b26183.

BuonOmo:
@nobu shall I apply the patch you linked, or is in your hands?

nobu:
Please apply.

BuonOmo:
@nobu done :)

PR #833, opened 2021-09-01
Try to load WEBrick for testing

voxik:
This is in similar vein Rake is optionally loaded in:

https://github.com/ruby/rdoc/blob/5437418baf8eeb984fe7e74b181d3da1b174d0f7/test/rdoc/test_rdoc_task.rb#L3-L6

### DIFF
--- a/test/rdoc/test_rdoc_servlet.rb
+++ b/test/rdoc/test_rdoc_servlet.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 require_relative 'helper'
+begin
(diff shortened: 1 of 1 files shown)


voxik:
Actually, both would deserve if there was some message provided when the require fails :thinking: similarly to:

https://github.com/ruby/rdoc/blob/5437418baf8eeb984fe7e74b181d3da1b174d0f7/test/rdoc/test_rdoc_i18n_locale.rb#L31-L36

PR #837, opened 2021-09-10
Convert upper case copyright and registered marks

nobu:


olleolleolle (approved):
This adds the case-insensitive check to the regular expression, and extends the tests.

PR #841, opened 2021-09-20
RDoc::Options support for serializing Regexp

flavorjones:
## Description of the problem

Commits f0141d4 and 58c0804 introduced `YAML.safe_load` for reading in serialized `RDoc::Options` from an `.rdoc_options` file.

The only permitted classes are `RDoc::Options` and `Symbol`. However, if someone uses the `--write-options` option, the `exclude` property is serialized as a `Regexp` and cannot be deserialized:

```sh
$ rdoc --write-options --exclude=ext/java

$ cat .rdoc_options
--- !ruby/object:RDoc::Options
encoding: UTF-8
static_path: []
rdoc_include:
- "."
- "/home/redacted/foo"
charset: UTF-8
exclude: !ruby/regexp /~\z|\.orig\z|\.rej\z|\.bak\z|\.gemspec\z|(?-mix:ext\/java)/
hyperlink_all: false
line_numbers: false
locale:
locale_dir: locale
locale_name:
main_page:
markup: rdoc
output_decoration: true
page_dir:
show_hash: false
tab_width: 8
template_stylesheets: []
title:
visibility: :protected
webcvs:

$ rdoc lib ext
uh-oh! RDoc had a problem:
Tried to load unspecified class: Regexp

```

The exception in this case is a `Psych::DisallowedClass` from `YAML.safe_load`.

## The fix

This PR first extracts the set of permitted classes into an array constant in `RDoc::Options` and then adds `Regexp` to that constant.

Test coverage has been added by changing `test_write_options` to invoke `#finish` on the options before writing them, for a true round-trip which coerces the `@exclude` ivar into a `Regexp`. `test_write_options` now tests equality of a specific set of attributes, rather than asserting equality on the entire `Options` object, because deserialization does not include some attributes (like `@op_dir`) that are inspected by `#==`.

### DIFF
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -121,7 +121,7 @@ class RDoc::Options
 
   ##
   # Classes that are permitted to be loaded by load_options
-  PERMITTED_CLASSES = [RDoc::Options, Symbol]
+  PERMITTED_CLASSES = [RDoc::Options, Symbol, Regexp]
 
   ##
   # Option validator for OptionParser that matches a directory that exists on
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -120,6 +120,10 @@ class RDoc::Options
   ]
 
   ##
+  # Classes that are permitted to be loaded by load_options
+  PERMITTED_CLASSES = [RDoc::Options, Symbol]
+
+  ##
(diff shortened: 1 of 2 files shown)


flavorjones:
Note that #839 seems to be addressing this issue in a different way. /cc @nobu

PR #842, opened 2021-09-23
feature: Render mixed-in methods and constants with `--embed-mixins`

flavorjones:
When `--embed-mixins` option is set:

- methods from an `extend`ed module are documented as singleton methods
- attrs from an `extend`ed module are documented as class attributes
- methods from an `include`ed module are documented as instance methods
- attrs from an `include`ed module are documented as instance attributes
- constants from an `include`ed module are documented

Sections are created when needed, and Darkfish's template annotates
each of these mixed-in CodeObjects.

This feature is inspired by Yard's option of the same name.

---

Example class:

```ruby
module IncMod
  # :section: Things That Were Included

  # incmod attribute
  attr :incmod_attr

  # constant via mixin
  INCMOD_CONST = "incmod_const"

  # instance method via mixin
  def incmod_method
    "incmod_method"
  end
end

module ExtMod
  # :section: Things That Were Extended

  # extmod attribute
  attr :extmod_attr

  # class method via mixin
  def extmod_method
    "extmod_method"
  end
end

class Foo
  include IncMod
  extend ExtMod
end
```

Currently this renders like:

Screenshot from 2021-09-23 02-01-18

With `--embed-mixins` it renders like:

Screenshot from 2021-09-23 02-01-38

### DIFF
--- a/lib/rdoc/code_object.rb
+++ b/lib/rdoc/code_object.rb
@@ -116,6 +116,7 @@ class RDoc::CodeObject
     @full_name        = nil
     @store            = nil
     @track_visibility = true
+    @mixin_from       = nil
 
     initialize_visibility
   end
--- a/lib/rdoc/code_object.rb
+++ b/lib/rdoc/code_object.rb
@@ -97,6 +97,11 @@ class RDoc::CodeObject
   attr_accessor :viewer
 
   ##
+  # When mixed-in to a class, this points to the Context in which it was originally defined.
+
+  attr_accessor :mixin_from
+
+  ##
   # Creates a new CodeObject that will document itself and its children
 
   def initialize
--- a/test/rdoc/test_rdoc_class_module.rb
+++ b/test/rdoc/test_rdoc_class_module.rb
@@ -1500,4 +1500,132 @@ class TestRDocClassModule < XrefTestCase
     assert_equal [a, c], @c1.extends
   end
 
+  class TestRDocClassModuleMixins < XrefTestCase
+    def setup
+      super
+
+      klass_tl = @store.add_file("klass.rb")
+      @klass = klass_tl.add_class(RDoc::NormalClass, "Klass")
+
+      incmod_tl = @store.add_file("incmod.rb")
+      @incmod = incmod_tl.add_module(RDoc::NormalModule, "Incmod")
+
+      incmod_const = @incmod.add_constant(RDoc::Constant.new("INCMOD_CONST_WITHOUT_A_SECTION", nil, ""))
+      incmod_const = @incmod.add_constant(RDoc::Constant.new("INCMOD_CONST", nil, ""))
+      incmod_const.section = @incmod.add_section("Incmod const section")
+
+      incmod_method = @incmod.add_method(RDoc::AnyMethod.new(nil, "incmod_method_without_a_section"))
+      incmod_method = @incmod.add_method(RDoc::AnyMethod.new(nil, "incmod_method"))
+      incmod_method.section = @incmod.add_section("Incmod method section")
+
+      incmod_attr = @incmod.add_attribute(RDoc::Attr.new(nil, "incmod_attr_without_a_section", "RW", ""))
+      incmod_attr = @incmod.add_attribute(RDoc::Attr.new(nil, "incmod_attr", "RW", ""))
+      incmod_attr.section = @incmod.add_section("Incmod attr section")
+
+      incmod_private_method = @incmod.add_method(RDoc::AnyMethod.new(nil, "incmod_private_method"))
+      incmod_private_method.visibility = :private
+
+      extmod_tl = @store.add_file("extmod.rb")
+      @extmod = extmod_tl.add_module(RDoc::NormalModule, "Extmod")
+
+      extmod_method = @extmod.add_method(RDoc::AnyMethod.new(nil, "extmod_method_without_a_section"))
+      extmod_method = @extmod.add_method(RDoc::AnyMethod.new(nil, "extmod_method"))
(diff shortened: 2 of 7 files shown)


flavorjones:
@aycabta I'm curious if you have feedback on this feature? It's off by default but I think would be a real improvement to standard-library classes like Array that inherit significant functionality from Enumerable.

flavorjones:
@aycabta This feature is live at https://nokogiri.org/rdoc/Nokogiri/XML/Node.html#method-i-at_css if you'd like to see what it looks like. Is this something you would consider merging?

flavorjones:
Rebased off v6.5.0.

I still think this is a valuable feature to consider. It's been in use at https://nokogiri.org/rdoc/ for the past year, see https://nokogiri.org/rdoc/Nokogiri/XML/Node.html for example.

flavorjones:
CI is broken until https://github.com/ruby/rdoc/pull/951 is merged.

flavorjones:
Rebased again off master. @aycabta @nobu would you consider merging this feature?

flavorjones:
I've rebased again onto master. I still think this would be a valuable feature to merge. @hsbt @drbrain I would appreciate any feedback you have for me.

flavorjones:
@hsbt If you get some time to review this, I still think it's a good feature, and it's been live on the nokogiri.org site for several years now.

lib/rdoc/class_module.rb
```diff
@@ -798,4 +799,41 @@ def update_extends
...
+  end
+
+  private
+
+  def prepare_to_embed(code_object, singleton=false)
+    code_object = code_object.dup
+    code_object.mixin_from = code_object.parent
+    code_object.singleton = true if singleton
+    set_current_section(code_object.section.title, code_object.section.comment)
+    self.visibility = code_object.visibility
```
st0012, line 836: It took me a while to understand why we need this. Let's document the reason down:

```suggestion
    # add_method and add_attribute will reassign self's visibility back to the method/attribute
    # so we need to sync self's visibility with the object's to properly retain that information
    self.visibility = code_object.visibility
```

lib/rdoc/code_object.rb
```diff
@@ -96,6 +96,11 @@ class RDoc::CodeObject
 
   attr_accessor :viewer
 
+  ##
+  # When mixed-in to a class, this points to the Context in which it was originally defined.
+
+  attr_accessor :mixin_from
```
st0012, line 102: Let's follow the convention of adding it to `initialize`?

st0012 (commented):
Left some small suggestions, but overall looks good to me.
How does it affect documentation displayed by `ri` though?

flavorjones:
@st0012 Thank you for your review!

> How does it affect documentation displayed by ri though?

I hadn't even considered ri as a target, since I don't usually use it. I will explore what's necessary to extend the ri content as well.

lib/rdoc/code_object.rb
flavorjones: Done.

flavorjones: Done.

test/rdoc/test_rdoc_class_module.rb
```diff
@@ -1500,4 +1500,132 @@ def test_update_extends_with_colons
...
+      assert(instance_method, "instance method from included mixin should be present")
+      refute(instance_method.singleton)
+      assert_equal(@incmod, instance_method.mixin_from)
+
+      instance_attr = @klass.attributes.find { |a| a.name == "incmod_attr" }
+      assert(instance_attr, "instance attr from included mixin should be present")
+      refute(instance_attr.singleton)
+      assert_equal(@incmod, instance_attr.mixin_from)
+
+      refute(@klass.method_list.find { |m| m.name == "incmod_private_method" })
```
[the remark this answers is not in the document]
flavorjones: Done! See commit on HEAD.

flavorjones:
@st0012 So `ri` is working out of the box with this.

If I generate the ri files with `rdoc --embed-mixins --format=ri` then the resulting behavior is what I'd expect, which is that it shows the mixed-in methods under all of the classes.

```
$ ri at_css
= .at_css

(from /home/flavorjones/code/oss/nokogiri/ri)
=== Implementation from Node
------------------------------------------------------------------------
  at_css(*rules, [namespace-bindings, custom-pseudo-class])

------------------------------------------------------------------------

Search this object for CSS rules, and return only the first match. rules
must be one or more CSS selectors.

See Searchable#css for more information.


(from /home/flavorjones/code/oss/nokogiri/ri)
=== Implementation from NodeSet
------------------------------------------------------------------------
  at_css(*rules, [namespace-bindings, custom-pseudo-class])

------------------------------------------------------------------------

Search this object for CSS rules, and return only the first match. rules
must be one or more CSS selectors.

See Searchable#css for more information.


(from /home/flavorjones/code/oss/nokogiri/ri)
=== Implementation from Searchable
------------------------------------------------------------------------
  at_css(*rules, [namespace-bindings, custom-pseudo-class])

------------------------------------------------------------------------

Search this object for CSS rules, and return only the first match. rules
must be one or more CSS selectors.

See Searchable#css for more information.


```

flavorjones:
No sweat, @st0012! Let me know if I can be helpful with anything.

flavorjones:
Rebased onto current `master` and force-pushed.

flavorjones:
Rebased onto current `master` and force-pushed.

st0012 (approved):
Tested with RDoc and I think it works well. Thank you for this great addition 🎉

flavorjones:
woo hoo! thank you, @st0012!

PR #843, opened 2021-09-29
Add Truffleruby head to CI

gogainda:


lib/rdoc/markup/to_html.rb
```diff
@@ -426,7 +426,10 @@ def list_end_for(list_type)
 
   def parseable? text
     verbose, $VERBOSE = $VERBOSE, nil
-    eval("BEGIN {return true}\n#{text}")
+    catch(:valid) do
```
gogainda, line 429: this is taken from how irb checks for parseable codehttps://github.com/ruby/irb/pull/134/files

hsbt:
Added truffleruby-head at https://github.com/ruby/rdoc/pull/977

PR #849, opened 2021-11-01
Fix RDoc::Markup::Document#accept to do what it says

gettalong:
Although the parts are not iterated in `Document#accept`, it seems they are iterated somewhere else according to the tests 

Not sure if the change is correct or if the method comment needs to be changed...

### DIFF
--- a/lib/rdoc/markup/document.rb
+++ b/lib/rdoc/markup/document.rb
@@ -60,12 +60,14 @@ class RDoc::Markup::Document
   end
 
   ##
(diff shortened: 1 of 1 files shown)


nobu:
`accept_document` should do it.

gettalong:
Ah, okay, thanks for the information! Is there some information on what writing a visitor class entails?

And maybe the method comment should be updated to make it clear that the parts of the document are **not** automatically run through the visitor.

PR #858, opened 2022-01-03
Windows pager support

YO4:
I want to use PAGER and colourful documents on windows too.

with Git for Windows installed,

```
set PAGER="c:\Program Files\Git\usr\bin\less.exe" -r
set RI=--format ansi
```

or with RubyInstaller's devkit,

```
set PAGER=ridk.cmd exec less -r
set RI=--format ansi
```

also, this affects IRB's Alt+d document browser.
(more.com not clears screen, so some corrupted display appears because of overwriting irb dialogs)

### DIFF
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -1544,7 +1544,7 @@ or the PAGER environment variable.
 
     jruby = RUBY_ENGINE == 'jruby'
 
-    pagers = [ENV['RI_PAGER'], ENV['PAGER'], 'pager', 'less', 'more', 'more.com']
+    pagers = [ENV['RI_PAGER'], ENV['PAGER'], 'pager', 'less', 'more']
 
     pagers.compact.uniq.each do |pager|
       next unless pager
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -1150,7 +1150,7 @@ or the PAGER environment variable.
   # Is +file+ in ENV['PATH']?
 
   def in_path? file
-    extensions = [''].union(RbConfig::CONFIG["EXECUTABLE_EXTS"]&.split || RbConfig::CONFIG["EXEEXT"].to_a)
+    extensions = [''].union(RbConfig::CONFIG['EXECUTABLE_EXTS']&.split || [RbConfig::CONFIG['EXEEXT' || '']])
     if absolute_path?(file)
       extensions.any? do |ext|
         File.exist?(file + ext)
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -1150,7 +1150,15 @@ or the PAGER environment variable.
   # Is +file+ in ENV['PATH']?
 
   def in_path? file
-    return true if file =~ %r%\A/% and File.exist? file
+    if File.respond_to?(:absolute_path?)
+      return true if File.absolute_path?(file) and File.exist?(file)
+    else
+      if RbConfig::CONFIG['target_os'] =~ /mswin|mingw/
+        return true if file =~ %r%\A[A-Za-z]:[/\\]% and File.exist? file
+      else
+        return true if file =~ %r%\A/% and File.exist? file
+      end
+    end
 
     ENV['PATH'].split(File::PATH_SEPARATOR).any? do |path|
       File.exist? File.join(path, file)
@@ -1526,7 +1534,11 @@ or the PAGER environment variable.
     pagers.compact.uniq.each do |pager|
       next unless pager
 
-      pager_cmd = pager.split(' ').first
+      if RbConfig::CONFIG['target_os'] =~ /mswin|mingw/
+        pager_cmd = pager.sub('Program Files', "\0").split(' ').first.sub("\0", 'Program Files').gsub('"', '')
+      else
+        pager_cmd = pager.split(' ').first
+      end
 
       next unless in_path? pager_cmd
 
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -1534,7 +1534,7 @@ or the PAGER environment variable.
     pagers.compact.uniq.each do |pager|
       next unless pager
 
-      pager_cmd = pager[/\A\s*(?:"\K[^"]+(?=")|\K\S+)/]}
+      pager_cmd = pager[/\A\s*(?:"\K[^"]+(?=")|\K\S+)/]
 
       next unless in_path? pager_cmd
 
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -1544,7 +1544,7 @@ or the PAGER environment variable.
 
     jruby = RUBY_ENGINE == 'jruby'
 
-    pagers = [ENV['RI_PAGER'], ENV['PAGER'], 'pager', 'less', 'more']
+    pagers = [ENV['RI_PAGER'], ENV['PAGER'], 'pager', 'less', 'more', 'more.com']
 
     pagers.compact.uniq.each do |pager|
       next unless pager
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -1150,7 +1150,7 @@ or the PAGER environment variable.
   # Is +file+ in ENV['PATH']?
 
   def in_path? file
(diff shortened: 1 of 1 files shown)


lib/rdoc/ri/driver.rb
```diff
@@ -1523,12 +1531,16 @@ def setup_pager
...
 
     pagers.compact.uniq.each do |pager|
       next unless pager
 
-      pager_cmd = pager.split(' ').first
+      if RbConfig::CONFIG['target_os'] =~ /mswin|mingw/
+        pager_cmd = pager.sub('Program Files', "\0").split(' ').first.sub("\0", 'Program Files').gsub('"', '')
+      else
+        pager_cmd = pager.split(' ').first
+      end
```
nobu, line 1543: It seems too fragile.

```suggestion
      pager_cmd = pager[/\A\s*(?:"\K[^"]+(?=")|\K\S+)/]}
```

YO4:
There are still pitfalls,
need to set pager_name with extension(.exe)
and with wrong settings, more.com is used silentry.

Nevertheless, we can use pager on windows.
thank you.

nobu:
The extension ".exe" should able to be omitted.

YO4:
Here it is, but I feel that code is too windows specific especially in RDoc codespace.
RDoc::RI::Driver.in_path? does similar to famous command `which` (or not very well known `where` on windows).
I want to use a method like FileTest.which if there is.

lib/rdoc/ri/driver.rb
```diff
@@ -1152,13 +1152,36 @@ def interactive
   # Is +file+ in ENV['PATH']?
 
   def in_path? file
-    return true if file =~ %r%\A/% and File.exist? file
+    extensions = [''].concat((windows? ? ENV['PATHEXT']&.split(File::PATH_SEPARATOR) : nil).to_a)
```
nobu, line 1155: `IO.popen` supports only executable by itself (".exe", ".com") and by `cmd.exe` (".cmd" and ".bat").
Others in `PATHEXT` by default, ".js", ".vbs", and so on, which need extra registry handling are not supported.

YO4: I'm realizing how difficult it is to take all cases into account...
I took a journey to win32.c and dln_find.c. On windows, IO.popen("commandline", "w") not always uses shell　I understood, is that correct?
It seems to be reasonable to use RbConfig::CONFIG["EXECUTABLE_EXTS"] instead of ENV['PATHEXT'], It's been prepared that way. I try it.

lib/rdoc/ri/driver.rb
```diff
@@ -1152,13 +1152,36 @@ def interactive
   # Is +file+ in ENV['PATH']?
 
   def in_path? file
-    return true if file =~ %r%\A/% and File.exist? file
+    extensions = [''].union(RbConfig::CONFIG["EXECUTABLE_EXTS"]&.split || RbConfig::CONFIG["EXEEXT"].to_a)
```
nobu, line 1155: `RbConfig::CONFIG["EXECUTABLE_EXTS"]` is set since 1.9.2.
`RbConfig::CONFIG["EXEEXT"]` is the single suffix to be built, not a list.

YO4: `RbConfig::CONFIG["EXECUTABLE_EXTS]` looks good enough and stable, I put `RbConfig::CONFIG["EXEEXT"]` there just in case.
But I tried ruby implementations other than MRI, JRuby does not have `RbConfig::CONFIG["EXECUTABLE_EXTS]`.
And I found `String` does not have `to_a`, it translate `nil` into `[]` if `RbConfig::CONFIG` not have `"EXEEXT"`. But it's broken.
`[RbConfig::CONFIG["EXEEXT"]]` insted of `RbConfig::CONFIG["EXEEXT"].to_a` fix this.

As another point of view, `RDoc::RI::Driver#find_pager_jruby`, that is used to create pipe for JRuby, seems to execute childprocess that has filename extension `.exe` `.com` `.bat` `.cmd`. But it can omit filename extension only `.exe`. Even if in_path? compensate `.com`, JRuby not yet found `more.com` when `set PAGER=more`. `set PAGER=more.com` is required.
I'm not familiar to Java and JRuby, I don't know which mechanism is doing this.

I think changing commandline filename in PAGER introduces complexity and I don't want to fall into the quote nightmare.
For JRuby on Windows, it's a reasonable choice to set PAGER to the filename with extension other than `.exe`. 
And add `more.com` fallback, it supports for most use cases.

lib/rdoc/ri/driver.rb
```diff
@@ -1152,13 +1152,36 @@ def interactive
   # Is +file+ in ENV['PATH']?
 
   def in_path? file
-    return true if file =~ %r%\A/% and File.exist? file
+    extensions = [''].union(RbConfig::CONFIG['EXECUTABLE_EXTS']&.split || [RbConfig::CONFIG['EXEEXT' || '']])
```
nobu, line 1155: `'EXEEXT' || ''` is always `'EXEEXT'`.

nobu, line 1155:
```suggestion
    extensions = [RbConfig::CONFIG['EXEEXT']].union(RbConfig::CONFIG['EXECUTABLE_EXTS']&.split(" ") || [])
```

YO4: Your suggestion offers `[".exe", ".com", ".cmd", ".bat"]` on windows but I expect `["", ".exe", ".com", ".cmd", ".bat"]`, works in case PAGER has filename with it's extension.

`['', RbConfig::CONFIG['EXEEXT']].union(RbConfig::CONFIG['EXECUTABLE_EXTS']&.split(" ") || [])` may work but I'm afraid that 'extension' is not a good name in this situation.
`additional_suffixes` or `candidate_suffixes` are my idea, should we change the name `extensions`?

YO4:
Sorry for the late reply and thank you for taking your valuable time to help me.

I made this pull request with a light heart, but I have changed my mind that this is a case where we need to make a policy in order to be able to work with various PAGER environment variables and to be able to provide a reason if it does not work.

The following topics on path interpretation may have been repeated many times, but I'll summarize them here to give form to my thoughts.

In the Windows environment, I believe that valid strings in the cmd.exe prompt should be accepted as PAGER, but the interpretation of quoting in cmd.exe and the extraction of program names by MRI's system and spawn functions are incompatible.
For example, the following command line might be generated from a PAGER created by concatenating the quoted path and file name

```
c:\>"c:\Program Files\Git\usr\bin\"less.exe -V  
# => less version appear

irb(main):001:0> IO.popen('"c:\Program Files\Git\usr\bin\"less.exe -V', "w")
# => (irb):1:in `spawn': Permission denied - "c:\Program Files\Git\\fnBin\"less.exe -V (Errno::EACCES)
```

It may be possible to interpret the command line and extract the program name in the way cmd.exe does, but it is terribly complicated and I don't think it is possible to do it perfectly except by having cmd.exe do it.
It seems that the current MRI on windows does not always go through the shell when executing external programs, so this would split the execution flow, whereas the previous patches were designed with integration of processing in mind. ...... only if, for example, SHELL is unset and COMSPEC points to cmd.exe.

The current MRI on windows doesn't always seem to go through the shell when executing external programs, so it splits the execution flow, whereas the previous patches were oriented towards integration of processing with other OSs. ...... only if, for example, SHELL is unset and COMSPEC points to cmd.exe.

Of course, there may be cases where PAGER is interpreted in /bin/sh and may be set in that fashion. In my environment, git seems to call PAGER via /bin/sh if the PAGER environment variable is set. It would be better to set it in .gitconfig in this case, though.
If the path is escaped in the Shellwords.escape fashion, we can leave execution of external programs to the MRI routines. Single quotes are used to escape paths, slashes are used as separators and backslashes are used only before spaces... Is that enough to properly distinguish it from the cmd.exe style? Maybe I should have coded it first instead of worrying about it.

I'll try it next time I have time.

nobu:
Eventually, removed the command search in ri/driver.rb, and left it to `IO.popen`.
As for ".com" extension, `IO.popen` (and `system`, `Process.spawn` etc) should search it automatically even in JRuby.

YO4:
Thank you for your response.
I thought the original code was checking in_path to ensure a fallback to the default pager.
I didn't know what kind of platform that requires to check the executable file, so I couldn't make a decision to delete in_path.
Also, I was struck by the depths of the strange and chaotic command line interpretation of cmd.exe.
Thank you again for the long dialogue.

PR #861, opened 2022-01-13
Use readline history when `ri` is running interactive

adam12:
With history support enabled, navigating deeper nested namespaces is a more enjoyable experience.


zenspider (commented):
I think this should go in

PR #862, opened 2022-01-17
Add Ruby 3.1 to test matrix

adam12:
Enable testing with Ruby 3.1 in CI.


adam12:
I just noticed #856. Closing.

PR #867, opened 2022-02-08
Skip parentheses on singleton class declaration

soutaro:
Parsing Ruby code with singleton class declaration with assignment generates a class name with `(`.

This is the example of the source code.

```rb
module DidYouMean
  class << (NameErrorCheckers = Object.new)
  end
end
```

This generates `DidYouMean::(NameErrorCheckers` class, and raises an error during load.

```
.../rdoc/lib/rdoc/store.rb:530:in `block in load_all': end pattern with unmatched parenthesis: /^DidYouMean::(NameErrorCheckers::[^:]+$/ (RegexpError)
```

This PR fixes the issue by ignoring surrounding parentheses.

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -400,6 +400,29 @@ class RDoc::Parser::Ruby < RDoc::Parser
   end
 
   ##
+  # Skip opening parentheses and yield the block.
+  # Skip closing parentheses too when exists.
+
+  def skip_parentheses(&block)
+    left_tk = peek_tk
+
+    if :on_lparen == left_tk[:kind]
+      get_tk
+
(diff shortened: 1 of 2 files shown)


soutaro:
I found the problem with the `rbs annotate` [command](https://github.com/ruby/rbs/pull/881) which calls `RDoc::Store#load_all`. `ri` shows a problem related to this with the following output:

```
$ make rdoc RDOCFLAGS="--visibility=nodoc"
$ ri --no-system -d .ext/rdoc DidYouMean::NameErrorCheckers
Nothing known about DidYouMean::NameErrorCheckers
Did you mean?  DidYouMean::(NameErrorCheckers
               DidYouMean::KeyErrorChecker
```

PR #882, opened 2022-04-16
New line-oriented `.document` file format

nobu:
Introduce the new format similar to the `.gitignore` file by starting with the comment `rdoc.document: 1`.

- one pattern per line
- negative pattern starting with `!`
- escaping space and `#` by a backslash


zenspider (approved):
I like it!

st0012:
Can you explain more on why this change is needed? Do we want to eventually migrate all `.document` to new versions?

PR #898, opened 2022-07-16
Handle operator-type call-seq

BurdetteLamar:


jeremyevans:
Can you please add tests for this behavior change?  I see this doesn't break anything (indicating no tests rely on the old behavior), but it seems like something we would want tests for.

nobu:
Shouldn't `RDoc::AnyMethod` be changed instead?

BurdetteLamar:
> Shouldn't `RDoc::AnyMethod` be changed instead?

I'm only aiming at Ruby-coded methods, and especially those with no call-seq.  A C-coded method with an explicit call-seq, I want to leave alone.

BurdetteLamar:
> Can you please add tests for this behavior change? I see this doesn't break anything (indicating no tests rely on the old behavior), but it seems like something we would want tests for.

Testing added for new method coerce_call_seq.  The method that calls it, parse_method_params_and_body, has no testing to add to.

test/rdoc/test_rdoc_parser_ruby.rb
```diff
@@ -4297,4 +4297,39 @@ class D
     assert_equal 'A::D', a_d.full_name
   end
 
+  COERCIBLE_OPERATORS = %w[ ! != !~ % & * ** + - / < << <= <=> == === =~ > >= >> ^ | ~ ]
```
peterzhu2118, line 4300: You can just use `RDoc::Parser::Ruby::OPERATOR_METHOD_NAMES` instead.

test/rdoc/test_rdoc_parser_ruby.rb
```diff
@@ -4297,4 +4297,39 @@ class D
...
+
+  def test_coerce_call_seq
+    util_parser('')
+    def coercible?(op)
+      COERCIBLE_OPERATORS.include?(op)
+    end
+    def _test_coercion(op, single_arg: true, singleton: false)
```
peterzhu2118, line 4307: This test is too hard to read. There's too much logic in the tests. Please make tests "dumber" and more verbose.

lib/rdoc/parser/ruby.rb
```diff
@@ -1470,10 +1470,27 @@ def parse_method_params_and_body container, single, meth, added_container
...
       end
 
+      coerce_call_seq(meth)
+
       parse_statements container, single, meth
     end
   end
 
+  # Coerce calling sequence if the given method is an operator method.
+  def coerce_call_seq(meth)
```
peterzhu2118, line 1480: This doesn't treat all operators correctly. For example, for the `!` method, it generates `self !` when it should be `!self`. It also doesn't handle the `[]` and `[]=` operators.

peterzhu2118 (changes requested):

test/rdoc/test_rdoc_parser_ruby.rb
BurdetteLamar: Thanks for 'NAMES';  I had violated my own rule for distinguishing between string names and other objects.

test/rdoc/test_rdoc_parser_ruby.rb
BurdetteLamar: Dumbness and verbosity added.

lib/rdoc/parser/ruby.rb
BurdetteLamar: `[]` and `[]=` (also `!`) are out of scope for this PR, which aims only at binary operators.

peterzhu2118: The `~` method also doesn't accept any arguments.

test/rdoc/test_rdoc_parser_ruby.rb
```diff
@@ -4297,4 +4297,50 @@ class D
     assert_equal 'A::D', a_d.full_name
   end
 
+  BINARY_OPERATOR_METHOD_NAMES =
+    %w[ != !~ % & * ** + - / < << <= <=> == === =~ > >= >> ^ | ~ ]
+
+  def test_coerce_call_seq
```
peterzhu2118, line 4303: This is still difficult to read. Can you split them into separate test methods?

test/rdoc/test_rdoc_parser_ruby.rb
peterzhu2118: I mean you can delete this line since it's just a duplicate of `RDoc::Parser::Ruby::OPERATOR_METHOD_NAMES`.

BurdetteLamar: I want to be able to check whether the method and its test agree on the names.  See the first assertion.

lib/rdoc/parser/ruby.rb
BurdetteLamar: Fixed.

test/rdoc/test_rdoc_parser_ruby.rb
BurdetteLamar: I'd rather not split.  I think there's value in seeing them all at once.  Does the reformatting help?

test/rdoc/test_rdoc_parser_ruby.rb
peterzhu2118: But that doesn't really test anything. A test is supposed to execute some code and check that its output matches what you expect. Checking that a constant is equal to a value isn't really testing anything meaningful (it's like testing that `assert_equal(1, 1)`).

test/rdoc/test_rdoc_parser_ruby.rb
peterzhu2118: Well I see four tests here:

One for when the operator is handled.

One for methods that aren't handled.

Multiple arguments to operators isn't handled.

Singleton methods aren't handled.

test/rdoc/test_rdoc_parser_ruby.rb
BurdetteLamar: If I add a new binary op method `@` and add its name to the constant, the test will fail.  That's what I want to check.

test/rdoc/test_rdoc_parser_ruby.rb
BurdetteLamar: Now split into multiple tests.

test/rdoc/test_rdoc_parser_ruby.rb
peterzhu2118: But they could just as easily also add it to this test and make it pass? This kind of things should be caught in reviews instead.

BurdetteLamar: I don't agree.  My point is to discover the error if they don't add it to the test.  Do you think that including this assertion is harmful?  If not, I'd like to retain it.

lib/rdoc/parser/ruby.rb
```diff
@@ -1470,10 +1470,30 @@ def parse_method_params_and_body container, single, meth, added_container
...
       end
 
+      coerce_call_seq(meth)
+
       parse_statements container, single, meth
     end
   end
 
+  # Coerce calling sequence if the given method is a binary operator method.
+  def coerce_call_seq(meth)
```
peterzhu2118, line 1480: We should make this method private and not directly test it, but rather test the encapsulating `parse_method` method, because I could delete the call to `coerce_call_seq` from `parse_method_params_and_body` and all the tests would still pass.

test/rdoc/test_rdoc_parser_ruby.rb
peterzhu2118: I still strongly believe that there's no point in testing constants, and the best way to prevent incorrect changes to these constants is by making it easy to understand for other developers through proper naming and comments, and having PR reviews. But I don't think this is a particularly productive discussion to have, so I'm ok with keeping it.

lib/rdoc/parser/ruby.rb
BurdetteLamar: I see your point, but there's no testing whatsoever for parse_method_params_and_body.  I don't think it should fall to me to take that on.  I'm happy to test the code I've written, as I've done.

test/rdoc/test_rdoc_parser_ruby.rb
BurdetteLamar: I will remove the test for the constant.  Should the test have its own constant, or should it rely on the one in the library?

lib/rdoc/parser/ruby.rb
```diff
@@ -1470,10 +1470,30 @@ def parse_method_params_and_body container, single, meth, added_container
...
+    # Only coerce certain methods.
+    return unless BINARY_OPERATOR_METHOD_NAMES.include?(meth.name)
+    # Ignore if singleton method.
+    return if meth.singleton
```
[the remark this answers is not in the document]
BurdetteLamar: Line the index and unary ops, this is not in scope for this PR.  (But more to come.)

lib/rdoc/parser/ruby.rb
peterzhu2118: I disagree. `coerce_call_seq ` (and arguably `parse_method_params_and_body`) are all private methods (i.e. implementation details) of `RDoc::Parser::Ruby`. Private methods shouldn't be directly tested, therefore it must be tested through `parse_method `.

lib/rdoc/parser/ruby.rb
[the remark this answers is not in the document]
peterzhu2118: Isn't implementing it as simple as just removing this line?

lib/rdoc/parser/ruby.rb
BurdetteLamar: Well, I started out trying to do a small good deed, but it seems that small and good, here, are mutually exclusive.  If there's no objection, I will close this PR and consider it a lesson learned.

peterzhu2118: I apologize if my comment sounds dismissive in any way, that is not my intention, I do think this is a valuable feature. I also think that it is my duty to ensure that the code base remains in a high quality, easy to understand, and well maintained condition, which is the reason for all of these comments. Your code looks fine, but tests are also equally as important since it prevents bugs or regressions from other developers. Therefore it's important that the tests are also of a high quality. A feature, no matter how small, needs to fit in the larger picture. What I'm asking here is to change your test to check that your code fits in the larger picture.

Again, it is not my intention to discourage you from working on this, so I apologize if it seems like such.

BurdetteLamar: I'm fine, Peter, with  your comments, impulses, and standards, so no problem there.  Nevertheless, I'm done with this.  There are many other things I can be working on that don't get me into such a tangle.

As always, thanks much for your help, which has been critical in keeping me away from many errors and blunders.

With your permission, I'll close this PR.

peterzhu2118: I don't want to make you feel like it's difficult to contribute code. I can understand that sometimes it feels like there's an endless stream of feedback and criticism, I've certainly felt that before. I want to be clear that I'm writing these comments with best intentions and I'm not trying to be difficult (and I realize I may seem as such).

I'll be sad if you threw out your hard work, especially since you're so close to merging it. I'm willing to provide any help or guidance you need to get this merged. However, if you feel like you certainly don't want to work on this any further, it's up to you to decide if you want to close this PR or not.

BurdetteLamar:
Poor ROI.  Closing, with malice toward none.

PR #899, opened 2022-07-25
Move doc/markup_reference.rb to doc/rdoc 

BurdetteLamar:


peterzhu2118 (commented):
I'm not sure what problem this is supposed to solve. It just includes the file `doc/markup_reference.rb` in the built `.gem` file that is built, but `markup_reference.rb` is just documentation, so I don't see why it's included in the built gem.

BurdetteLamar:
See https://ruby.slack.com/archives/C02A3SL0S/p1658264615750489.  I think this PR solves the problem.

peterzhu2118:
I don't see how this PR solves the problem. docs.ruby-lang.org is built from the ruby/ruby repository. I don't think this will sync the `doc/markup_reference.rb` file to the ruby/ruby repository.

BurdetteLamar:
Somehow, doc/ files from ruby/csv get into ruby/ruby, but doc/ files from ruby/rdoc don't.  If this is not the way, do you know how to make that happen?

peterzhu2118:
From what I can tell, it looks manually sync'd, not automatically. Looks like that tool to do the sync is in `tool/sync_default_gems.rb` and you can run it using `tool/sync_default_gems.rb -a rdoc`.

BurdetteLamar:
So I would add to the clause ` when "rdoc"` in that program?
`

peterzhu2118:
It would be a good idea to move all the files in the `doc` directory of RDoc to `doc/rdoc` to prevent name collisions. Then you can add something like:

```ruby
cp_r("#{upstream}/doc/rdoc", "doc")
```

BurdetteLamar:
> It would be a good idea to move all the files in the `doc` directory of RDoc to `doc/rdoc` to prevent name collisions. Then you can add something like:
> 
> ```ruby
> cp_r("#{upstream}/doc/rdoc", "doc")
> ```

Done.

peterzhu2118 (approved):
Can you change the PR title to match the changes? Otherwise it looks good.

PR #901, opened 2022-07-25
[DOC] Move section Directives into section Blocks

BurdetteLamar:
Section "Blocks" contains a subsection for each type of block, except for "Directives"; this moves section "Directives" into section "Blocks" as a subsection, just like the others.


peterzhu2118 (approved):
Looks good!

PR #903, opened 2022-07-26
[DOC] Clarifications for directives

BurdetteLamar:
- Former section "Directives in Trailing Comments" is reworked.  The important thing about a directive is what it does, not whether it's trailing or stand-alone.  Therefore I've worked the directives in the former section into the appropriate sections, based on function.
- Each directive is now explicitly marked as trailing or stand-alone.
- C-code directives are mentioned only for those directives that actually appear in our ruby/ruby C files, which are :startdoc:, :stopdoc:, :enddoc:, :include:, and :call-seq:.  What effect, if any, other directives have in C, I'm not sure about.


doc/rdoc/markup_reference.rb
```diff
@@ -911,4 +937,4 @@ def method(foo, bar)
     yield 'baz'
   end
 
-end
+end
```
peterzhu2118, line 940: Please keep the end of file newline

peterzhu2118 (approved):
Looks good!

PR #910, opened 2022-08-04
[DOC] Removes remaining old Markup Reference

BurdetteLamar:
Removes the remaining old Markup Reference in markup.rb.  (The previous removing PR did not get it all.)  The link to the new MarkupReference is already there.


lib/rdoc/markup.rb
```diff
@@ -99,355 +99,6 @@
 #
 # See RDoc::MarkupReference.
 #
-# === Escaping Text Markup
```
peterzhu2118, line 102: I don't see this information in `RDoc::MarkupReference`. This section should be incorporated in the `RDoc::MarkupReference` file so we don't lose information.

BurdetteLamar: Thanks, @peterzhu2118, for catching this important error.  In studying the omitted text, I've had some interesting insights about the rest of this page, which I'm still thinking about.  For the moment, I've just ported the omitted text without change.  More to come.

peterzhu2118 (approved):

PR #916, opened 2022-08-11
[DOC] Mods to section Text Markup

BurdetteLamar:
Takes note that text markup includes not only typeface, but also character conversions, links, and escapes.


peterzhu2118 (approved):
Looks good

PR #925, opened 2022-09-06
RBS: Add support for generics

raosush:
- Support generics for classes and modules
- Add a new syntax in RDoc comments for generating generics docs in Ruby


.ruby-version
```diff
@@ -0,0 +1 @@
+3.1.0
```
hsbt, line 1: Don't add your local configuration. Please remove this.

raosush: Updated the changes

PR #930, opened 2022-10-06
Remove unused abbrev require

composerinteralia:
This library originally used [`abbrev`](https://github.com/ruby/abbrev) to expand abbreviations into fully-qualified classes, but that was replaced in
https://github.com/ruby/rdoc/commit/f9ffe6684e2afeac65c62bf1a5a2fce729f21001

`abbrev` is no longer used anywhere, so this commit removes the require.


hsbt:
Nice! Thanks.

PR #935, opened 2022-10-07
Update rake ghpages task

nevans:
If the GitHub pages source is set to the "gh-pages" branch and the "/doc" folder, then this should get everything back up-to-date.

The older version should probably work, too.  But I never like running `rm -rf *` if I can avoid it!  I think the other changes are minor improvements, too.

It should be possible to run this rake task from a GitHub action.

-----

I ran the following:
```shell
bundle exec rake ghpages RDOC_DOCUMENTED_BRANCH=v6.3.2
git push
```

And the output can be seen here: https://nevans.github.io/rdoc

FYI: this is my GitHub Pages configuration:

Screenshot 2022-10-07 at 13-07-53 Build software better together

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -27,13 +27,43 @@ RDoc::Task.new do |doc|
   doc.rdoc_files = FileList.new %w[lib/**/*.rb *.rdoc doc/rdoc/markup_reference.rb] - PARSER_FILES
 end
 
-task ghpages: :rdoc do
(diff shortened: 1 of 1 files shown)


hsbt:
@nevans Thanks for your suggestion. I deployed gh-pages via https://github.com/ruby/rdoc/blob/master/.github/workflows/gh-pages.yml.

PR #941, opened 2022-11-22
Darkfish: Nest sidebar ToC as a tree of headings

nevans:
This uses `<details><summary>heading</summary><ul>nested</ul></detail>`, similar to how the classes and pages lists are now nested.

PR #942, opened 2022-11-22
Darkfish: group method call-seq in div.method-header

nevans:
This way, custom CSS styles can be easily applied to the entire method header at once.  Otherwise, it can be tricky to make a border that goes around the entire set of method call-seq, but not the method description.


nevans:
I've updating `net-imap` documentation.  To demonstrate this PR (as well as #941 and a few other monkeypatches that aren't ready or suitable for PRs), see the "Table of Contents" here: https://nevans.github.io/net-imap/Net/IMAP.html

PR #951, opened 2022-12-08
ci: ensure ubuntu image has libyaml-dev for psych 5

flavorjones:
The Github Actions default `ubuntu-latest` image does not have `libyaml-dev`, and `setup-ruby` does not install it. We need to explicitly install it until this issue is resolved somehow. See https://github.com/ruby/setup-ruby/issues/409 for discussion.


hsbt:
Thanks!

PR #962, opened 2023-01-05
Fix to use KeyboardEvent.key over keyCode

gemmaro:
`KeyboardEvent.keyCode` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode), so this replaces them with `KeyboardEvent.key`.


nobu:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#browser_compatibility

nobu (approved):

PR #965, opened 2023-01-15
Add Truffleruby to CI

gogainda:


hsbt:
Added truffleruby at https://github.com/ruby/rdoc/pull/977

PR #969, opened 2023-01-27
Fix reference to proper list type RDoc::MarkupReference

cjilbert504:
The "Lettered List" section referred to the previous "Numbered List" section, so this PR fixes that reference to point instead of lettered lists.


cjilbert504:
Thank you for merging, @nobu!

PR #971, opened 2023-02-01
Add rdoc:coverage default task

### LINKED ISSUE #268: [feature] RDoc::Task should include option to generate coverage report
Given a Rakefile with RDoc::Task configured for files, or using `.document`, it would be nice to generate a coverage report with the project file structure already known.

For example:

``` ruby
# Rakefile
require 'rdoc/task'

RDoc::Task.new do |rdoc|
  rdoc.rdoc_files.include("lib/foo.rb", "lib/foo/bar.rb")
end
```

We could then do something like:

``` bash
rake rdoc --coverage
```

To generate the coverage report, given the known files and other configuration from the task.

This way generating a report for a project is much simpler then adding all project files to the coverage report command (`rdoc -C ...`)

zzak:
Closes #268

### DIFF
--- a/lib/rdoc/task.rb
+++ b/lib/rdoc/task.rb
@@ -261,6 +261,7 @@ class RDoc::Task < Rake::TaskLib
         $stderr.puts "rdoc #{args.join ' '}" if Rake.application.options.trace
       end
     end
+
     self
   end
 
--- a/lib/rdoc/task.rb
+++ b/lib/rdoc/task.rb
@@ -50,6 +50,9 @@ require 'rake/tasklib'
 # [rerdoc]
 #   Rebuild the rdoc files from scratch, even if they are not out of date.
 #
+# [rdoc:coverage]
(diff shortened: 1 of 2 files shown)


lib/rdoc/task.rb
```diff
@@ -248,6 +251,16 @@ def define
...
+    namespace rdoc_task_name do
+      desc coverage_task_description
+      task coverage_task_name do
+        @before_running_rdoc.call if @before_running_rdoc
+        opts = option_list << "-C"
+        args = opts + @rdoc_files
+
+        $stderr.puts "rdoc #{args.join ' '}" if Rake.application.options.trace
+      end
+    end
```
olleolleolle, line 263:
```suggestion
    end

```

I think the earlier one had a newline before returning `self`.

olleolleolle (approved):
Looks good to me.

(I tried adding `-C -C` as options to rdoc, which didn't complain.)

PR #972, opened 2023-02-06
Actually execute RDoc document task for coverage

zzak:
There aren't currently any functional tests for the coverage task, I spent about 30 minutes but gave up. This is a small enough change, and I've tested it locally with Rails.

Was able to add some tiny coverage for `-C` flag however.


zzak:
Since this is broken, and #971 was merged, I'm going to push this. :bow:

PR #978, opened 2023-02-20
Omit descriptions and parameter lists for methods defined in C not mentioned in call-seq

jeremyevans:
This allows RDoc to better generate documentation for methods following the [Ruby core documentation guide](https://github.com/ruby/ruby/blob/master/doc/contributing/documentation_guide.md) (which omits aliases in call-seq in most cases).  This makes documentation for methods defined in C more similar to methods defined in Ruby.  For methods defined in Ruby, the method description of the aliased method is already not used (you have to explicitly document the alias to use it).

Internally, this adds AnyMethod#has_call_seq? and #skip_description?, and updates Darkfish to:

* only show the method name if there is a call-seq for the method, but the call-seq omits the method
* to omit the method description if the method is an alias or has aliases and has a call-seq that does not include the method

See discussion in https://github.com/ruby/ruby/pull/7316 for discussion leading to this pull request.

### DIFF
--- a/lib/rdoc/any_method.rb
+++ b/lib/rdoc/any_method.rb
@@ -116,6 +116,13 @@ class RDoc::AnyMethod < RDoc::MethodAttr
   end
 
   ##
+  # Whether the method has a call-seq.
+
+  def has_call_seq?
+    !!(@call_seq || is_alias_for&._call_seq)
+  end
+
+  ##
   # Loads is_alias_for from the internal name.  Returns nil if the alias
   # cannot be found.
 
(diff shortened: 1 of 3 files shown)


zzak:
@jeremyevans I appreciate the vote of confidence in requesting my review, but I'm not too familiar with this part of RDoc. You probably have more context then I do, we can ask in slack about getting you commit access if that helps.

Happy to pull this down and try it out too if that is helpful!

jeremyevans:
@zzak That's OK, I just requested review from you as I saw you had recent commits.  I already have commit access, so I could merge this myself, but I thought it best to have it reviewed by someone possibly more familiar with the internals. Taking another look at the commit log, maybe I can ask @nobu.

lib/rdoc/any_method.rb
```diff
@@ -296,6 +303,14 @@ def param_seq
     params
   end
 
+  ##
+  # Whether to skip the method description, true for methods that have
+  # aliases with a call-seq that doesn't include the method name.
+
+  def skip_description?
+    has_call_seq? && call_seq.nil? && !!(is_alias_for || !aliases.empty?)
```
nobu, line 311: The condition around aliases with `has_call_seq?` seems complicated a little.

nobu (approved):
Generally looks good.

PR #1007, opened 2023-05-09
Link checker for for RDoc

BurdetteLamar:
Checks links in locally-produced HTML pages a clone of ruby/rdoc.


BurdetteLamar:
Have moved link checker from lib/rdoc to test/links.

zzak:
I'm in the camp that this is fine to be in RDoc, and at least one other project (Rails) could benefit from having this type of check in place for the documentation. The problem with a gem is having to vet yet another dependency, and trust that it will be available in Rails build tools. What happens if you no longer are able to maintain it?

If this doesn't make sense for the rdoc gem, then at least I'd like to see it in the standard library where other committers have access to maintain and release it :pray:

BurdetteLamar:
Well, I am 80, so will not maintain forever.

    On Friday, May 19, 2023 at 06:38:58 PM CDT, zzak ***@***.***> wrote:  
 
 


I'm in the camp that this is fine to be in RDoc, and at least one other project (Rails) could benefit from having this type of check in place for the documentation. The problem with a gem is having to vet yet another dependency, and trust that it will be available in Rails build tools. What happens if you no longer are able to maintain it?

If this doesn't make sense for the rdoc gem, then at least I'd like to see it in the standard library where other committers have access to maintain and release it 🙏

BurdetteLamar:
Should this be brought up at a dev meeting?  Would be nice to get a definitive ruling.

    On Friday, May 19, 2023 at 06:38:58 PM CDT, zzak ***@***.***> wrote:  
 
 


I'm in the camp that this is fine to be in RDoc, and at least one other project (Rails) could benefit from having this type of check in place for the documentation. The problem with a gem is having to vet yet another dependency, and trust that it will be available in Rails build tools. What happens if you no longer are able to maintain it?

If this doesn't make sense for the rdoc gem, then at least I'd like to see it in the standard library where other committers have access to maintain and release it 🙏

PR #1016, opened 2023-06-02
Add HTML fragment references for constants

p8:
In large libraries with a lot of constants (Rails for example), it would be nice to deep link to these constants.

### DIFF
--- a/lib/rdoc/constant.rb
+++ b/lib/rdoc/constant.rb
@@ -60,6 +60,13 @@ class RDoc::Constant < RDoc::CodeObject
   end
 
   ##
+  # HTML fragment reference for this Constant
+
+  def aref
+    "#constant-#{full_name}"
+  end
+
+  ##
   # A constant is documented if it has a comment, or is an alias
   # for a documented class or module.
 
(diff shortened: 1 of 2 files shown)


p8:
It would be nice if we could deeplink to the following constants in the Rails API docs:
image
https://api.rubyonrails.org/classes/ActionDispatch/Http/Cache/Response.html

It helps when trying to communicate the values of these constants with others.
For example I could share the `DATE` constant here:
https://api.rubyonrails.org/classes/ActionDispatch/Http/Cache/Response.html#constant- ActionDispatch::Http::Cache::Response::DATE

nobu:
> https://api.rubyonrails.org/classes/ActionDispatch/Http/Cache/Response.html

Is that page generated with RDoc?

p8:
Yes. It uses https://github.com/rails/sdoc which is built on top of RDoc.

nobu:
It seems a role of views.
Shouldn't it be done in template files?

nobu:
SDoc doesn't seem adding `id` to constants. https://github.com/rails/sdoc/blob/4d1c71adff9c5f0d4fcf8c50eab29f4bcf301abf/lib/rdoc/generator/template/rails/_context.rhtml#L93C1
Shouldn't it be fixed first?

PR #1017, opened 2023-06-02
Add source_code, homepage and changelog uris to gemspec metadata

vinistock:
Add source_code, homepage and changelog uris to the gemspec metadata. These fields are used to populate links on `rubygems.org`. Also, dependabot uses `changelog_uri` on update PRs, which is a nice convenience.

### DIFF
--- a/rdoc.gemspec
+++ b/rdoc.gemspec
@@ -28,6 +28,10 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
   s.homepage = "https://ruby.github.io/rdoc"
   s.licenses = ["Ruby"]
 
+  s.metadata["homepage_uri"] = s.homepage
+  s.metadata["source_code_uri"] = "https://github.com/ruby/rdoc"
+  s.metadata["changelog_uri"] = "#{s.metadata["source_code_uri"]}/releases"
+
   s.bindir = "exe"
   s.executables = ["rdoc", "ri"]
   s.require_paths = ["lib"]


rdoc.gemspec
```diff
@@ -28,6 +28,10 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
   s.homepage = "https://ruby.github.io/rdoc"
   s.licenses = ["Ruby"]
 
+  s.metadata["homepage_uri"] = s.homepage
```
nobu, line 31: Doesn’t rubygems.org fallback to `#homepage`?

vinistock: Yeah, I think a lot of tools favour `homepage` over `metadata`, but I thought it wouldn't harm to add it. Do you prefer not adding `homepage_uri`?

nobu: Just I was curious a little.

nobu (approved):

rdoc.gemspec
amatsuda: `metadata.homepage_uri` value is going to be copied from `homepage` when `metadata.homepage_uri` was not given. So, it'd be better to keep `metadata.homepage_uri` empty if there's no special reason to specify both. We don't need to repeat what RubyGems internally does.

amatsuda: see: https://github.com/rubygems/rubygems/blob/57b6c04dc2aac0607d52468918b45babadd0e10d/lib/rubygems/specification_policy.rb#L411-L421

nobu: @amatsuda That method looks like just validating `Specification#homepage`, but nothing about `metadata["homepage_uri"]`, no?

amatsuda: Indeed! My bad, please forget about my comment above...!

PR #1019, opened 2023-06-06
Embedded racc parser for potability

hsbt:
The current `rdoc/rd/inline_parser.rb` and `rdoc/rd/block_parser.rb` have `require "racc/parser"` for external dependency. We should avoid to depend it for RD parser.


hsbt:
Note: `racc/parser` still have `racc/info`. We also embed it to `racc/parser`.

voxik:
For those who wonders how this impacts RDoc license, there should be no impact IMHO, because Racc explicitly allows to use the embedded parser with whatever license:

https://github.com/ruby/racc/blob/5eb07b28bfb3e193a1cac07798fe7be7e1e246c4/lib/racc/parser.rb#L8-L10

And FTR, maybe this could be mentioned somewhere in LEGAL.rdoc

PR #1025, opened 2023-07-05
Add support for mobile screen sizes on Darkfish

MatheusRich:
This adds support for reading the Darkfish-generated docs in mobile devices. I tried to keep the changes minimal, and the current layout was mostly preserved.

The most notable change is the navigation sidebar, which is now hidden by default on "small screens" (everything below 1024px). It can be toggled by the button on the top left corner. This button implements the ARIA pattern for a [disclosure widget]. The icon for the button was taken from [Iconoir], which is licensed under the MIT license.

The design and some of the implementation were loosely inspired by the [Elixir lang docs].

[disclosure widget]: https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/
[Iconoir]: https://iconoir.com/
[Elixir lang docs]: https://hexdocs.pm/elixir/1.15.2/Kernel.html

### General look (before/after)

<div style="display:flex;">
image
image
</div>

### Code blocks (before/after)

<div style="display:flex;">
image
image
</div>

### Desktop view (before/after)
The most notable change here is that I limited the content size to about 1000px. Long lines are hard to read, so I added this little quality-of-life improvement for desktop users.

image
image

The sidebar can also be closed on desktop:

image

## Video

https://github.com/ruby/rdoc/assets/20938712/0aabe07f-4c9e-409f-8e28-2222881320ea

## Testing

I wasn't sure if this change required tests. LMK if I need to add them.

### DIFF
--- a/lib/rdoc/generator/template/darkfish/_sidebar_toggle.rhtml
+++ b/lib/rdoc/generator/template/darkfish/_sidebar_toggle.rhtml
@@ -1,26 +1,3 @@
 <button id="navigation-toggle" class="navigation-toggle" aria-label="Toggle sidebar" aria-expanded="true" aria-controls="navigation">
-  <!--
-    Copyright (c) 2021 Luca Burgio
-
-    Permission is hereby granted, free of charge, to any person obtaining a copy
-    of this software and associated documentation files (the "Software"), to deal
-    in the Software without restriction, including without limitation the rights
-    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-    copies of the Software, and to permit persons to whom the Software is
-    furnished to do so, subject to the following conditions:
-
-    The above copyright notice and this permission notice shall be included in all
-    copies or substantial portions of the Software.
-
-    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-    SOFTWARE.
-  -->
-  <svg width="20px" height="24px" stroke-width="1.5" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg" color="#000000">
-    <path d="M3 5h18M3 12h18M3 19h18" stroke="#000000" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"></path>
-  </svg>
+  <span aria-hidden>☰</span>
 </button>
--- a/lib/rdoc/generator/template/darkfish/_sidebar_toggle.rhtml
+++ b/lib/rdoc/generator/template/darkfish/_sidebar_toggle.rhtml
@@ -1,0 +1,26 @@
+<button id="navigation-toggle" class="navigation-toggle" aria-label="Toggle sidebar" aria-expanded="true" aria-controls="navigation">
+  <!--
+    Copyright (c) 2021 Luca Burgio
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
+    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+    copies of the Software, and to permit persons to whom the Software is
+    furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included in all
+    copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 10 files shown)


lib/rdoc/generator/template/darkfish/_sidebar_toggle.rhtml
```diff
@@ -0,0 +1,26 @@
+<button id="navigation-toggle" class="navigation-toggle" aria-label="Toggle sidebar" aria-expanded="true" aria-controls="navigation">
+  <!--
+    Copyright (c) 2021 Luca Burgio
```
MatheusRich, line 3: I'm not sure this is the right way to attach the license.

lib/rdoc/generator/template/darkfish/servlet_not_found.rhtml
MatheusRich, line 1: I'm not sure what this file does, but I added it for completness.

lib/rdoc/generator/template/darkfish/servlet_root.rhtml
MatheusRich, line 1: I'm not sure what this file does, but I added it for completeness.

lib/rdoc/generator/template/darkfish/table_of_contents.rhtml
```diff
@@ -1,4 +1,13 @@
 <body id="top" class="table-of-contents">
+<%= render '_sidebar_toggle.rhtml' %>
+
+<nav id="navigation" role="navigation">
+  <div id="project-navigation">
+    <%= render '_sidebar_navigation.rhtml' %>
+
+    <%= render '_sidebar_search.rhtml' %>
+  </div>
+</nav>
```
MatheusRich, line 10: I'm not sure this needs more content, but I needed a sidebar on this page, otherwise, it's harder to return to the Home page.

lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -192,24 +192,40 @@ nav {
...
 }
 
 main {
   display: block;
-  margin: 1em;
+  margin: 3em 1em 1em;
   min-width: 340px;
   font-size: 16px;
+  width: 100%;
+  max-width: 64em;
```
MatheusRich, line 221: IMO, this could be even smaller.

lib/rdoc/generator/template/darkfish/_sidebar_toggle.rhtml
MatheusRich: Alternatively, maybe we could use U+2630 (☰) as an icon.

nobu: Or U+2261(≡)?

nobu:
Thank you, seems very nice.
Just one concern, when resizing the navigation pane, the menu icon doesn't move.

MatheusRich:
> when resizing the navigation pane, the menu icon doesn't move.

@nobu, I totally missed that it is resizable! Would it be a problem if we make it fixed-sized?

MatheusRich:
@nobu could you approve the workflow so we can see if this works on CI? Thank you!

MatheusRich:
It looks like the tests breaking here are the same as the ones in the main branch.

MatheusRich:
@nobu @hsbt is there something I can do to move this forward?

MatheusRich:
@nobu @hsbt I've rebased this PR. CI should pass now.

MatheusRich:
Sorry to ping you again, @nobu. Can we get a second look at this?

MatheusRich:
cc @eregon @hsbt

MatheusRich:
@peterzhu2118 can you take a look at this?

MatheusRich:
@hsbt any plans on merging this (or not)?

hsbt:
I have no opinion that.

@nobu How about your opinion?

MatheusRich:
@colby-swandale LMK if you have any thoughts on this

lib/rdoc/generator/template/darkfish/_sidebar_toggle.rhtml
st0012: Let's give `U+2630` a try? I'd prefer not importing additional asset for this.

lib/rdoc/generator/template/darkfish/css/rdoc.css
st0012: Can you explain why you think that's the case? (Asking as someone who has few css experience)

lib/rdoc/generator/template/darkfish/table_of_contents.rhtml
st0012: I think it makes sense. And normally this page would have rich space on both sides, so adding it should be ok.

st0012 (commented):
I tested the changes locally and I'd really love to see this merged 👍 But unfortunately I don't know enough css to give a confident review.

lib/rdoc/generator/template/darkfish/css/rdoc.css
MatheusRich: Very wide lines make text harder to read. I think what I have here is enough, but we could go narrower if people want to.

lib/rdoc/generator/template/darkfish/_sidebar_toggle.rhtml
MatheusRich: Updated [here](https://github.com/ruby/rdoc/pull/1025/commits/98f670e0fe55efb42554610cb195551cdd02cbec)

lib/rdoc/generator/template/darkfish/css/rdoc.css
MatheusRich: Here's a comparison on my monitor:

Max width of 1024px:
image


Max width of 896px:
image

st0012: Thanks for the explanation! Yeah I agree that the current full-width is not ideal. I don't have preference between 1024px and 896px tho, so let's keep it as is for now.

st0012 (approved):
Since RDoc's own documentation [ruby.github.io/rdoc/](https://ruby.github.io/rdoc/) is built with the latest RDoc, I think we can merge this PR and see how these changes play out.

jhawthorn:
Is this behaving as intended? On desktop (macbook air M2, seems to be the same in firefox, safari, and chrome) I get a sidebar open by default covering a large portion of the page (but also not the full page height)

Screenshot 2024-08-07 at 19-02-42 class RDoc Store - rdoc 6 7 0 Documentation

MatheusRich:
@jhawthorn there's an open PR to fix that already

PR #1031, opened 2023-07-25
handle symbols declared with %s

### LINKED ISSUE #1030: [Bug] Symbol is modified when declared with %s
when we declare a symbol like `%s[symbol]`, it is changed into `:symbol` in the documentation.

It confused me in this documentation: https://ruby-doc.org/3.2.2/syntax/literals_rdoc.html#label-25s-3A+Symbol+Literals

theo-squadracer:
Close #1030

Symbols declared with `%s[symbol]` the same way :'symbol', the entire string is used as a symbol token.


theo-squadracer:
Done, I don't know if I should do anything else

PR #1032, opened 2023-07-26
Use a more portable way to check if code is parseable

eregon:
* The same as used in irb: https://github.com/ruby/irb/pull/134/files
* And in `Forwardable._valid_method?`.
* This works on all Ruby implementations, unlike `return` in BEGIN which can be quite difficult to support.

### DIFF
--- a/lib/rdoc/markup/to_html.rb
+++ b/lib/rdoc/markup/to_html.rb
@@ -430,7 +430,9 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
 
   def parseable? text
     verbose, $VERBOSE = $VERBOSE, nil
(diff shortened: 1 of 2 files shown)


lib/rdoc/markup/to_html.rb
```diff
@@ -430,7 +430,10 @@ def list_end_for(list_type)
 
   def parseable? text
     verbose, $VERBOSE = $VERBOSE, nil
-    eval("BEGIN {return true}\n#{text}")
+    catch(:valid) do
+      eval("BEGIN { throw :valid, true }\n#{text}")
+      false
+    end
```
nobu, line 436: `false` line should never reach.

eregon: Right, removed.

PR #1038, opened 2023-09-05
Drop code for old versions

nobu:
We have already dropped the support for 2.5 or earlier because of CVE-2021-31799.

### DIFF
--- a/lib/rdoc/comment.rb
+++ b/lib/rdoc/comment.rb
@@ -133,12 +133,7 @@ class RDoc::Comment
   # HACK dubious
 
   def encode! encoding
-    # TODO: Remove this condition after Ruby 2.2 EOL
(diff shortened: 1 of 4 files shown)


hsbt:
Should we also bump up `required_ruby_version` to 2.7?

https://github.com/ruby/rdoc/blob/master/rdoc.gemspec#L233

nobu:
Dropped 2.5 and earlier because of the security issue, but I'm not sure about 2.6.
And why 2.7 not 3.0, is 2.7 needed to be supported still?

hsbt:
`rails/rails` still support Ruby 2.7 at `main` branch. If we drop Ruby 2.7 support from `rdoc`, it will break `rails` CI.

PR #1039, opened 2023-09-11
Delay DidYouMean until NotFoundError#message is called

tompng:
`expand_class` is slow for long class name that has no document.

```ruby
irb(main):001> driver = RDoc::RI::Driver.new
=> 
#<RDoc::RI::Driver:0x0000000105c03180
...
irb(main):002> measure
TIME is added.
=> nil
irb(main):003> driver.expand_class('MyFooBarApp::FooBarController::FooBar') rescue nil
processing time: 0.227079s
```

It's because did_you_mean is slow and always calculated before raising NotFoundError.

This pull request delays calculation of DidYouMean until `NotFoundError#message` is called.
Processing time reduces `0.227079s → 0.006436s`

## Background

This will fix https://github.com/ruby/irb/issues/709

### DIFF
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -34,9 +34,9 @@ class RDoc::RI::Driver
 
   class NotFoundError < Error
 
-    def initialize(klass, suggestions = nil) # :nodoc:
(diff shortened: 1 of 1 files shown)


MatheusRich:
Looks good to me. TY

st0012 (approved):

PR #1042, opened 2023-09-25
CI: use the v1 tag for ruby/setup-ruby

olleolleolle:
This PR changes the used versions for the GitHub Action setup-ruby from having used exact commits to using the "latest of the major v1 release".

Assumption: We trust the release process of that GH Action.


hsbt:
I know we can use `v1` version for updating automatically. But I want to track the result of `ruby/setup-ruby` each version.

Issue #1048, opened 2023-10-30
Unstable TestRDocGeneratorJsonIndex#test_generate test


junaruga:
I am seeing the following failure that happens randomly on Travis Ubuntu jammy ppc64le environment in ruby/ruby. The latest failure is at the relatively recent commit https://github.com/ruby/ruby/commit/1d51e4cadfbc1f445d00887189e06047759c9aec .

https://app.travis-ci.com/github/ruby/ruby/jobs/612435804#L2936

```
  1) Failure:
TestRDocGeneratorJsonIndex#test_generate [/home/travis/build/ruby/ruby/test/rdoc/test_rdoc_generator_json_index.rb:108]:
.js files should be the same timestamp of original
```

The test is built and executed with the following steps.

```
+ export NPROC="$(nproc)"
+ export JOBS=-j$((1+${NPROC}))
+ export RUBY_TESTOPTS="$JOBS -q --tty=no"
+ export optflags=-O1
+ export debugflags=-ggdb3
```

```
+ export CC=gcc-11
+ gcc-11 --version
gcc-11 (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
$ ../tool/travis_wait.sh $SETARCH make -s test-all RUBYOPT="-w"
...
Run options: 
  --seed=2200
  "--ruby=./miniruby -I../lib -I. -I.ext/common  ../tool/runruby.rb --extout=.ext  -- --disable-gems"
  --excludes-dir=../test/.excludes
  --name=!/memory_leak/
  -j3
  -q
  --tty=no
```

junaruga:
Another error log after removing the `optflags=-O1` in `.travis.yml`. So, the `optflags=-O1` is not criteria to make this issue happen.
https://app.travis-ci.com/github/ruby/ruby/jobs/612440144

junaruga:
I was able to capture the failing test, and compared values.

https://app.travis-ci.com/github/ruby/ruby/jobs/612446128#L2936

```
  1) Failure:
TestRDocGeneratorJsonIndex#test_generate [/home/travis/build/ruby/ruby/test/rdoc/test_rdoc_generator_json_index.rb:108]:
.js files should be the same timestamp of original.
[...]

PR #1049, opened 2023-10-30
test/rdoc/test_rdoc_generator_json_index.rb: Use assert_equal.

junaruga:
Related to https://github.com/ruby/rdoc/issues/1048.

Use `assert_equal` instead of of `assert`. It's better because assert_equal prints the values when it fails. I want to see the actual values in the failing case.

The `assert_equal` prints the values like this.

```
assert_equal 'a', 'b', 'dummy'
```

```
dummy
<"a"> expected but was
<"b">
```


junaruga:
Let me merge this PR. I think that no one is disturbed by this small change.

PR #1055, opened 2023-11-16
Fix `NoMethodError` for `tokens_to_s` method

toshimaru:
Calling `tokens_to_s` gets an error if `token_stream` is nil:

```
undefined method `compact' for nil:NilClass (NoMethodError)
```

So, fall back to an empty array if `@token_stream` is `nil`.

## The case `token_stream` is nil

```rb
> rdoc_method.inspect
=> "#<RDoc::AnyMethod:0x5b090 AbstractController::Translation#l (public) (alias for localize)>"
> rdoc_method.token_stream
=> nil
```

### DIFF
--- a/lib/rdoc/token_stream.rb
+++ b/lib/rdoc/token_stream.rb
@@ -105,7 +105,7 @@ module RDoc::TokenStream
   # Current token stream
 
   def token_stream
-    @token_stream
+    @token_stream || []
   end
 
   ##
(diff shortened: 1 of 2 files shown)


toshimaru:
@nobu thanks for fixing issue caused by this change. 🙏 
72897d3

jonathanhefner:
This seems like a breaking change, as demonstrated by https://github.com/rails/sdoc/pull/346.

`token_stream` is part of the public API, and now returns a different value when the method source is unavailable.

It is easy to handle this change in SDoc, but I wanted to point it out.

PR #1056, opened 2023-11-19
fix: Return empty string when `token_stream` is `nil`

toshimaru:
The change in #1055 could be a breaking change (see. https://github.com/ruby/rdoc/pull/1055#issuecomment-1816762305).

As a simple workaround, return empty string when `token_stream` is `nil`.

Additionally, a test for `test_token_stream` has been added.

### DIFF
--- a/lib/rdoc/token_stream.rb
+++ b/lib/rdoc/token_stream.rb
@@ -105,14 +105,14 @@ module RDoc::TokenStream
   # Current token stream
 
   def token_stream
-    @token_stream || []
+    @token_stream
   end
 
   ##
   # Returns a string representation of the token stream
 
   def tokens_to_s
-    token_stream.compact.map { |token| token[:text] }.join ''
+    Array(token_stream).compact.map { |token| token[:text] }.join ''
   end
 
 end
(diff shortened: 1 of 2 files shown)


lib/rdoc/token_stream.rb
```diff
@@ -105,13 +105,15 @@ def pop_token
...
+    @token_stream
   end
 
   ##
   # Returns a string representation of the token stream
 
   def tokens_to_s
+    return token_stream.to_s unless token_stream
+
     token_stream.compact.map { |token| token[:text] }.join ''
```
jonathanhefner, line 117: Using `Array()` could unify this code:

```suggestion
    Array(token_stream).compact.map { |token| token[:text] }.join ''
```

Alternatively, returning `+""` might be more explicit:

```suggestion
    return +"" if token_stream.nil?

    token_stream.compact.map { |token| token[:text] }.join ''
```

However, I am not an RDoc committer, and they may have a different opinion. :smile:

toshimaru: > Using `Array()` could unify this code:

Okay, I'll take it, thanks.

PR #1075, opened 2024-01-01
[DOC] MarkupReference

BurdetteLamar:
I want to establish that concepts ("Margins," "Blocks," etc.) are not just for RDoc comments, but also for .rdoc and .md text.  I chose not to call a chunk of RDoc text a _section_, b/c there's already a _section_ concept in RDoc;  so instead I called the chunk a _segment_.

### DIFF
--- a/doc/rdoc/markup_reference.rb
+++ b/doc/rdoc/markup_reference.rb
@@ -49,7 +49,7 @@ require 'rdoc'
 # - <tt>.rdoc</tt> (\RDoc markup text file) or <tt>.md</tt> (\RDoc markdown text file):
 #   markup is parsed from the entire file.
 #   The text is not associated with any code object,
-#   but may (depending on how the documentation is built),
+#   but may (depending on how the documentation is built)
 #   become a separate page.
 #
 # An <i>RDoc document</i>:
--- a/doc/rdoc/markup_reference.rb
+++ b/doc/rdoc/markup_reference.rb
@@ -63,15 +63,6 @@ require 'rdoc'
 # - The entire markup (<tt>.rdoc</tt>) file or markdown (<tt>.md</tt>) file,
 #   (which is usually multi-line).
 #
-# === Margins
-#
-# In a multi-line segment,
-# \RDoc looks for the segment's natural left margin,
-# which becomes the <em>base margin</em> for the segment
-# and is the initial <em>current margin</em> for the segment.
-#
-# The current margin can change, and does so, for example in a list.
-#
 # === Blocks
 #
 # It's convenient to think of a segment as a sequence of _blocks_
@@ -105,6 +96,10 @@ require 'rdoc'
 # - Any block may appear independently
 #   (that is, not nested in another block);
 #   some blocks may be nested, as detailed below.
+# - In a multi-line block,
+#   \RDoc looks for the block's natural left margin,
+#   which becomes the <em>base margin</em> for the block
+#   and is the initial <em>current margin</em> for the block.
 #
 # ==== Paragraphs
 #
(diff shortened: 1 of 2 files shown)


doc/rdoc/markup_reference.rb
```diff
@@ -7,72 +7,74 @@
...
+#   that is, they have no explicit comment markers like Ruby <tt>#</tt>
+#   or C <tt>/* ... */</tt>.
 # - An example that shows rendered HTML output
 #   displays that output in a blockquote:
 #
-#   Rendered HTML:
 #   >>>
 #     Some stuff
 #
 # \RDoc-generated documentation is derived from and controlled by:
```
peterzhu2118, line 24: Did you mean to remove this line too since the bullet points below are removed?

doc/rdoc/markup_reference.rb
```diff
@@ -7,72 +7,74 @@
...
+#   a function that implements a Ruby method,
+#   or otherwise immediately precedes the definition of a Ruby object,
+#   becomes the documentation for that object.
+# - <tt>.rdoc</tt> (\RDoc markup text file) or <tt>.md</tt> (\RDoc markdown text file):
+#   markup is parsed from the entire file.
+#   The text is not associated with any code object,
+#   but may (depending on how the documentation is built),
+#   become a separate page.
+#
+# An <i>RDoc segment</i> (hereafter simply called a _segment_) is:
```
peterzhu2118, line 59: I think RDoc internally calls this an "RDoc document", so maybe we can be consistent with that here?

doc/rdoc/markup_reference.rb
```diff
@@ -7,72 +7,74 @@
...
-# In a multi-line comment,
-# \RDoc looks for the comment's natural left margin,
-# which becomes the <em>base margin</em> for the comment
-# and is the initial <em>current margin</em> for the comment.
+# In a multi-line segment,
+# \RDoc looks for the segment's natural left margin,
+# which becomes the <em>base margin</em> for the segment
+# and is the initial <em>current margin</em> for the segment.
 #
 # The current margin can change, and does so, for example in a list.
```
peterzhu2118, line 73: Would it make sense to talk about margins in the context of blocks since each block has the same margin?

doc/rdoc/markup_reference.rb
BurdetteLamar: Removed.

doc/rdoc/markup_reference.rb
BurdetteLamar: Done.

doc/rdoc/markup_reference.rb
```diff
@@ -7,72 +7,61 @@
...
-# The current margin can change, and does so, for example in a list.
+# - <tt>.c</tt> (C code file): markup is parsed from C comments.
+#   A comment that immediately precedes
+#   a function that implements a Ruby method,
+#   or otherwise immediately precedes the definition of a Ruby object,
+#   becomes the documentation for that object.
+# - <tt>.rdoc</tt> (\RDoc markup text file) or <tt>.md</tt> (\RDoc markdown text file):
+#   markup is parsed from the entire file.
+#   The text is not associated with any code object,
+#   but may (depending on how the documentation is built),
```
peterzhu2118, line 52: Nit: I don't think a comment is needed here.

```suggestion
#   but may (depending on how the documentation is built)
```

doc/rdoc/markup_reference.rb
```diff
@@ -7,72 +7,61 @@
...
+#   become a separate page.
+#
+# An <i>RDoc document</i>:
+#
+# - A (possibly multi-line) comment in a Ruby or C file
+#   that generates \RDoc documentation (as above).
+# - The entire markup (<tt>.rdoc</tt>) file or markdown (<tt>.md</tt>) file,
```
peterzhu2118, line 59: Nit: I don't think a comment is needed here.

```suggestion
# - The entire markup (<tt>.rdoc</tt>) file or markdown (<tt>.md</tt>) file
```

peterzhu2118 (approved):

PR #1082, opened 2024-01-12
Fix ri completion to always return candidates start with a given name

tompng:
Fixes https://bugs.ruby-lang.org/issues/20171

RDoc has a test code like this
```ruby
assert_equal %w[Foo#Bar  Foo::bar], @driver.complete('Foo.')
```
Reline doesn't accept this completion list because it does not start from `'Foo.'`

To make it work with Reline, `completion_proc` should always return a list of string that starts with the given completion_target text.

This will also fix this weird behavior in Ruby 3.2 using readline
```
>> Array.t[TAB]
↓
>> Array
(tab complete deletes text)
```

### DIFF
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -744,7 +744,7 @@ or the PAGER environment variable.
     complete_klass  name, klass, selector, method, completions
     complete_method name, klass, selector,         completions
 
-    completions.sort.uniq
+    completions.uniq.select {|s| s.start_with? name }.sort
   end
 
   def complete_klass name, klass, selector, method, completions # :nodoc:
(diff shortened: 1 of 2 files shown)


lib/rdoc/ri/driver.rb
```diff
@@ -744,7 +744,7 @@ def complete name
     complete_klass  name, klass, selector, method, completions
     complete_method name, klass, selector,         completions
 
-    completions.sort.uniq
+    completions.sort.uniq.select {|s| s.start_with? name }
```
st0012, line 747: Just a suggestion: maybe call `.sort` last would be slightly more efficient given the list size being smaller?

st0012 (approved):
Thanks!

PR #1091, opened 2024-02-24
Add support for meta tags

vinistock:
## Motivation

Without being able to add SEO meta tags, such as `description` or `keywords`, documentation websites generated by RDoc often don't rank well in search engines. For example, Ruby's official documentation (generated by RDoc), is rarely the first result when searching for "Ruby documentation".

I think it would be really beneficial to allow developers to define meta tags for improving the SEO of their RDoc generated websites - and Ruby itself can benefit from it.

## Implementation

I split the implementation by commit:
1. Added `meta_tags` to options
2. Added `meta_tags` to the `RDoc::Task`
3. Used the `meta_tags` to create the entries in the HTML
4. Started using `meta_tags` for RDoc's own documentation

## Concerns

My main concern with this implementation is the format of the `meta_tags` option. I used a hash to provide flexibility in which meta tags the users can define. I believe the flexibility is important, because there are many meta tags and trying to account for all of them would be impractical (e.g.: `keywords`, `description`, `og:description`, `og:title` and so many others).

However, the current implementation of RDoc only allows for CLI-style options. Every option has to be passed as if it were coming from the command line (e.g.: `--something=otherthing`). This makes passing a hash to the `meta_tags` option quite weird. I'm currently converting it into a JSON, so that I can parse it back into a hash after the `OptionParse` extracts the options.

I don't really like the approach, but my two arguments for proposing this solution are:

1. If we're going to support meta tags, I believe flexibility is needed. There's no point in supporting only `keywords` and `description` as two hardcoded options
2. Refactoring RDoc to allow for options that don't conform to the CLI style is a bigger effort

Let me know what you think about this.

## Testing this change

1. Switch to this branch
2. Run `bundle exec rake rerdoc` to generate the website
4. Verify that the `_site/index.html` contains the following tags
```html
<meta name="keywords" content="...">
<meta name="description" content="...">
```

### DIFF
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -780,4 +780,19 @@ class RDoc::Generator::Darkfish
     template
   end
 
+  # Returns an excerpt of the content for usage in meta description tags
+  def excerpt(content)
+    text = content.is_a?(RDoc::Comment) ? content.text : content
+
+    # Match from a capital letter to the first period, discarding any links, so
+    # that we don't end up matching badges in the README
+    first_paragraph_match = text.match(/[A-Z][^\.:\/]+\./)
+    return text[0...150].gsub(/\n/, " ").squeeze(" ") unless first_paragraph_match
+
+    extracted_text = first_paragraph_match[0]
+    second_paragraph = first_paragraph_match.post_match.match(/[A-Z][^\.:\/]+\./)
+    extracted_text << " " << second_paragraph[0] if second_paragraph
+
+    extracted_text[0...150].gsub(/\n/, " ").squeeze(" ")
+  end
 end
--- a/test/rdoc/test_rdoc_generator_darkfish.rb
+++ b/test/rdoc/test_rdoc_generator_darkfish.rb
@@ -322,6 +322,51 @@ class TestRDocGeneratorDarkfish < RDoc::TestCase
     assert_main_title(File.binread('index.html'), title)
   end
 
+  def test_meta_tags_for_index
+    @options.title = "My awesome Ruby project"
+    @g.generate
+
+    content = File.binread("index.html")
+
+    assert_include(content, '<meta name="keywords" content="ruby,documentation,My awesome Ruby project">')
+    assert_include(content, '<meta name="description" content="Documentation for My awesome Ruby project">')
+  end
+
+  def test_meta_tags_for_classes
+    top_level = @store.add_file("file.rb")
+    top_level.add_class(@klass.class, @klass.name)
+    inner = @klass.add_class(RDoc::NormalClass, "Inner")
+    inner.add_comment("This is a normal class. It is fully documented.", top_level)
+
+    @g.generate
+
+    content = File.binread("Klass/Inner.html")
+    assert_include(content, '<meta name="keywords" content="ruby,class,Klass::Inner">')
+    assert_include(
+      content,
+      '<meta name="description" content="class Klass::Inner: This is a normal class. It is fully documented.">',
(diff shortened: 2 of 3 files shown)


vinistock:
I like the idea of using declaration information as part of the meta tags. How does Sdoc handle a scenario where there's no single declaration on the top level? For example this code:

```ruby
class Foo
end

class Bar
end
```
Does it consider the first declaration as the one to be used in meta tags? Or does it only support one top level namespace per file?

Also, how can we support meta tags for the main page (often the README)?

lib/rdoc/options.rb
```diff
@@ -790,6 +800,15 @@ def parse argv
         @skip_tests = false
       end
 
+      opt.separator nil
+
+       opt.on("-mTAGS", "--meta-tags=TAGS",
```
nobu, line 805:
```suggestion
      opt.on("-mTAGS", "--meta-tags=TAGS",
```

test/rdoc/test_rdoc_generator_darkfish.rb
```diff
@@ -322,6 +322,22 @@ def test_title_escape
...
+    }
+    @g.generate
+
+    content = File.binread("index.html")
+
+    assert_match("<meta name=\"keywords\" content=\"ruby,programming,software\">", content)
+    assert_match(
+      "<meta name=\"description\" content=\"RDoc is an awesome documentation generator for Ruby!\">",
+      content
+    )
```
nobu, line 338:
```suggestion
    assert_include(content, '<meta name="keywords" content="ruby,programming,software">')
    assert_include(
      content,
      '<meta name="description" content="RDoc is an awesome documentation generator for Ruby!">'
    )
```

vinistock:
In the latest commit, I changed the implementation to follow the approach suggested by @p8. I agree that page specific meta tags will definitely be better than a global definition. And it solves the issue of the weird JSON arguments.

I studied the approach SDoc is using a bit. It's using Nokogiri to find the first paragraph of text in any given page and then it uses that for the description.

I assumed that we cannot have a dependency on Nokogiri in RDoc, so I took a much simpler approach. The idea is to try to extract an excerpt of the class/module comment or the page's content up until the second period, the first period or at most 150 characters (which ever matches first).

It's definitely not perfect and it will be thrown off by other unrelated dot characters like the ones present in URLs or inside examples of method invocations in documented code. If you have any other ideas, please let me know.

@nobu I'd love to hear your thoughts on the approach.

nobu:
Sorry to be lazy.
LGTM

st0012 (approved):
Tested against `ruby/ruby` and it works great. Thank you!

lib/rdoc/generator/darkfish.rb
```diff
@@ -780,4 +780,19 @@ def template_for file, page = true, klass = ERB
     template
   end
 
+  # Returns an excerpt of the content for usage in meta description tags
+  def excerpt(content)
+    text = content.is_a?(RDoc::Comment) ? content.text : content
+
+    # Match from a capital letter to the first period, discarding any links, so
+    # that we don't end up matching badges in the README
+    first_paragraph_match = text.match(/[A-Z][^\.:\/]+\./)
```
nobu, line 789: Not All paragraphs start with a capital letter, as usual English text.
E.g., README.md in drb has the paragraph:

> dRuby is a distributed object system for Ruby.  It allows an object in one
> Ruby process to invoke methods on an object in another Ruby process on the
> same or a different machine.

I think it should be any text, from the alphabet to the first period, excluding some symbols.
However, this is still too strict for most natural languages.

PR #1095, opened 2024-03-09
Test only with truffleruby-head

nobu:
The old version is already known to fail.


nobu:
@eregon Hmmm, any plan to backport and release oracle/truffleruby#3378?

eregon:
24.0 is out, so I made https://github.com/ruby/rdoc/pull/1099

Issue #1107, opened 2024-04-29
RDoc doesn't have complete files when being installed from git/github source


st0012:
### Context

- RDoc uses parser generator `kpeg` for its markdown parser, which reads `lib/markdown.kpeg` to output `lib/markdown.rb`.
- RDoc doesn't commit the generated `.rb` files, but relies on its [`generate` rake task](https://github.com/ruby/rdoc/blob/master/Rakefile#L67-L96) to generate them.
- `generate` is a prerequisite of the `build` task, which is part of the `release` task. So files like `lib/markdown.rb` will be generated and packaged as part of the gem.

However, if users install RDoc from a git/github source, those generated files will be missing, and thus cause errors.

### Reproduction Steps

1. Run the command below

```
$ bundle exec ruby -e 'require "rdoc"; require "rdoc/markdown"; puts RDoc::Markdown'
RDoc::Markdown
```

2. Add `gem "rdoc", github: "ruby/rdoc"` to the Gemfile
3. Run `bundle install`
4. Run the same command again, which should now show an error like:

```
bundle exec ruby -e 'require "rdoc"; require "rdoc/markdown"; puts RDoc::Markdown'
/opt/rubies/3.3.0/lib/ruby/3.3.0/rdoc/markdown.rb:182:in `require': cannot load such file -- /Users/hung-wulo/.gem/ruby/3.3.0/bundler/gems/rdoc-8a68a016b25a/lib/rdoc/markdown (LoadError)
        from /opt/rubies/3.3.0/lib/ruby/3.3.0/rdoc/markdown.rb:182:in `<top (required)>'
        from -e:1:in `require'
        from -e:1:in `<main>'
```

### Possible solutions

- Commit the generated files into the git repository too, and have a way to automatically update them when their grammar files changes.
- Utilize `extconf.rb` to run `rake build`, which will be executed every time the gem is installed ([PoC](https://github.com/ruby/rdoc/commit/9d1df6b96143fb1edd06f3c6de8e318b966fe91c)).
- Rely on other markdown parsers that don't require parser generator?

colby-swandale:
Not a maintainer, but wanted to provide my 2c

> Commit the generated files into the git repository too, and have a way to automatically update them when their grammar files changes.

This is the easiest and fastest way to resolve this issue. I don't see any downside of doing this? Long term, the Markdown parser should be reviewed to see if the extra steps can be simplified/reduced into something not requiring compilation.

zenspider:
[...]

PR #1113, opened 2024-05-29
Allow installing RDoc from a git source

st0012:
This implements option 2 in #1107, which is to use `extconf.rb` to conditionally run `rake generate` when the gem is installed as a git source.

To make sure the target machine has the required dependencies for the generation, I modified `rake generate` to conditionally install the `kpeg` and `racc`.


Rakefile
```diff
@@ -92,8 +92,18 @@ parsed_files = PARSER_FILES.map do |parser_file|
...
 
 task "#{path}.gem" => package_parser_files
-desc "Generate all files used racc and kpeg"
-task :generate => parsed_files
+
+desc "Generate all files use racc and kpeg"
+task :generate do
+  unless ENV.key?('BUNDLE_GEMFILE')
+    Gem.install 'racc', '> 1.4.10'
+    Gem.install 'kpeg', '>= 1.3.3'
```
st0012, line 100: @colby-swandale do you know if there's any rubygems API to better handle these dependencies without bundler?

ext/rdoc/extconf.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Generate all parse files using racc and kpeg. This is not necessary for regular gem installation, but it is when
+# installing RDoc from the git source. Without this, the generated parse files would not exist and RDoc would not work
+system("rake generate") if Dir.exist?(File.join("..", "..", ".git"))
```
nobu, line 5: `.git` may be a regular file, when using `--git-dir` option.
```suggestion
system("rake generate") if File.exist?("../../.git")
```

Rakefile
```diff
@@ -92,8 +92,18 @@ parsed_files = PARSER_FILES.map do |parser_file|
...
+  end
+
+  parsed_files.each do |file_name|
+    Rake::Task[file_name].invoke
+  end
+end
```
nobu, line 106:
```suggestion
task :generate => [:gem_install, *parsed_files]

task :gem_install do
  unless ENV.key?('BUNDLE_GEMFILE')
    Gem.install 'racc', '> 1.4.10'
    Gem.install 'kpeg', '>= 1.3.3'
  end
end
```

nobu: This may be better.

```suggestion
task :generate => parsed_files

unless ENV.key?('BUNDLE_GEMFILE')
  task :gem_install do
    Gem.install 'racc', '> 1.4.10'
    Gem.install 'kpeg', '>= 1.3.3'
  end
  file parsed_files => :gem_install
end
```

Rakefile
colby-swandale: Which scenarios are you working with where Bundler would not be available?

st0012: Bundler will be available, but this will be executed in 2 scenarios:
- When a developer runs `rake generate` (or any tasks that depend on it) manually
- When a user installs RDoc through git source

In the later case, I want to avoid installing all of RDoc's development dependencies as most of them aren't needed for generating the missing files.

st0012:
@nobu Thanks for the suggestions 👍

st0012:
@nobu Would you mind giving this another look? If you're not against it, I'd like to give it a try.

ioquatix:
This does not work, at least when I tested it.

The problem is, `extconf.rb` is executed from the root directory, so the relative path (computed from `Dir.pwd` is wrong):

The fix is:

```ruby
git_path = File.expand_path("../../.git", __dir__)
if File.exist?(git_path)
	system("rake generate")
end
```

At least that gets us to the next issue:

```
> ruby ./ext/rdoc/extconf.rb
Generating lib/rdoc/markdown.rb...
rake aborted!
Gem::GemNotFoundException: can't find gem kpeg (>= 0) with executable kpeg (Gem::GemNotFoundException)
/home/samuel/.gem/ruby/3.4.1/bundler/gems/rdoc-c75f094920b3/Rakefile:80:in 'block (2 levels) in <top (required)>'
Tasks: TOP => generate => lib/rdoc/markdown.rb
(See full trace by running task with --trace)
creating Makefile
```

ioquatix:
Would it be crazy to do this:

```
gemspec

  if ENV["RDOC_FROM_SOURCE"] # or something better???
    gem "kpeg"
    ...
```

tompng:
#1107 is solved with #1410 (option 1 mentioned in the original issue)

PR #1114, opened 2024-06-01
Group code object files into the same directory

st0012:
It's hard to distinguish code object classes by their file names alone. And given that we have 18 such classes, it'd make the codebase a lot easier to understand if we grouped them into a single directory.

Given that these classes are all autoloaded in `lib/rdoc.rb` instead of required individually, this change should have minimum impact on projects using RDoc as they generally just require `rdoc`, not individual files. An example is Rails' `sdoc`:
https://github.com/rails/sdoc/blob/main/lib/sdoc/rdoc_monkey_patches.rb


hsbt:
I'm +1 for this. But I'm not sure how libraries refer these files directory like `require "rdoc/context"`.

Can you investigate that?

st0012:
- [Result](https://github.com/search?q=require+rdoc%2Ftop_level+path%3A*.rb&type=Code&ref=advsearch&l=&l=&p=1) for `rdoc/top_level`
- [Result](https://github.com/search?q=require+rdoc%2Fcontext+path%3A*.rb&type=code&ref=advsearch) for `rdoc/context`

Most of the results are from fork or vendored RDoc. A few gems that actually require RDoc explicitly all have been unmaintained for more than 10 years ([example1](https://github.com/chinasaur/rdoc_rails), [example2](https://github.com/8thlight/limelight_docs/blob/204cf2cb9d9751d2b3568b0b7fb565e4c4d98a81/production/__resources/gems/gems/limelight_rdoc-0.1.14/lib/limelight_rdoc/documentation.rb#L1)).

I think it's mostly because 13 years ago RDoc started [autoloading those classes](https://github.com/ruby/rdoc/commit/734550f3134612b06c63626835fccd9df3095e4a). And given the number of the classes needed, most libs probably just switched to `require "rdoc"` like Rails' `sdoc` currently does:

- [`github/markup`](https://github.com/github/markup/blob/a849a12ac5100d7d7acba30834eab73d0cf61983/lib/github/markup/rdoc.rb#L2)
- [`apipie-rails`](https://github.com/Apipie/apipie-rails/blob/eef47704a9d4961f4af031bf7ce8336f392b23fc/lib/apipie/markup.rb#L8)

simi (commented):
LGTM!

colby-swandale (approved):

PR #1115, opened 2024-06-01
Add ruby-core CI suite

st0012:
This is similar what [IRB](https://github.com/ruby/irb/blob/master/.github/workflows/ruby-core.yml) and [Reline](https://github.com/ruby/reline/blob/master/.github/workflows/ruby-core.yml) have: a CI suite that runs RDoc changes against Ruby's master branch. It's been very useful to catch issues that would otherwise break CRuby's CI suites.

### DIFF
--- a/.github/workflows/ruby-core.yml
+++ b/.github/workflows/ruby-core.yml
@@ -1,0 +1,59 @@
+name: ruby-core
+
+on:
+  pull_request:
+
+  push:
+    branches:
+      - master
+
+concurrency:
+  group: ci-${{ github.ref }}-${{ github.workflow }}
+
+permissions: # added using https://github.com/step-security/secure-workflows
+  contents: read
+
+jobs:
+  ruby_core:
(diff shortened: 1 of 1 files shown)


martinemde (commented):
Makes sense to me. I read the workflow but I haven't seen the others, however, if it runs and passes then I say merge it.

.github/workflows/ruby-core.yml
```diff
@@ -0,0 +1,59 @@
+name: ruby-core
+
+on:
+  pull_request:
+
+  push:
+    branches:
+      - master
+
+concurrency:
```
simi, line 10: Any particular reason for this?

hsbt (approved):
Looks good.

We should consider to move this job as re-using workflow on https://github.com/ruby/actions.

The repositories under the Ruby org only call it.

st0012:
@hsbt Thx, I added it to my todo list 👍

PR #1117, opened 2024-06-06
feat(darkfish): Add `Attributes` section to TOC and sidebar

### LINKED ISSUE #1053: Attributes not in methods TOC?
A class/module page documents its attributes in a section separate from that for its other instance methods.  The TOC at the left links to the non-attribute methods, but not to the attribute-generated methods.

Internally, there seems to be no distinction between non-attribute methods and attribute-generated methods.  Over at stackoverflow.com, queries about determining which way a method was defined go unanswered.

So my question is this:  If an attribute-generated method is a full-fledged method, indistinquishable from a non-attribute method, should it not be in the left TOC?

okuramasafumi:
This attempts to close #1053 

This is NOT meant to get merged soon, but the purpose is to confirm that this approach is correct.
After getting some positive comments (hopefully), I'll add tests for this addition and then it gets ready for merge.

### DIFF
--- a/lib/rdoc/generator/template/darkfish/_sidebar_attributes.rhtml
+++ b/lib/rdoc/generator/template/darkfish/_sidebar_attributes.rhtml
@@ -1,0 +1,12 @@
+<%- unless klass.attributes.empty? then %>
+<!-- Method Quickref -->
+<div id="attribute-list-section" class="nav-section">
+  <h3>Atributes</h3>
+
+  <ul class="link-list" role="directory">
+    <%- klass.attributes.each do |attribute| -%>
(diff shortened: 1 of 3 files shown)


BurdetteLamar:
For what it's worth, I still think we should treat an attribute the same as any other method (but I'm happy for this to be done either way).

st0012:
I'm open to give this a try, but IMO attributes should at least be splitted by newlines?

Screenshot 2024-07-12 at 21 38 43

BurdetteLamar:
> I'm open to give this a try, but IMO attributes should at least be splitted by newlines?

That's affirmative.

PR #1122, opened 2024-06-22
cleanup: Remove `webcvs` option and cvs-related code

okuramasafumi:
`webcvs` option is used for `cvs_url` method only, but `cvs_url` method seems NOT be used anywhere.

### DIFF
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -326,11 +326,6 @@ class RDoc::Options
   attr_accessor :verbosity
 
   ##
-  # URL of web cvs frontend
-
-  attr_accessor :webcvs
-
-  ##
   # Minimum visibility of a documented method. One of +:public+, +:protected+,
   # +:private+ or +:nodoc+.
   #
@@ -387,7 +382,6 @@ class RDoc::Options
     @update_output_dir = true
     @verbosity = 1
     @visibility = :protected
-    @webcvs = nil
     @write_options = false
     @encoding = Encoding::UTF_8
     @charset = @encoding.name
@@ -415,7 +409,6 @@ class RDoc::Options
     @template_dir   = map['template_dir']
     @title          = map['title']
     @visibility     = map['visibility']
-    @webcvs         = map['webcvs']
 
     @rdoc_include = sanitize_path map['rdoc_include']
     @static_path  = sanitize_path map['static_path']
@@ -447,7 +440,6 @@ class RDoc::Options
     @template_dir   = map['template_dir']   if map.has_key?('template_dir')
     @title          = map['title']          if map.has_key?('title')
     @visibility     = map['visibility']     if map.has_key?('visibility')
-    @webcvs         = map['webcvs']         if map.has_key?('webcvs')
 
     if map.has_key?('rdoc_include')
       @rdoc_include = sanitize_path map['rdoc_include']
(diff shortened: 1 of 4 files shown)


martinemde (commented):
It does look to me like this method is not used. I don't think anything external could call into it and CVS is quite old at this point (I assume most have moved to modern revision control systems like git).

Looks good to me.

lib/rdoc/options.rb
```diff
@@ -325,11 +325,6 @@ class RDoc::Options
 
   attr_accessor :verbosity
 
-  ##
-  # URL of web cvs frontend
-
-  attr_accessor :webcvs
```
simi, line 331: To be on super super safe side, would it make sense to make this no-op? Seems there are still some old gems using this like https://github.com/sunspot/sunspot/blob/c9c17d3d0e5869d157d682059b864d758ca12a8b/sunspot/sunspot.gemspec#L35.

related GH search https://github.com/search?q=rdoc_options+webcvs+language%3ARuby+&type=code

okuramasafumi: Or marking it as deprecated?

martinemde: The people who will see the deprecation probably aren't the author and can't fix it since it's in the gemspec.

lib/rdoc/options.rb
```diff
@@ -1017,17 +1008,6 @@ def parse argv
         @static_path << value
       end
 
-      opt.separator nil
-
-      opt.on("--webcvs=URL", "-W",
```
hsbt, line 1022: We should keep this for https://github.com/ruby/rdoc/pull/1122#discussion_r1649806944

okuramasafumi: @hsbt It has been restored. I believe `attr_accessor` part should remain removed since `RDoc::Option` class is not meant to be used directly by other gems.

st0012 (commented):
Thanks for the PR 👍  
In the cases of removing deadcode, if the the target has been exposed as public options or API, I'd keep it low priority unless it's blocking refactors or even features/bug fixes. And in this case, there doesn't seem to be a huge gain other than having slightly fewer code, especially when we still need to maintain some deprecation behaviours (e.g. we should not remove `cvs_url` right away but printing a warning instead).

But more importantly, I did find puppet [supporting](https://github.com/puppetlabs/puppet/blob/8047d1cdbf33affa85bc134cf4e6c264bffd7530/lib/puppet/util/rdoc/generators/puppet_generator.rb#L624) this option in their generator. So this brings a question: what'd be the reason we use to convince them to drop it? Similarly, when we accidentally break some users' libs/apps that actually still rely on it, how do we justify it?

lib/rdoc/options.rb
hsbt: Can you also update this description with deprecated warning?

And we carefully to remove these code with https://github.com/ruby/rdoc/pull/1122#pullrequestreview-2139737789

okuramasafumi:
@st0012 My original motivation is to remove the option so that we can clean up a usage documentation of RDoc a bit. Now that we know it's still used somewhat (thank you everyone!), we can keep this code and close this PR.

Btw, I don't understand how Puppet code you mentioned above works. `require 'rdoc/generators/html_generator'` doesn't work for the current RDoc, and there's no such file in the repository.

st0012:
> Btw, I don't understand how Puppet code you mentioned above works.

Sorry I won't have time to help you verify this. Please share with us if you find out more about it tho 🙂 

> we can keep this code and close this PR.

Cool I'll close it for now then. We can revisit this after we finished other less risky cleanups first.

PR #1125, opened 2024-06-26
Move `markup_reference.rb` into `misc` directory

### LINKED ISSUE #1123: [Proposal] `bundle exec rdoc` command should work for rdoc itself
## Problem

Currently, running `bundle exec rdoc` in the repository root causes the following error:

```
uh-oh! RDoc had a problem:

Directory doc already exists, but it looks like it isn't an RDoc directory.

Because RDoc doesn't want to risk destroying any of your existing files,
you'll need to specify a different output directory name (using the --op <dir>
option)


```

## Solution

That seems caused by `doc` directory. Inside it there's `rdoc/markup_reference.rb` file only.
While this file seems to be processed by `Rakefile`, it doesn't have to exist in `doc` directory since the file path is hardcoded.
So moving `rdoc/markup_reference.rb` file into some other place should be enough.

okuramasafumi:
Close #1123


okuramasafumi:
https://github.com/ruby/ruby/blob/master/tool/sync_default_gems.rb#L154
This line in ruby/ruby fails, and I don't know what to do... Could we simply delete that line?
I still believe this change worth, but if it's too difficult we need to abandon this PR.

st0012:
We need to update the line so it copies the file from the new destination instead.
I think we may limit disruption if we change things this way:
1. We merge this PR, which will NOT be synced to ruby/ruby because the script will fail. But it should not affect other projects' syncing nor fail core's build
    - @hsbt is my understanding correct?
2. Then we open a PR to ruby/ruby to update the script and manually sync the files

tompng:
`markup_reference.rb` moved to `doc/markup_reference/rdoc.rdoc` in #1542

PR #1129, opened 2024-06-28
Fix some typos

ydah:


lib/rdoc/rdoc.rb
```diff
@@ -422,10 +422,10 @@ def parse_files files
   end
 
   ##
-  # Removes file extensions known to be unparseable from +files+ and TAGS
+  # Removes file extensions known to be unparsable from +files+ and TAGS
   # files for emacs and vim.
 
-  def remove_unparseable files
```
st0012, line 428: Let's not change method names in typo fixes.

st0012 (changes requested):

lib/rdoc/rdoc.rb
ydah: Method name changes were excluded from this PR change.

st0012 (approved):
Thanks 👍

PR #1132, opened 2024-07-02
chore(bin): Add `bin/console` for better DX

okuramasafumi:
I wanted to try RDoc methods and noticed there's no `bin` dir. This `bin/console` loads `doc/rdoc/markup_reference` for easy testing and debugging.

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -1,0 +1,10 @@
+#!/usr/bin/env ruby
+
+require 'bundler/setup'
+require 'rdoc'
+
+# This is the easy way to prepare various kinds of RDoc objects.
+require_relative '../doc/rdoc/markup_reference'
+
+require 'irb'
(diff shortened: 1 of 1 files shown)


bin/console
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env ruby
+
+require 'bundler/setup'
+require 'rdoc'
+
+require_relative '../doc/rdoc/markup_reference'
```
colby-swandale, line 6: Would you be able to add a comment to explain why this is needed for future reference.

colby-swandale (approved):

bin/console
okuramasafumi: Done!

st0012 (approved):
Thanks! But do remember to include this file in #1125 after rebasing 🙂

PR #1139, opened 2024-07-12
Improve rubocop setup

st0012:
1. The previous `rubocop` rake task has been renamed to `format_generated_files` to express its intention clearer. The rubocop configs for this task now lives in `.generated_files_rubocop.yml` and is referenced explicitly in the new task's options.
2. A new `.rubocop.yml` is added with very simple rules to serve as a starting point. `bundle exec rubocop` should be used to format RDoc's source code.
3. A new `lint` job is created to enforce the rules in `.rubocop.uml`, and perform sanity check for the new `format_generated_files` task.


colby-swandale (approved):
:shipit:

PR #1140, opened 2024-07-13
Require space between hash/content in ATX heading

skipkayhil:
Ref rails/rails#50759
Ref rails/rails#50063

While writing some Markdown documentation for Rails, I came across an interesting case where trying to link to an instance method at the start of a line would instead parse as an H1 heading:

```markdown
#response_body=
```

Expected:

```html
<a href=""><code>#response_body=</code></a>
```

Actual:

```html
<h1>response_body=</h1>
```

According to the CommonMark spec:

> At least one space or tab is required between the # characters and the
> heading’s contents, unless the heading is empty. Note that many
> implementations currently do not require the space. However, the space
> was required by the original ATX implementation, and it helps prevent
> things like the following from being parsed as headings:
>
> Example 64

So while some implementations do not follow this requirement, I believe RDoc should because it makes it easy to write text similar to Example 64 (which was used in the new test) and it also enables automatically linking to instance methods at the start of a line.

### DIFF
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -530,7 +530,7 @@ AtxInline = !@Newline !(@Sp /#*/ @Sp @Newline) Inline
 AtxStart =  < /\#{1,6}/ >
             { text.length }
 
-AtxHeading = AtxStart:s @Sp AtxInline+:a (@Sp /#*/ @Sp)?  @Newline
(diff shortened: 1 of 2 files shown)


st0012 (approved):
Yeah I think the current behaviour is likely surprising and also confusing to most devs (myself included). Thanks for the change 👍

zzak:
Could you also use `rdoc-ref` in places where the autolinking is imperfect?
https://ruby.github.io/rdoc/RDoc/MarkupReference.html#class-RDoc::MarkupReference-label-Links

PR #1141, opened 2024-07-15
Expand rdoc-ref targets at the end of ri output

### LINKED ISSUE #1097: PLEASE make cross references actually work w/ ri
```
9982 $ ri File.new | tail -6
Optional keyword arguments opts specify:

* {Open Options}[rdoc-ref:IO@Open+Options].
* {Encoding options}[rdoc-ref:encodings.rdoc@Encoding+Options].


9983 $ ri rdoc-ref:IO@Open+Options
Nothing known about rdoc-ref
9984 $ ri rdoc-ref:encodings.rdoc@Encoding+Options
Nothing known about rdoc-ref
```

(Also, why are there 2 blank lines at the end of `File.new` output?)

or:

```
9989 $ ri Array#pack
= Array#pack

(from ruby core)
------------------------------------------------------------------------
  pack(template, buffer: nil) -> string

------------------------------------------------------------------------

Formats each element in self into a binary string; returns that string.
See {Packed Data}[rdoc-ref:packed_data.rdoc].


9990 $ ri rdoc-ref:packed_data.rdoc | head -1
Nothing known about rdoc-ref
9991 $ ri ruby:packed_data.rdoc | head -1
= Packed Data
9992 $ ri ruby:packed_data | head -1
= Packed Data
```

This just isn't right. `rdoc-ref` isn't a thing yet ri refers to it ALL OVER the place. There's not the faintest clue that `ruby:` is the correct prefix. Please make this tool usable again OR stop extracting the useful bits of the doco into cross referenced files that don't work.

st0012:
There have been several document refactors in ruby/ruby that extract individual methods/classes' documentation into separate files, like ruby/ruby#6567

Because RI is not capable of rendering those references, RI users are left with dramatically fewer documentation on those methods/classes.

This commit adds a new option `--expand-ref` (default: true) to expand all the rdoc-ref targets at the end of the output.

Closes #1097

### DIFF
--- a/lib/rdoc/ri/driver.rb
+++ b/lib/rdoc/ri/driver.rb
@@ -79,6 +79,7 @@ class RDoc::RI::Driver
     options[:interactive] = false
     options[:profile]     = false
     options[:show_all]    = false
+    options[:expand_refs] = true
     options[:use_stdout]  = !$stdout.tty?
     options[:width]       = 72
 
@@ -241,6 +242,12 @@ or the PAGER environment variable.
              "otherwise.  Valid formatters are:",
              "#{formatters.join(', ')}.", formatters) do |value|
         options[:formatter] = RDoc::Markup.const_get "To#{value.capitalize}"
+      end
+
+      opt.separator nil
+
+      opt.on("--[no-]expand-refs", "Expand rdoc-refs at the end of output") do |value|
+        options[:expand_refs] = value
       end
 
       opt.separator nil
@@ -425,6 +432,7 @@ or the PAGER environment variable.
     @use_stdout  = options[:use_stdout]
     @show_all    = options[:show_all]
     @width       = options[:width]
+    @expand_refs = options[:expand_refs]
   end
 
   ##
@@ -549,11 +557,8 @@ or the PAGER environment variable.
   # Looks up the method +name+ and adds it to +out+
 
   def add_method out, name
-    filtered   = lookup_method name
-
-    method_out = method_document name, filtered
-
-    out.concat method_out.parts
+    filtered = lookup_method name
+    method_document out, name, filtered
   end
 
   ##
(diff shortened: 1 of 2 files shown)


adam12:
Tested this on a few core/stdlib methods and it worked well! :+1:

adam12:
One issue that's not directly related but should be worth noting is it fails to work on `[#module-JSON-label-Parsing+Options]`.

I know it doesn't start with `rdoc-ref` but it appears to be a valid reference at least on the HTML generated site.

(the example is from `JSON.parse`, by the way).

zenspider:
Might be valid... but it might need to be changed just to fit this PR. Frankly I don't care as long as `ri Array#pack` works... But I do agree that xrefs for options NEED to be expanded sooner or later. Getting doco that says `**opts` and then has a broken xref you can't actually use isn't helping ANYONE.

eregon:
Would it be possible to expand the references inline like detailed in https://github.com/ruby/rdoc/issues/1097#issuecomment-2517618447 ? That would make ri usable again for such cases and be very similar to what used to be before these extractions

st0012:
@eregon My understanding is that many of the extracted pages are much longer than the original segments. For example, [`Array#pack`](https://docs.ruby-lang.org/en/3.1/Array.html#method-i-pack) before extraction is much shorter than the extracted [packed data page](https://docs.ruby-lang.org/en/master/packed_data_rdoc.html). So expanding links inline does not necessarily "restore" the previous experience. IMO, it will give a much worse experience if users suddenly encountered this expanded content when they scroll through the documentation of `Array`, for example.

st0012:
@adam12 @zenspider @eregon I've released this feature in RDoc 6.8.0. Can you give it a try to see if there's any issue?

eregon:
This seems to work nicely, thanks!

eregon:
Ref: PR to add back the concise table for pack/unpack docs: https://github.com/ruby/ruby/pull/12349

PR #1144, opened 2024-07-22
Add new ruby parser that uses Prism

tompng:
Adds a new ruby parser `rdoc/parser/prism_ruby.rb` that uses Prism.
The added new ruby parser is enabled by `ENV['RDOC_USE_PRISM_PARSER']`

# Background

RDoc should use Prism as a parser to fix parsing bugs someday. These are few examples.
```ruby
# endless method is not supported
def foo() = 1
# include ends with semicolon is ignored, making hard to write RDoc's test
class A; include M; end
```

Ruby parser should be refactored. Even small bugs are hard to fix because parser and comment handling is tightly coupled.
So I think `rdoc/parser/ruby.rb` needs to be completely rewritten.

This pull request completely rewrites `lib/rdoc/parser/ruby.rb` using Prism.
Note that this change will fix many bugs but might introduce unintentional changes or bugs.

# Implementation details

## Comments
There are two kind of comments. consecutive comments and modifier comments.

### Consecutive comment
Consecutive comment is a set of continual comments written in blank line. They are linked to the next non-blank line.
```ruby
01| # consecutive comment 1
02| # consecutive comment 1, linked to line 4
03|
04| # consecutive comment 2
05| # consecutive comment 2, linked to line 7
06|
07| puts(:hello)
08|
09| # consecutive comment 3, linked to line 11
10|
11| # consecutive comment 4
12| # consecutive comment 4, linked to line 14
13|
14| end
```

Consecutive comment that starts with `##\n` is a metaprogramming method comment, but this rule is not strictly followed. RDoc sourcecode itself does not follow this rule.

### Modifier comment

Modifier comment is a comment written in non-blank line.

```ruby
puts :hello # modifier comment 1
def foo # modifier comment 2
  42
end # modifier comment 3
```

## AST Traversing

Prism::Visitor is used to traverse AST node and to collect documentable comments and AST nodes.
When visitor visits a documentable node, we need to first process consecutive comments before start_line of the node which is not processed yet.
These unprocessed consecutive comment might be a metaprogramming comment that starts with `##\n`.
After that, we can document the node itself with consecutive comment linked to `node.start_line`.
In this phase, RDoc also looks modifier comment on `node.start_line` and on `node.end_line` for some node types.

```ruby
# standalone consecutive comment 2

# consecutive comment linked to undocumentable node
1 + 2

# consecutive comment to documentable node
def foo # modifier comment
  42
end # modifier comment
```

When visitor enters class/module node, we need to push that class/module to module_nesting stack.
Before leaving class/module node, we need to process unprocessed consecutive comments because these comments might be a metaprogramming comment that defines a method to the current class/module node. After that, we can pop class/module from module_nesting.

```ruby
# consecutive comment to module A
module A
  # consecutive comment to class B
  class B
    # consecutive comment to def f
    def f; end
    ##
    # standalone consecutive comment
    # :method: method_of_class_b
  end
  ##
  # standalone consecutive comment 2
  # :method: method_of_module_a
end
```

# Test

A new test `test_rdoc_parser_prism_ruby.rb` is added. Some ruby code passed to `util_parser` are extracted from `test_rdoc_parser_ruby.rb`.
To ensure there is no bad difference between the original RDoc::Parser::Ruby and the new RDoc::Parser::PrismRuby, test is run with both parsers.
Unfortunately, there are many difference that I think it's a bug. Assertion for these cases are skipped by `unless accept_legacy_bug?`

# Output HTML file differences

These are the difference of the generated html files between master branch and this pull request.
Most file are identical except for newlines between html tags.

## Removed files, these are not linked from any another html files.
```
RDoc/Markdown/Literals/MemoEntry.html
RDoc/Markdown/Literals/ParseError.html
RDoc/Markdown/Literals/RuleInfo.html
RDoc/Markdown/MemoEntry.html
RDoc/Markdown/ParseError.html
RDoc/Markdown/RuleInfo.html
RDoc/Parser/RipperStateLex/InnerStateLex.html
Rake.html
```

## Link in html file changed
```
RDoc/Encoding.html
  Link changed, improved
    - <a href="RDoc.html"><code>RDoc</code></a>
    + <a href="../RDoc.html"><code>RDoc</code></a>

RDoc/Generator.html
  Link changed, improved
    - <a href="RDoc.html"><code>RDoc</code></a>
    + <a href="../RDoc.html"><code>RDoc</code></a>
  Link is gone
    - <h2>Adding <a href="Options.html"><code>Options</code></a> to <code>rdoc</code>
    + <h2>Adding Options to <code>rdoc</code>

RDoc/Generator/Markup.html
  Link is gone
    - It allows RDoc’s <a href="../CodeObject.html"><code>CodeObject</code></a> tree to avoid
    + It allows RDoc’s CodeObject</a> tree to avoid
```

## Other changes
```
RDoc/Markup.html
  Constant `Heading` is added, bugfix

RDoc/Markup/Rule.html
  Superclass is changed from `Struct.new :weight` to `Object`
  RDoc does not support Struct as a superclass yet. It should be blank, unknown, or `Object` as a fallback.

RDoc/MarkupReference.html
  space is added after `<pre>:call-seq:`

RDoc/Parser/C.html
  `@enclosure_dependencies.extend TSort` is removed from extended module list. bugfix

RDoc/Parser/ChangeLog/Git.html
  Constant `LogEntry` is added, bugfix

RDoc/Parser/Ruby.html
  space is added after `:call-seq:`. Comment is bad because `:call-seq:` is written inside meta-comment style comment that start with `##\n`
  + <p>... either the :call-seq: , :arg: or :args: directives.</p>
  - <p>... either the :call-seq:, :arg: or :args: directives.</p>

RDoc/RI/ErRDoc/RI/Error.htmlror.html
  bugfix
  + <p class="link">RDoc::Error
  - <p class="link">RDoc::RDoc::Error

RDoc/Stats.html
  `def report; extend RDoc::Text; end` is excluded from extended module list. bugfix

RDoc/TopLevel.html
  wrong source of `def add_include` is fixed
  alias of name is added, (influence of the above source bug)
```


lib/rdoc/parser/ruby.rb
```diff
@@ -169,2204 +147,937 @@ def initialize(top_level, file_name, content, options, stats)
...
-
-    att
-  end
+  # Scans this Ruby file for Ruby constructs
 
-  ##
-  # Creates a module alias in +container+ at +rhs_name+ (or at the top-level
-  # for "::") with the name from +constant+.
+  def scan
+    @tokens = RDoc::Parser::RipperStateLex.parse(@content).sort_by { |t| [t.line_no, t.char_no] }
```
kddnewton, line 200: If you wanted, you could use `Prism.parse_lex`, which will give you a parse result with both the AST and the tokens. Alternatively, you could use `Prism.lex` or `Prism.lex_compat` to get back a list of tokens in a separate pass.

tompng: Thank you for letting me know.
Ideally, we should use it, but `RDoc::Parser::RipperStateLex` is doing some concatenation of tokens and RDoc's source code colorization seems to depends on it.

I want to make that part out of scope of this pull request.

kddnewton: That makes sense!

lib/rdoc/parser/ruby.rb
```diff
@@ -145,21 +136,8 @@ class RDoc::Parser::Ruby < RDoc::Parser
 
   parse_files_matching(/\.rbw?$/)
```
st0012, line 137: To make it easier to rollback or compare behaviour between the 2 parsers, can we create a new `parser/prism_ruby.rb` file to host the new parser, and do the same for the test file too?
And to switch we simply remove this line from the current parser and add it to the new one.

tompng: Updated :+1:

Added parser `Parser::PrismRuby` can be enabled by setting `ENV['RDOC_USE_PRISM_PARSER']`
Added test `test_rdoc_parser_prism_ruby.rb` tests both `Parser::PrismRuby` and `Parser::Ruby` (the above env is not needed for this test)

hsbt:
In my understanding, `Prism` only support Ruby 2.7+.

https://github.com/ruby/rdoc/blob/master/rdoc.gemspec#L233

This PR drop to support Ruby 2.6 from `RDoc`.

kddnewton:
2.6 has been EOL for 2 years now. That being said, if it's a blocker, I can add 2.6 support to Prism.

st0012:
Personally, I'm fine with dropping Ruby 2.6 support because `irb`, `reline`, and `debug` have all dropped it. Also, the community should be encouraged to upgrade their Ruby versions in pursuit of better tools that are built on better infrastructure. Expanding Prism support to EOL Ruby versions will be sending the opposite message IMO.

lib/rdoc/markup/pre_process.rb
```diff
@@ -220,11 +223,11 @@ def handle_directive prefix, directive, param, code_object = nil,
       # remove parameter &block
       code_object.params = code_object.params.sub(/,?\s*&\w+/, '') if code_object.params
 
-      code_object.block_params = param
+      code_object.block_params = param || ''
 
       blankline
     else
-      result = yield directive, param if block_given?
+      result = yield directive, param, line if block_given?
```
st0012, line 230: What are these changes for? Looks like they're fixing issues that'd also affect the default Ruby parser?

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,1081 @@
...
+    end
+  end
+
+  class MethodSignature < Prism::Visitor # :nodoc:
+    attr_reader :yields, :calls_super, :params
+    def initialize(def_node)
+      @yields = []
+      @calls_super = false
+      @params = "(#{def_node.parameters&.slice})"
+      def_node.body&.accept(self)
```
st0012, line 1061: I think for readability visitors should NOT invoke itself, so we should move this out of `initialize` and do this in `visit_def_node` instead

```rb
      if node.body
        signature = MethodSignature.new(node)
        node.body.accept(signature)
      end
```

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,1081 @@
...
+      when Prism::ConstantReadNode
+        node.name.to_s
+      when Prism::ConstantPathNode
+        parent_name = node.parent ? constant_path_string(node.parent) : ''
+        "#{parent_name}::#{node.name}" if parent_name
+      end
+    end
+  end
+
+  class MethodSignature < Prism::Visitor # :nodoc:
```
st0012, line 1055: Let's rename it to `MethodSignatureVisitor` and put it inside `RDocVisitor`?

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,1081 @@
...
+        singleton = true
+        visibility = :public
+      end
+    end
+
+    meth.singleton = singleton
+    receiver.add_method(meth) # should add after setting singleton and before setting visibility
+    meth.visibility = visibility
+    if signature
+      meth.params ||= signature.params
```
st0012, line 643: Let's avoid passing visitor objects around and simply pass `params`, `calls_super`, and `block_params` as separate arguments.

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,1081 @@
...
+    end
+
+    def visit_super_node(node)
+      @calls_super = true
+      super
+    end
+
+    def visit_forwarding_super_node(node)
+      @calls_super = true
+    end
```
st0012, line 1079: I noticed that these `@calls_super = true` aren't cover by tests?

rdoc.gemspec
```diff
@@ -233,5 +233,6 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
   s.required_ruby_version = Gem::Requirement.new(">= 2.6.0")
```
st0012, line 233: Let's bump it to `2.7` to match Prism's requirement.

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,1081 @@
...
+        node.location.end_line
+      )
+      @scanner.skip_comments_until(node.location.end_line)
+      # Do not traverse rhs not to document `A = Struct.new{def undocumentable_method; end}`
+    end
+
+    private
+
+    def constant_arguments_names(call_node)
+      return unless call_node.arguments
```
st0012, line 1021: Let's move these argument attribute's presence check to the caller so readers can see the code flow more easily.

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,1081 @@
...
+
+    def symbol_arguments(call_node)
+      arguments_node = call_node.arguments
+      return unless arguments_node && arguments_node.arguments.all? { |arg| arg.is_a?(Prism::SymbolNode)}
+      arguments_node.arguments.map { |arg| arg.value.to_sym }
+    end
+
+    def visibility_method_arguments(call_node, singleton:)
+      arguments_node = call_node.arguments
+      return unless arguments_node
```
st0012, line 1034: Same as the above.

lib/rdoc/markup/pre_process.rb
tompng: In the default ruby parser, the third parameter `line` is just ignored.

The line where `:method:` is written is the source line number of ghost method.
```ruby
1| class A
2|   ##
3|   # this is a ghost method `foo(x, y). Source location is line 4
4|   # :method: foo
5|   # :args: x, y
6| end
```

The original code is calculating line number by this code.
https://github.com/ruby/rdoc/blob/1bb3becd00704de879343eec31d202cf5e665862/lib/rdoc/parser/ruby.rb#L1094-L1095

`pre_process.rb` parses comment directives, `parse_comment` also parses some comment directives.
To avoid re-implementing these kind of thing, I changed `PreProcess#handle` to also provide line_number for each directive.

This is not the best way. pre_process has problem and I think it should be fixed in the future. pre_process needs code_object already created, but we need to parse the comment first to identify code_object type which can be either attribute or method.

lib/rdoc/parser/prism_ruby.rb
tompng: done!

lib/rdoc/parser/prism_ruby.rb
tompng: added in https://github.com/ruby/rdoc/pull/1144/commits/83b0fdefae3e292c76f26f0f7b434a2e8dee6bcc

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,1081 @@
...
+#   # :method: ghost_method
+#   # There is a method here, but you can't see it!
+#
+#   ##
+#   # this is a comment for a regular method
+#
+#   def regular_method() end
+#
+# Note that by default, the :method: directive will be ignored if there is a
+# standard rdocable item following it.
```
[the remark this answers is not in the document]
tompng: All of them, I think. And we should.
I removed these comments because it is a duplication of `lib/rdoc/parser/ruby.rb`. We can restore it to this file when removing `lib/rdoc/parser/ruby.rb`

lib/rdoc/parser/prism_ruby.rb
tompng: I think we should leave this.
`Prism::CallNode#arguments` returns nil when there is no argument. If it returns an empty argument node, it was simple but it's not. I think this is to make the footprint small. Maybe related to https://github.com/ruby/prism/issues/1915

In my opinion, caller side should not know about the detail of CallNode structure.
If it is a call_node, caller should just call `constant_arguments_names` to check if call_node's argument is a list of constants.

Added some test for this in https://github.com/ruby/rdoc/pull/1144/commits/65317242a1089171047c03d4b6ef4b2a30a1e432, Found documenting code that includes `alias_method 42 if false` fails with error in master

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,975 @@
...
+    mod.line = start_line
+    record_location(mod)
+    handle_modifier_directive(mod, start_line)
+    handle_modifier_directive(mod, end_line)
+    mod.add_comment(comment, @top_level) if comment
+    mod
+  end
+
+  class RDocVisitor < Prism::Visitor # :nodoc:
+    DSL = {
```
st0012, line 657: Can we make these `visit_call` methods more easily distinguishable from the visitor's external interface? I think adding `_` as prefix + moving them to the private section should be enough.

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,975 @@
...
+    end
+
+    # Example: line_no = 5, start_line = 2, comment_text = "# comment_start_line\n# comment\n"
+    # 1| class A
+    # 2|   # comment_start_line
+    # 3|   # comment
+    # 4|
+    # 5|   def f; end # comment linked to this line
+    # 6| end
+    @unprocessed_comments = consecutive_comments.reject(&:empty?).map do |comments|
```
st0012, line 155: We could avoid allocating multiple arrays by changing this to:

```suggestion
    consecutive_comments.reject!(&:empty?)
    @unprocessed_comments = consecutive_comments.map! do |comments|
```

I normally consider this micro-optimization. But given RDoc could have long consecutive comments, I feel this may help in some cases.

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,975 @@
...
+    def visit_def_node(node)
+      start_line = node.location.start_line
+      end_line = node.location.end_line
+      @scanner.process_comments_until(start_line - 1)
+
+      case node.receiver
+      when Prism::NilNode, Prism::TrueNode, Prism::FalseNode
+        visibility = :public
+        singleton = false
+        receiver_name = { Prism::NilNode => 'NilClass', Prism::TrueNode => 'TrueClass', Prism::FalseNode => 'FalseClass' }[node.receiver.class]
```
st0012, line 834: Could this trigger new hash allocation every time it's called?

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,975 @@
...
+        a = RDoc::Attr.new(@container, attr, rw, processed_comment)
+        a.store = @store
+        a.line = line_no
+        a.singleton = @singleton
+        record_location(a)
+        @container.add_attribute(a)
+        a.visibility = visibility
+      end
+    elsif line_no || node
+      method_name ||= call_node_name_arguments(node).first if is_call_node
```
st0012, line 283: Could we reuse `add_method` for this case?

lib/rdoc/parser/prism_ruby.rb
tompng: I added `_` 👍 
Visibility is still public because the code below is calling method through `v.` so we can't change the visibility to private.
```ruby
-> (v, call_node) { v._visit_call_attr_reader_writer_accessor(call_node, 'R') }
```
Of course we can use instance_eval or instance_exec to remove `v.` part, but I think not using `instance_eval/exec` is more straightforward.

lib/rdoc/parser/prism_ruby.rb
tompng: Changed to `case` `when` :+1:

lib/rdoc/parser/prism_ruby.rb
tompng: Reusing is a bit difficult because of the current comment handling. For now, I extracted common part to `def internal_add_method`.
I think we can make it clean after comment parsing logic is refactored.

st0012 (commented):
I think we're very close to merging this 👍

lib/rdoc/parser/prism_ruby.rb
tompng: Seems good and this way we can make it private!
done

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -0,0 +1,999 @@
...
+    def visibility_method_arguments(call_node, singleton:)
+      arguments_node = call_node.arguments
+      return unless arguments_node
+      symbols = symbol_arguments(call_node)
+      return symbols.map(&:to_s) if symbols # module_function :foo, :bar
+      return unless arguments_node.arguments.size == 1
+      arg = arguments_node.arguments.first
+      return unless arg.is_a?(Prism::DefNode)
+      # `module_function def foo; end` or `private_class_method def self.foo; end`
+      [arg.name.to_s] if singleton ? arg.receiver.is_a?(Prism::SelfNode) : arg.receiver.nil?
```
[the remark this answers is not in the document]
tompng: Updated and add these missing test case to test code.
```ruby
# should be ignored
private def self.foo; end
private def something.foo; end
private_class_method def foo; end
private def something.foo; end
``

File: test/rdoc/test_rdoc_parser_ruby.rb [hunk 17]
```diff
@@ -2,6 +2,8 @@
 
 require_relative 'helper'
 
+return if ENV['RDOC_USE_PRISM_PARSER']
```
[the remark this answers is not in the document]
st0012: I was planning to update CI setups myself after the PR is merged as it's easier to test those with write access.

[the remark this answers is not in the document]
tompng: The added test `test_rdoc_parser_prism_ruby.rb` already runs with both RDoc::Parser::Ruby and RDoc::Parser::PrismRuby on CI.
I don't have a good idea to run other test with `RDoc::Parser::Ruby` without using this environment variable.

st0012 (approved):
Thanks for the impressive work!
I'll merge this PR now and add a temporary CI step for it.
Let's start refactoring parsing process to make this new parser less coupled with legacy components before replacing the old one completely.

hsbt:
@st0012 @tompng https://github.com/ruby/rdoc/blob/master/rdoc.gemspec#L233 and https://github.com/ruby/rdoc/blob/master/rdoc.gemspec#L236 are bit of aggressive changes.

The release model of `rdoc` is branch cut from master HEAD. Now, If we have a plan to release 6.7.1, that teeny release contains `required_ruby_version` change and add C ext dependency. So, `gem i rdoc -v 6.7.1` needs C compiler. It's not useful.

We should keep `prism` is optional in 6.7.x at least.

st0012:
After discussing with other maintainers, we agreed to hold off making `prism` an official dependency of RDoc until 2025. We decided to be more conservative on this because no other Ruby default gems have used it as a dependency and the gem also hasn't reached 1.0 release yet.
Also, in 2025 with Ruby 3.5's development, RDoc will become a bundled gem, which should make such change easier too.

That being said, I think it's still great that we have this PR merged as we can use the Prism parser to evaluate all the upcoming refactorings 👍

PR #1145, opened 2024-07-31
Follow up changes for Prism Ruby parser

st0012:
1. When running RDoc with the Prism Ruby parser, there should be a clear message output
2. We should run tests with `RDOC_USE_PRISM_PARSER` on CI too
3. Since the Prism Ruby parser is optional and disabled by default, it doesn't make sense to require `prism` as dependency just yet


hsbt (approved):
Thanks for my consideration🙏

st0012:
@tompng FYI

PR #1148, opened 2024-08-04
Fix darkfish responsiveness issue on screens between 1024px and ~1650px 

st0012:
In that range, the sidebar would be displayed and cover part of the content. This PR fixes that by adding a left margin to the content when the screen size is greater than 1024px.

I also improved how the footer is displayed: instead of cutting the sidebar short to avoid covering it, we can increase the footer's z-index instead.

## Before

Screenshot 2024-08-04 at 14 46 41

## After

Screenshot 2024-08-04 at 14 46 53

Screenshot 2024-08-04 at 15 09 42

### DIFF
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -223,7 +223,8 @@ nav {
 
 main {
   display: block;
-  margin: 3em 1em 1em;
+  margin: 3em auto 1em;
+  padding: 0 1em; /* Add padding to keep space between main content and sidebar/right side of the screen */
   min-width: 340px;
   font-size: 16px;
   width: 100%;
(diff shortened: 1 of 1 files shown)


lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -232,8 +226,7 @@ main {
 
 @media (min-width: 1024px) {
   main {
-    margin-left: auto;
-    margin-right: auto;
+    margin-left: 300px;
```
MatheusRich, line 229: Is 300px a specific value that must be used here and somewhere else? Is it the sidebar width, for example? It might be an opportunity to use a CSS (or SASS) variable instead.

st0012: Yes it's `nav`'s width. I've added `--sidebar-width` variable for it.

colby-swandale (approved):

PR #1149, opened 2024-08-04
Change comment directive parsing

tompng:
Fix comment directive parsing problem

# Problem of comment parsing

The main problem is that `@preprocess.handle` parses comment, removes directive, and process code_object at the same time.
This pull request change RDoc to parse comment and extract directives first, and then apply directives to code object.

## Flow of legacy RDoc parsing method
For example parsing this code
```ruby
class A
  # :yields: x, y
  # :args:   a, b
  # :call-seq: 
  #--
  # :not-new:
  # :category: foobar
  #++
  #   initialize(x, y, z)
  def initialize(*args, &block); end
end
```

### Step 1
RDoc performs `@preprocess.hanlde` to RDoc::NormalClass.
- `:category:` is applied to klass and replaced with blank line
- `:not-new:` and `:yields:` are replaced with blank line. maybe bug.
- `:args:   a, b` is replaced with `:args: a, b`

### Step 2
RDoc performs `@preprocess.hanlde` to RDoc::AnyMethod.
`:args: a, b` is applied to `meth.params`.

### Step 3
RDoc removes private section that starts with `#--` and ends with `#++`.

### Step 4
RDoc normalizes comment by removing `#` and indentation.

### Step 5
RDoc extracts `":call-seq:\n  initialize(x, y, z)` from comment and apply to method object.

## Problems

RDoc removes directives and expand `:include:`  twice in some case, and once in other case.
To avoid all directives removed in the first `@preprocess.handle`, preprocess needs directive-replace mechanizm which is making things complex.

Private section and call-seq are processed later. This is making RDoc accept weird comment like directive inside private section and private section inside call-seq.

Handling meta programming method is also hard.
`@preprocess.handle(comment, code_object)` requires code object already created.
We need to parse the comment to know the code object type (method or attribute). After that, we can finally parse the comment with the code object.

C comments are also complicated. :include: can include text containing `*/`.
Removing directive line and private section from the comment might remove `/*` and `*/` which makes normalize_comment fail.
The original implementation was avoiding this by using different processing order than ruby parser. This is not consistent.

# Solution

We need to parse comment first and only once to extract directives.
Expand `:include:`, read directives (including `:call-seq:`), remove private section at the same time.
Comment parser should return normalized comment text and directives as an attribute hash. Directive should also contain line number.

# Changed things

## :call-seq:
New type of directive called "multiline directive" is introduced to make `:call-seq:` also a directive.
```
# :multiline-directive:
#   html
#     head
#       title
#
#     body
#       header
#       footer
```
Multiline directive ends with blank line. This restriction is for compatibility with old RDoc.
Some invalid multiline directive (unindented, ends with other directive) is also accepted with warning.

The resuld of parsing this call-seq is changed. I think it get better.
```
# :call-seq:
#   STDIN.getc()     -> string # Only this line was call-seq
#
#   STDIN.getc(a)    -> string
#
#   STDIN.getc(a, b) -> string
#   $stdin.getc(c)   -> string # It's now call-seq until this line
#
# :other:
```

## Private section

`#----foobar` was accepted as private section start.
`#++++foobar` was decomposed to `#++`(private end) and `++foobar`(normal comment).
Start is now `/^#-{2,}$/` (two or more -), end is now `/^#\+{2}$/` (exactly two +).

## Unhandled directives

In old RDoc, unhandled directive `# :unknown: foo` remain in normal comment.
Now it is removed just like other directives. Unhandled directive is appended to code object's metadata. It does not make sence to leave metadata in the comment. I think this was just a side effect of avoiding double parsing problem.

## Normalize and remove private section

Everything is done in parse phase

## C and Simple parser

C used to accept `/*\n# :directive:\n*/` but now only accepts `* :directive:`.
Changes for call-seq, private section and unhandled directive described above are also applied to C and Simple parser.

# Old comment parsing

`RDoc::Markup::PreProcess#handle` `RDoc::Comment#extract_call_seq` `RDoc::Comment#remove_private` is only used from `RDoc::Parser::Ruby`. We can remove them in the future.

# Diff

I compared generated html files of rdoc itself and c files in `ruby/ruby/*.c`.
Diff are in `File/Stat.html`, `Thread.html` and `RDoc/Parser/Ruby.html` which I think acceptable.

## File/Stat.html

Parsing this `call-seq:` is improved
```
ruby/ruby/file.c:5671

/*
 * call-seq:
 *
 *   File::Stat.new(file_name)  -> stat
 *
 * Create a File::Stat object for the given file name (raising an
 * exception if the file doesn't exist).
 */
```

## Thread.html

Parsing this `call-seq:` began to fail but I think indentation of this comment is wrong.
```
ruby/ruby/thread.c:875

/*
 * call-seq:
 *  Thread.new { ... }			-> thread
 *  Thread.new(*args, &proc)		-> thread
 *  Thread.new(*args) { |args| ... }	-> thread
 *
 *  Creates a new thread executing the given block.
 *  ...
 */
```

## RDoc/Parser/Ruby.html
```
lib/rdoc/parser/ruby.rb:122

##
# You can define arguments for metaprogrammed methods via either the
# :call-seq:, :arg: or :args: directives.
```

This is a metaprogramming comment that stars with `##` and contains `:call-seq:` directive. It should be escaped like
```
# \:call-seq:, ...
```

### DIFF
--- a/lib/rdoc/comment.rb
+++ b/lib/rdoc/comment.rb
@@ -165,6 +165,12 @@ class RDoc::Comment
     self
   end
 
+  # Change normalized, when creating already normalized comment.
+
+  def normalized=(value)
+    @normalized = value
+  end
+
   ##
   # Was this text normalized?
 
@@ -226,4 +232,164 @@ class RDoc::Comment
     @format == 'tomdoc'
   end
 
+  MULTILINE_DIRECTIVES = %w[call-seq].freeze # :nodoc:
+  private_constant :MULTILINE_DIRECTIVES
+
+  class << self
+
+    # Parse comment, collect directives as an attribute and return [normalized_comment_text, directives_hash]
+    # This method expands include and removes everything not needed in the document text, such as
+    # private section, directive line, comment characters `# /* * */` and indent spaces.
+    #
+    # RDoc comment consists of include, directive, multiline directive, private section and comment text.
+    #
+    # Include
+    #   # :include: filename
+    #
+    # Directive
+    #   # :directive-without-value:
+    #   # :directive-with-value: value
+    #
+    # Multiline directive (only :call-seq:)
+    #   # :multiline-directive:
+    #   #   value1
+    #   #   value2
+    #
+    # Private section
+    #   #--
+    #   # private comment
+    #   #++
+
+    def parse(text, filename, line_no, type)
+      case type
+      when :ruby
+        private_start_regexp = /^#?-{2,}$/
+        private_end_regexp = /^#?\+{2}$/
+        indent_regexp = /^#?\s*/
+        text = text.sub(/\A#\#$/, '')
+      when :c
+        private_start_regexp = /^(\s*\*)?-{2,}$/
+        private_end_regexp = /^(\s*\*)?\+{2}$/
+        indent_regexp = /^\s*(\/\*+|\*)?\s*/
+        text = text.sub(/\s*\*+\/\s*\z/, '')
+        # TODO: should not be here. Looks like another type of directive
+        text = text.gsub %r%Document-method:\s+[\w:.#=!?|^&<>~+\-/*\%@`\[\]]+%, ''
+      when :simple
+        # Unlike other types, this implementation only looks for two
+        # dashes at the beginning of the line.  Three or more dashes are considered
+        # to be a rule and ignored.
+        private_start_regexp = /^-{2}$/
+        private_end_regexp = /^\+{2}$/
+        indent_regexp = /^\s*/
+      end
+
+      directives = {}
+      lines = text.split("\n")
+      in_private = false
+      comment_lines = []
+      until lines.empty?
+        line = lines.shift
+        read_lines = 1
+        if in_private
+          in_private = false if line.match?(private_end_regexp)
+          line_no += read_lines
+          next
+        elsif line.match?(private_start_regexp)
+          in_private = true
+          line_no += read_lines
+          next
+        end
+
+        prefix = line[indent_regexp]
+        prefix_indent = ' ' * prefix.size
+        line = line.byteslice(prefix.bytesize..)
+        /\A(?<colon>\\?:|:?)(?<directive>[\w-]+):(?<param>.*)/ =~ line
+        directive = directive&.downcase
+
+        if colon == '\\:'
+          # unescape if escaped
+          comment_lines << prefix_indent + line.sub('\\:', ':')
+        elsif !directive || param.start_with?(':') || (colon.empty? && directive != 'call-seq')
+          # Something like `:toto::` is not a directive
+          # directive without colon prefix is only allowed for call-seq
+          comment_lines << prefix_indent + line
+        elsif directive == 'include'
+          filename_to_include = param.strip
+          yield(filename_to_include, prefix_indent).lines.each { |l| comment_lines << l.chomp }
+        elsif MULTILINE_DIRECTIVES.include?(directive)
+          param = param.strip
+          value_lines = take_multiline_directive_value_lines(directive, filename, line_no, lines, prefix_indent, indent_regexp, private_start_regexp)
+          read_lines += value_lines.size
+          lines.shift(value_lines.size)
+          unless param.empty?
+            # Accept `:call-seq: first-line\n  second-line` for now
+            value_lines.unshift(param)
+          end
+          value = value_lines.join("\n")
+          directives[directive] = [value.empty? ? nil : value, line_no]
+        else
+          value = param.strip
+          directives[directive] = [value.empty? ? nil : value, line_no]
+        end
+        line_no += read_lines
+      end
+      # normalize comment
+      min_spaces = nil
+      comment_lines.each do |l|
+        next if l.match?(/\A\s*\z/)
+        n = l[/\A */].size
+        min_spaces = n if !min_spaces || n < min_spaces
+      end
+      comment_lines.map! { |l| l[min_spaces..] || '' } if min_spaces
+      comment_lines.shift while comment_lines.first&.empty?
+
+      [String.new(encoding: text.encoding) << comment_lines.join("\n"), directives]
+    end
+
+    # Take value lines of multiline directive
+
+    private def take_multiline_directive_value_lines(directive, filename, line_no, lines, base_indent, indent_regexp, private_start_regexp)
+      return [] if lines.empty?
+
+      base_indent_size = base_indent.size
+
+      if lines.first[indent_regexp].size <= base_indent_size && !lines.first.sub(indent_regexp, '').chomp.empty?
+        # Unindented invalid block directive
+        warn "#{filename}:#{line_no} Multiline directive :#{directive}: should be indented."
+        # Take until next blank line, directive line or private section start to accept this invalid format
+        #   # :multiline-directive:
+        #   # line1
+        #   # line2
+        #   # :other-directive:
+        value_lines = lines.take_while do |l|
+          next false if l.match?(private_start_regexp)
+          l = l.sub(indent_regexp, '').chomp
+          !l.empty? && !l.start_with?(/:[\w-]+:/)
+        end.map do |l|
+          l.sub(indent_regexp, '').chomp
+        end
+      else
+        # Take indented lines accepting blank lines between them
+        first_indent = nil
+        value_lines = lines.take_while do |l|
+          l = l.rstrip
+          indent = l[indent_regexp]
+          if indent == l || (first_indent && indent.size >= first_indent)
+            true
+          elsif first_indent.nil? && indent.size > base_indent_size
+            first_indent = indent.size
+            true
+          end
+        end
+        first_indent ||= base_indent_size
+        value_lines.map! { |l| (l[first_indent..] || '').chomp }
+
+        if value_lines.size != lines.size && !value_lines.last.empty?
+          warn "#{filename}:#{line_no} Multiline directive :#{directive}: should end with a blank line."
+        end
+        value_lines.pop while value_lines.last&.empty?
+      end
+      value_lines
+    end
+  end
 end
--- a/lib/rdoc/markup/pre_process.rb
+++ b/lib/rdoc/markup/pre_process.rb
@@ -97,18 +97,15 @@ class RDoc::Markup::PreProcess
   # RDoc::CodeObject#metadata for details.
 
   def handle text, code_object = nil, &block
-    first_line = 1
     if RDoc::Comment === text then
       comment = text
       text = text.text
-      first_line = comment.line || 1
     end
 
     # regexp helper (square brackets for optional)
     # $1      $2  $3        $4      $5
     # [prefix][\]:directive:[spaces][param]newline
-    text = text.lines.map.with_index(first_line) do |line, num|
-      next line unless line =~ /\A([ \t]*(?:#|\/?\*)?[ \t]*)(\\?):([\w-]+):([ \t]*)(.+)?(\r?\n|$)/
+    text = text.gsub(/^([ \t]*(?:#|\/?\*)?[ \t]*)(\\?):([\w-]+):([ \t]*)(.+)?(\r?\n|$)/) do
       # skip something like ':toto::'
       next $& if $4.empty? and $5 and $5[0, 1] == ':'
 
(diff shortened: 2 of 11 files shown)


st0012:
Thanks for the PR! But because 1) the new maintainers are still learning the codebase, and 2) this PR touches a core part of RDoc, I'll hold off merging it until version 6.8.0 series is released 🙏

lib/rdoc/comment.rb
```diff
@@ -223,14 +229,169 @@ def tomdoc?
...
+
+      if has_param
+        # :multiline-directive: line1
+        #   line2
+        #   line3
+        #
+        value_lines = lines.take_while do |l|
+          l.rstrip[indent_regexp].size > base_indent_size
+        end
+        min_indent = value_lines.map { |l| l[indent_regexp].size }.min
```
st0012, line 376: Can this be:


```suggestion
        min_indent = value_lines.min_by { |l| l[indent_regexp].size }.size
```

lib/rdoc/comment.rb
```diff
@@ -223,14 +229,169 @@ def tomdoc?
...
+      case type
+      when :ruby
+        text = text.gsub(/^#+/, '') if text.start_with?('#')
+        private_start_regexp = /^-{2,}$/
+        private_end_regexp = /^\+{2}$/
+        indent_regexp = /^\s*/
+      when :c
+        private_start_regexp = /^(\s*\*)?-{2,}$/
+        private_end_regexp = /^(\s*\*)?\+{2}$/
+        indent_regexp = /^\s*(\/\*+|\*)?\s*/
```
st0012, line 284: Can we store these static regexp objects in constants instead?

lib/rdoc/comment.rb
```diff
@@ -223,14 +229,169 @@ def tomdoc?
...
+
+      directives = {}
+      lines = text.split("\n")
+      in_private = false
+      comment_lines = []
+      until lines.empty?
+        line = lines.shift
+        read_lines = 1
+        if in_private
+          in_private = false if line.match?(private_end_regexp)
```
st0012, line 305: Why do we need to flip `in_private` here? Can we add a comment explaining it?

lib/rdoc/comment.rb
```diff
@@ -223,14 +229,169 @@ def tomdoc?
...
+        elsif line.match?(private_start_regexp)
+          in_private = true
+          line_no += read_lines
+          next
+        end
+
+        prefix = line[indent_regexp]
+        prefix_indent = ' ' * prefix.size
+        line = line.byteslice(prefix.bytesize..)
+        /\A(?<colon>\\?:|:?)(?<directive>[\w-]+):(?<param>.*)/ =~ line
```
st0012, line 317: Let's move this regexp into a constant with an expressive name too.

lib/rdoc/comment.rb
```diff
@@ -223,14 +229,169 @@ def tomdoc?
...
+          end
+          value = value_lines.join("\n")
+          directives[directive] = [value.empty? ? nil : value, line_no]
+        else
+          value = param.strip
+          directives[directive] = [value.empty? ? nil : value, line_no]
+        end
+        line_no += read_lines
+      end
+      # normalize comment
```
st0012, line 346: Can this part be extracted into a `normalize_comment` method?

lib/rdoc/comment.rb
```diff
@@ -223,14 +229,169 @@ def tomdoc?
...
+          in_private = true
+          line_no += read_lines
+          next
+        end
+
+        prefix = line[indent_regexp]
+        prefix_indent = ' ' * prefix.size
+        line = line.byteslice(prefix.bytesize..)
+        /\A(?<colon>\\?:|:?)(?<directive>[\w-]+):(?<param>.*)/ =~ line
+
```
st0012, line 318: Let's add `param.strip!` here so we don't need to call `param.strip` in multiple branches?

lib/rdoc/comment.rb
```diff
@@ -223,14 +229,169 @@ def tomdoc?
...
+    #   # :multiline-directive:
+    #   #   value1
+    #   #   value2
+    #
+    # Private section
+    #   #--
+    #   # private comment
+    #   #++
+
+    def parse(text, filename, line_no, type)
```
st0012, line 274: Let's declare the block for `include` directive in parameters too, like:

```suggestion
    def parse(text, filename, line_no, type, &include_callback)
```

lib/rdoc/comment.rb
tompng: It can be like this
```ruby
value_lines = ['    arg2,', '    arg3,', '  )']
indent_regexp = /^\s*/

min_indent = value_lines.min_by { |l| l[indent_regexp].size }[indent_regexp].size #=> 2
```
but I think `map{}.min` is better

lib/rdoc/comment.rb
tompng: private_start_regexp (matches to `--`) begins private section (`in_private = true`)
private_end_regexp (matchs to `++`) ends private section (`in_private = false`)

comment added

lib/rdoc/comment.rb
tompng: Good catch :+1:
changed to:
```ruby
raw_param = directive_match[:param]
param = raw_param.strip
```
Unstripped raw_param is also used to reject `:toto::` like in the old implementation in pre_process.rb
https://github.com/ruby/rdoc/blob/a253a8d163c36ede4064083d5d43f69f789d0c51/lib/rdoc/markup/pre_process.rb#L112

lib/rdoc/comment.rb
tompng: Because these locals are a layer to absorbs differences between types
```ruby
when :ruby
  private_start_regexp = /^-{2,}$/
when :c
  private_start_regexp = /^(\s*\*)?-{2,}$/
when :simple
  private_start_regexp = /^-{2}$/
end
...
line.match?(private_start_regexp)
...
value_lines = take_multiline_directive_value_lines(..., indent_regexp, ...)
```

I think it could be simplified to:
1. Remove # or * depend on type first
2. Use the same regexp for all types

But it will introduce incompatibility.
```
horizontal divider between two paragraphs in :simple
---

--
only two dashes starts private section in :simple
++

#-----
# Two or more dashes after # are private section start in :ruby
# Tested in RDocCommentTest#test_remove_private_long
#++
```

Removing spaces before `*` instead of replacing `*` to a space (needed to distinguish `*---` from `* ---`) also brings changes in a c comment where * are not aligned. (I think this is a small problem though)
```
  /*
*    :call-seq:
   *   foo(bar)
    */
```

lib/rdoc/comment.rb
```diff
@@ -223,14 +229,189 @@ def tomdoc?
...
+      [normalized_comment, directives]
+    end
+
+    # Remove preceding indent spaces and blank lines from the comment lines
+
+    private def normalize_comment_lines(lines)
+      blank_line_regexp = /\A\s*\z/
+      min_spaces = lines.map do |l|
+        l[/\A */].size unless l.match?(blank_line_regexp)
+      end.compact.min
```
nobu, line 367:
```suggestion
      min_spaces = lines.map do |l|
        l.match(/\A *(?=\S)/)&.begin(0)
      end.compact.min
```

lib/rdoc/comment.rb
```diff
@@ -223,14 +229,189 @@ def tomdoc?
...
+      if min_spaces && min_spaces > 0
+        lines = lines.map { |l| l[min_spaces..] || '' }
+      else
+        lines = lines.dup
+      end
+      lines.shift while lines.first&.match?(blank_line_regexp)
+      lines.pop while lines.last&.match?(blank_line_regexp)
```
nobu, line 374: Stripping blank lines can't be done first?

lib/rdoc/markup/pre_process.rb
```diff
@@ -122,21 +119,49 @@ def handle(text, code_object = nil, &block)
...
+
+  # Apply directives to a code object
+
+  def run_pre_processes(comment_text, code_object, start_line_no, type)
+    comment_text, directives = parse_comment(comment_text, start_line_no, type)
+    directives.each do |directive, (param, line_no)|
+      handle_directive('', directive, param, code_object)
+    end
+    if code_object.is_a?(RDoc::AnyMethod) && (call_seq, = directives['call-seq']) && call_seq
+      code_object.call_seq = call_seq.lines.map(&:chomp).reject(&:empty?).join("\n") if call_seq
```
st0012: Is this `if call_seq` still necessary given the `&& call_seq` above?

lib/rdoc/parser/c.rb
```diff
@@ -621,7 +619,6 @@ def find_body(class_name, meth_name, meth_obj, file_content, quiet = false)
       override_comment = find_override_comment class_name, meth_obj
       comment = override_comment if override_comment
 
-      comment.normalize
```
st0012, line 624: Love these removals 👍

lib/rdoc/parser/c.rb
```diff
@@ -823,10 +814,10 @@ def find_override_comment(class_name, meth_obj)
     comment = if @content =~ %r%Document-method:
                                 \s+#{class_name}#{prefix}#{name}
                                 \s*?\n((?>.*?\*/))%xm then
-                "/*#{$1}"
+                "/*\n#{$1}"
               elsif @content =~ %r%Document-method:
```
st0012, line 818: Can we also merge the condition like: `if @content =~ regexp1 || @content =~ regexp2`

lib/rdoc/parser/c.rb
```diff
@@ -1102,35 +1093,10 @@ def load_variable_map(map_name)
   # Both :main: and :title: directives are deprecated and will be removed in RDoc 7.
 
   def look_for_directives_in(context, comment)
-    @preprocess.handle comment, context do |directive, param|
-      case directive
-      when 'main' then
```
st0012, line 1107: I still want to print these deprecation warnings until we drop them in v7 release

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -505,63 +509,67 @@ def add_extends(names, line_no) # :nodoc:
...
-    # defined in c and ruby just like the old parser did.
-    if meth.name == 'initialize' && !singleton
-      if meth.dont_rename_initialize
-        meth.visibility = :protected
-      else
-        meth.name = 'new'
-        meth.singleton = true
-        meth.visibility = :public
-      end
+  private def internal_add_method(method_name, container, comment:, dont_rename_initialize: false, directives:, modifier_comment_lines: nil, line_no:, visibility:, singleton:, params:, calls_super:, block_params:, tokens:) # :nodoc:
```
st0012, line 535: If `modifier_comment_lines` always means `start_line, args_end_line, end_line`, I prefer making them keyword arguments here directly.

test/rdoc/rdoc_parser_simple_test.rb
```diff
@@ -26,7 +26,7 @@ def test_initialize_metadata
 
     assert_includes @top_level.metadata, 'unhandled'
 
-    assert_equal ":unhandled: \n", parser.content
+    assert_equal "", parser.content
```
st0012:
```suggestion
    assert_empty parser.content
```

test/rdoc/rdoc_parser_c_test.rb
```diff
@@ -1629,11 +1624,11 @@ def test_handle_singleton
   def test_look_for_directives_in
     parser = util_parser
 
-    comment = RDoc::Comment.new "# :other: not_handled\n"
+    comment = RDoc::Comment.new "* :other: not_handled\n"
 
     parser.look_for_directives_in @top_level, comment
 
-    assert_equal "# :other: not_handled\n", comment.text
+    assert_equal "", comment.text
```
st0012:
```suggestion
    assert_empty comment.text
```

lib/rdoc/parser/c.rb
tompng: Warn is not removed. The same warning still exist in `pre_process.rb`. It was a duplicated code.

This method `look_for_directives_in` calls `@preprocess.run_pre_processes`, it calls `PreProcess#handle_directive`, warning is threre

st0012: Ah right. I forgot that I duplicated the warning in 2 places. This is great!

lib/rdoc/markup/pre_process.rb
tompng: Thanks. Removed this useless condition

lib/rdoc/parser/prism_ruby.rb
tompng: The order is not important, and I added `.uniq` to avoid processing modifier comments several times
```ruby
def f(arg1, arg2) # :nodoc:
end # :nodoc:
#=> [1, 1, 2] #=> [1, 2]


def f(arg1, arg2); end # :nodoc:
#=> [16, 16, 16] #=> [16]
```

st0012 (approved):
Thank you! Let's rebase and merge it 👍

PR #1150, opened 2024-08-05
Use ascii chatacter in HTML file to fix ruby ci failure

tompng:
Running test with ascii encoding fails `LANG="" rake test`
```
Error: test_document_with_dry_run(TestRDocRDoc): RDoc::Error: error generating index.html: "\xE2" on US-ASCII (Encoding::InvalidByteSequenceError)
```
http://ci.rvm.jp/logfiles/brlog.trunk.20240805-004507#L1586

Just like RDoc uses `&rarr;` instead of `→` in output html, this pull request changes `☰` to `&#9776;` to make the output html ascii-only again.


olleolleolle (approved):
This fixes the hamburger menu icon ("trigram for heaven") in all situations.

PR #1151, opened 2024-08-05
Basic server mode support

st0012:
Currently, `rdoc` doesn't support server mode, which makes both using it and developing it extremely inconvenient.

(`ri --server` is RI's web interface, which lets users read indexed RI documents in the browser. It does NOT run current project's RDoc documentation in server mode. To avoid confusion, I moved `RDoc::Servlet` to `RDoc::RI::Servlet`.)

So in this PR, I want to introduce a very rough server support to RDoc:

- Can be used with `rdoc --server` or `rdoc -s`
    - It's marked as experimental for now as it's very slow (see below)
    - The port is bound to `8080` atm but we can make it assignable too. No strong opinion
- Will use `./tmp` to temporarily host the generated files so they don't override the original doc folder
- It reparses and regenerates files on EVERY SINGLE REQUEST, which is pretty slow, but still faster than running `rdoc` or `rake rdoc` repeatedly
    - We can add smarter reparsing/regenerating support, but it will first take some refactoring as the current design doesn't account for those


lib/rdoc/generator/darkfish.rb
```diff
@@ -259,6 +259,22 @@ def generate
     raise
   end
 
+  def generate_for_server
+    write_style_sheet
```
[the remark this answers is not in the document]
st0012: 👍 Updated

lib/rdoc/options.rb
```diff
@@ -1116,6 +1122,15 @@ def parse argv
 
       opt.separator nil
 
+      opt.on(
+        "--server[=PORT]",
+        Integer,
+        "[Experimental] Run WEBrick server with generated documentation.",
```
colby-swandale, line 1128: I feel we can remove the `[Experiment]` flag, we're not shipping something we're not expecting will work or introduce any breaking changes.

lib/rdoc/rdoc.rb
```diff
@@ -507,6 +514,26 @@ def document options
...
 
+  def start_server
+    begin
+      require 'webrick'
+    rescue LoadError
+      abort "webrick is not found. You may need to `gem install webrick` to install webrick."
+    end
+
+    # Change the output directory to tmp so it doesn't overwrite the current documentation
+    Dir.chdir "tmp" do
```
colby-swandale, line 525: Discussed this in Slack. Let's look at utilizing `mktmpdir` instead of `$CWD/test`

lib/rdoc/options.rb
st0012: My interpretation of "experimental" is: the feature works but may not live up to expectation at the moment and has a slight chance to be removed if it's proven problematic.

In this case, I don't expect it to be removed later but it doesn't have the maturity to work like `yard server` or some static site generators' server mode.

lib/rdoc/server.rb
```diff
@@ -0,0 +1,182 @@
...
+    @generator.clear_data
+
+    # Reparse the files
+    @rdoc.parse_files(@rdoc.options.files)
+
+    # Update the data from store
+    @generator.update_data_from_store
+
+    # Regenerate the documentation and asserts
+    @generator.generate(server_mode: true)
```
st0012, line 152: Through implementing this feature I learned a lot about how data are passed around, which sometimes simply duplicate in multiple classes. And I plan to do some refactor around them.

faustind:
Hi @st0012 

Please tell me: Will your work here allow one to consult the standard library's documentation locally using `rdoc`? 

`ri`'s server mode seems to require the installation of `webrick`. Will `rdoc --server` have the same dependency?

st0012:
Closing in favor or #1620

PR #1153, opened 2024-08-10
[doc] Mention `.document` file in README

okuramasafumi:
Rdoc has `.document` file in its root directory, allowing us to determine which files are the target for documentation. However, this fact is not mentioned anywhere in the doc including README, which makes it quite difficult to utilize this feature. This commit adds a mention to `.document` file in README so that users can find the feature more easily.

One thing I'm not sure is if it's the right place. Currently README includes some usage information, which can be a separate file.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -60,6 +60,10 @@ To generate documentation programmatically:
   rdoc.document options
   # see RDoc::RDoc
 
+You can specify the target files for document generation with +.document+ file in the project root directory.
++.document+ file contains a list of file and directory names including comment lines starting with '#'.
+See https://github.com/ruby/rdoc/blob/master/.document as an example.
+
 == Writing Documentation
 
 To write documentation for RDoc place a comment above the class, module,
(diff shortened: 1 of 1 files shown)


colby-swandale (approved):

README.rdoc
```diff
@@ -60,6 +60,10 @@ To generate documentation programmatically:
   rdoc.document options
   # see RDoc::RDoc
 
+You can specify the target files for document generation with +.document+ file in the project root directory.
++.document+ file contains a list of file and directory names including comment lines starting with '#'.
```
st0012, line 64: nit: I think we can use code highlight for `#`

```suggestion
+.document+ file contains a list of file and directory names including comment lines starting with +#+.
```

st0012 (approved):
Thanks!

README.rdoc
```diff
@@ -60,6 +60,10 @@ To generate documentation programmatically:
   rdoc.document options
   # see RDoc::RDoc
 
+You can specify the target files for document generation with +.document+ file in the project root directory.
++.document+ file contains a list of file and directory names including comment lines starting with +#+.
```
st0012, line 64: Uh I didn't know `+` doesn't work for it, sorry 🤦‍♂️
 
```suggestion
+.document+ file contains a list of file and directory names including comment lines starting with '#'.
```

PR #1155, opened 2024-08-19
Avoid potentially loading the same extension from different versions of the same gem

deivid-rodriguez:
Avoids warnings like

```
/path/to/ruby/3.3.4/lib/ruby/gems/3.3.0/gems/rbs-3.4.0/lib/rdoc/discover.rb:10: warning: method redefined; discarding old scan
/path/to/ruby/3.3.4/lib/ruby/gems/3.3.0/gems/rbs-3.5.1/lib/rdoc/discover.rb:10: warning: previous definition of scan was here
```

### DIFF
--- a/lib/rdoc/rdoc.rb
+++ b/lib/rdoc/rdoc.rb
@@ -544,7 +544,7 @@ end
 begin
   require 'rubygems'
 
-  rdoc_extensions = Gem.find_files 'rdoc/discover'
+  rdoc_extensions = Gem.find_latest_files 'rdoc/discover'
 
   rdoc_extensions.each do |extension|
     begin


olleolleolle:
Today I learned: https://github.com/rubygems/rubygems/blob/633c1bffd4aa4b2eb9b37576fcb16d6f0a40ae93/lib/rubygems.rb#L529 is super-nice.

deivid-rodriguez:
I don't really like this "auto-discovery" mechanism for loading plugins. Minitest does something similar but allows to toggle off auto-discovery with an ENV variable, and also has recently added `Minitest.register_plugin` for a more explicit approach. I'd be happy if rdoc follows the same path, but I think this little change is an improvement regardless!

olleolleolle (approved):
Let's use this!

deivid-rodriguez:
By the way, funny that Gem.find_latest_files docs using exactly [this example](https://github.com/rubygems/rubygems/blob/633c1bffd4aa4b2eb9b37576fcb16d6f0a40ae93/lib/rubygems.rb#L521), yet the real code was not actually using it 😅. 

Thanks for having a look @olleolleolle!

PR #1156, opened 2024-08-19
Align behaviour between `bundle exec rdoc` and `rake rdoc`

### LINKED ISSUE #1123: [Proposal] `bundle exec rdoc` command should work for rdoc itself

st0012:
By using `.rdoc_options` to specify the RDoc options, we can make most of the rake task options the in-repo defaults.

This allows us to use the same options for both `bundle exec rdoc` and `rake rdoc` and get a consistent experience.

Closes #1123


Rakefile
```diff
@@ -21,10 +21,8 @@ task :default => :test
 
 task rdoc: :generate
 RDoc::Task.new do |doc|
-  doc.main = 'README.rdoc'
-  doc.title = "rdoc #{RDoc::VERSION} Documentation"
-  doc.rdoc_dir = '_site' # for github pages
-  doc.rdoc_files = FileList.new %w[lib/**/*.rb *.rdoc *.md doc/rdoc/markup_reference.rb] - PARSER_FILES
+  # RDoc task defaults to /html and overrides the op_dir option in .rdoc_options
+  doc.rdoc_dir = "_site"
```
olleolleolle, line 25:
```suggestion
  doc.rdoc_dir = "_site" # for github pages
```

olleolleolle (commented):
How about keeping the contextual knowledge that _site is chosen for GitHub Pages?

.rdoc_options
```diff
@@ -0,0 +1,8 @@
+exclude:
+  - lib/rdoc/rd/block_parser.ry
+  - lib/rdoc/rd/inline_parser.ry
+  - lib/rdoc/markdown.kpeg
+  - lib/rdoc/markdown/literals.kpeg
+op_dir: _site # for GH pages and should match the config of RDoc task in Rakefile
```
olleolleolle, line 6: Minor: by using the casing in the branding, the reader has a higher chance to differentiate what kinds of pages we are referring to

```suggestion
op_dir: _site # for GitHub Pages and should match the config of RDoc task in Rakefile
```

Rakefile
```diff
@@ -21,10 +21,8 @@ task :default => :test
 
 task rdoc: :generate
 RDoc::Task.new do |doc|
-  doc.main = 'README.rdoc'
-  doc.title = "rdoc #{RDoc::VERSION} Documentation"
-  doc.rdoc_dir = '_site' # for github pages
-  doc.rdoc_files = FileList.new %w[lib/**/*.rb *.rdoc *.md doc/rdoc/markup_reference.rb] - PARSER_FILES
+  # RDoc task defaults to /html and overrides the op_dir option in .rdoc_options
+  doc.rdoc_dir = "_site" # for github pages
```
olleolleolle, line 25:
```suggestion
  doc.rdoc_dir = "_site" # for GitHub Pages
```

olleolleolle (commented):
I liked that the comments follow into the configuration file. I added tiny changes, minor.

olleolleolle (approved):
Thanks, this is super!

I recommend this be merged!

p8:
I'd love to see this getting merged so we can move the following to .rdoc_options in Rails:
https://github.com/rails/rails/blob/4bb2227640531c877e30cc96f10df0c298dc3331/railties/lib/rails/api/task.rb#L9-L117
If RDoc can generate documentation without having to install the dependencies (so just `rdoc` instead of `bundle exec rdoc`, we could run RDOC for older versions of gems/rails without having to install old unsupported/deprecated dependencies.

.rdoc_options
```diff
@@ -0,0 +1,8 @@
+exclude:
+  - lib/rdoc/rd/block_parser.ry
+  - lib/rdoc/rd/inline_parser.ry
+  - lib/rdoc/markdown.kpeg
+  - lib/rdoc/markdown/literals.kpeg
+op_dir: _site # for GitHub Pages and should match the config of RDoc task in Rakefile
```
nobu, line 6: `--op-dir` option is not saved by `--write-options`.
Is this expected?

st0012: I don't think it is. But the bigger problem is that the presence of options (and some of their names) are different across Rake task, command line, and `.rdoc_options`. I think we should do a review and consolidation on them next year.

PR #1160, opened 2024-08-24
Make summary triangle appear in correct place when summary text overflows to next line

paracycle:
Before:

image

After:

image


st0012 (approved):
Thanks!

PR #1162, opened 2024-08-24
Make darkfish more responsive and readable on mobile devices

st0012:
- Make the sidebar toggle fixed on all devices
- Prevent default zooming on mobile devices
- Improve sidebar opening on mobile devices

### Before

https://github.com/user-attachments/assets/954d2686-9f83-4c40-94e4-97c00b791652

### After

https://github.com/user-attachments/assets/2cf087dd-977d-43cf-8841-f13c513852c8

### DIFF
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -219,17 +219,15 @@ nav a:hover {
 #navigation-toggle {
   z-index: 1000;
   font-size: 2em;
-  position: absolute;
+  display: block;
+  position: fixed;
+  top: 10px;
+  left: 20px;
 }
 
 #navigation-toggle[aria-expanded="true"] {
   top: 10px;
   left: 250px;
-}
-
-#navigation-toggle[aria-expanded="false"] {
-  top: 10px;
-  left: 20px;
 }
 
 nav ul li details {
@@ -265,6 +263,7 @@ main {
   font-size: 16px;
   line-height: 1.6;
   color: var(--text-color);
+  box-sizing: border-box;
 }
 
 @media (min-width: 1024px) {
@@ -715,4 +714,31 @@ main .attribute-access-type {
   font-family: var(--font-code);
 }
 
+@media (max-width: 480px) {
+  nav {
+    width: 100%;
+  }
+
+  main {
+    margin: 1em auto;
+    padding: 0 1em;
+    max-width: 100%;
+  }
+
+  #navigation-toggle {
+    right: 10px;
+    left: auto;
+  }
+
+  #navigation-toggle[aria-expanded="true"] {
+    left: auto;
+  }
+
+  table {
+    display: block;
+    overflow-x: auto;
+    white-space: nowrap;
+  }
+}
 /* @end */
+


lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -715,4 +717,50 @@ main .attribute-access-type {
   font-family: var(--font-code);
 }
 
+@media (max-width: 1023px) {
+  body {
+    font-size: 14px;
```
MatheusRich, line 722: Any particular reason not to use a bigger font size like the browser default 16px? I know it takes more space, but reading on mobile is harder (in particular for us, wearing glasses 😅)

lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -219,7 +219,8 @@ nav a:hover {
 #navigation-toggle {
   z-index: 1000;
   font-size: 2em;
-  position: absolute;
+  display: block;
+  position: fixed;
```
MatheusRich, line 223: Non-blocking: I actually like the sidebar on the left, even if it moves. I think most websites work like that.

lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -265,6 +266,7 @@ main {
   font-size: 16px;
   line-height: 1.6;
   color: var(--text-color);
+  box-sizing: border-box;
```
MatheusRich, line 269: This could probably go on the `html` selector if it's not too disruptive. I think most sites use it because it's more intuitive.

st0012: Moving it to `body` actually would make the mobile view zoom in again but I'm not sure why. So I'll leave it here for now but I'd appreciate a later PR to put it to a better place.

lib/rdoc/generator/template/darkfish/css/rdoc.css
st0012: Sidebar is still on the left.

MatheusRich: Sorry. I meant the icon.

MatheusRich: In any case, we could add a background color to the icon, so when it is on top of text it is still visible. I think I had that in the first PR

lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -715,4 +717,35 @@ main .attribute-access-type {
...
+  }
+
+  main {
+    margin: 1em auto;
+    padding: 0 1em;
+    max-width: 100%;
+  }
+
+  #navigation-toggle {
+    top: 10px;
```
colby-swandale, line 732: This rule is overwritten by `#navigation-toggle[aria-expanded="false"]` and becomes redundant.

lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -715,4 +717,35 @@ main .attribute-access-type {
...
+    right: 10px;
+  }
+
+  #navigation-toggle[aria-expanded="true"] {
```
colby-swandale, line 736: We're setting `left: auto` for when `aria-expanded` is both true and false, can this be collapsed into a single rule?

st0012: Updated 👍

colby-swandale (approved):

PR #1165, opened 2024-08-27
Add home page link to output message

st0012:
The link is clickable in the terminal and opens the home page in the browser:

Screenshot 2024-08-27 at 23 12 47

This will make it easier for users to visit the generated document.

### DIFF
--- a/lib/rdoc/rdoc.rb
+++ b/lib/rdoc/rdoc.rb
@@ -520,6 +520,7 @@ The internal error was:
       Dir.chdir @options.op_dir do
         unless @options.quiet then
           $stderr.puts "\nGenerating #{@generator.class.name.sub(/^.*::/, '')} format into #{Dir.pwd}..."
+          $stderr.puts "\nYou can visit the home page at: \e]8;;file://#{Dir.pwd}/index.html\e\\file://#{Dir.pwd}/index.html\e]8;;\e\\"
         end
 
         @generator.generate


colby-swandale (approved):
This change is good, but I wonder if there is a way to add a test for this?

st0012:
Currently there is no tests covering `RDoc#generate` (which can be verified by adding a `raise` in it and still see all tests pass). If we were to add one, it'll require a new set of test infra to properly assert and clean up generated files.
It'll definitely be a good thing to have as it'll enable true end-to-end testing for RDoc. But considering our current capacity, I'd like to prioritize other improvements first (e.g. #1176)

colby-swandale:
Sounds good! @st0012

st0012:
@colby-swandale opened https://github.com/ruby/rdoc/issues/1180

nobu:
This looks useless unless printing to a tty.
https://github.com/ruby/ruby/actions/runs/22793981161/job/66125592137#step:16:819
> You can visit the home page at: ]8;;file:///home/runner/work/ruby/ruby/html/index.html\file:///home/runner/work/ruby/ruby/html/index.html]8;;\

nobu:
Another issue: this message is shown even when that file is not generated, the output format is not HTML.

nobu:
I think the generated entry page name should also be in the style that asks the generator.

PR #1174, opened 2024-09-06
Add scroll-margin-top on all :target elements

omegahm:
This sets `scroll-margin-top` on all `h1` through `h6` when applied with the element state `target`. This happens when you click a link and are taking to that anchor (with `#`). 

`scroll-margin-top` adds a margin of `1rem` where it jumps to, adding a bit of space at the top of the page.

| Before | After |
|-|-|
| Screenshot 2024-09-06 at 21 44 07 | Screenshot 2024-09-06 at 21 43 51 |


st0012 (approved):
This is great, thank you!

PR #1175, opened 2024-09-06
Use pointer cursor for navigation toggle

omegahm:
~~This hides the `#navigation-toggle` class, if the screen is larger than 1024px.~~

It also adds a proper pointer icon, if you can indeed see and click it.

Screenshot 2024-09-06 at 22 08 11

### DIFF
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -223,6 +223,7 @@ nav a:hover {
   position: fixed;
   top: 10px;
   left: 20px;
+  cursor: pointer;
 }
 
 #navigation-toggle[aria-expanded="true"] {


lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -343,6 +344,9 @@ main h6 {
 /* 11. Media Queries */
 @media (min-width: 1024px) {
   /* Styles for larger screens */
+  #navigation-toggle {
+    display: none;
```
st0012, line 348: Thanks for the PR! I love the pointer improvement, but I think the navigation toggle design needs a more thorough rework. So can we drop this part for now?

omegahm: Do you want me to take a stab at this thorough rework of the navigation bar/toggle?

st0012: If you can help, I'd definitely appreciate it 🙏 

IMO, on narrower screens, a sticky sidebar should appear to hold the toggle button and maybe a few other info (e.g. logo, search bar, page title...etc.). For the first iteration tho, just having the toggle button would be good already. We can add other stuff later.

### Examples

#### Rust doc (width <= 700px)

Screenshot 2024-09-08 at 13 09 35

#### Python doc (width < 1024px)

Screenshot 2024-09-08 at 13 13 10

st0012 (approved):
Thanks!

Issue #1164, opened 2024-08-26
RDoc color scheme


colby-swandale:
I think it's time to take a look at having something more friendly and accessible to the eyes. Let's use this ticket to capture some ideas and see if we can break some ground on putting a new set of colors together for RDoc.

Long live the green!

st0012:
We're also improving RDoc's development process (e.g. #1151). But before that's merged, you can use these steps to generate the documentation and see the changes:

1. Run `bundle install`
2. Run `bundle exec rake`
3. Run `bundle exec rdoc` or `bundle exec rake rdoc` to generate the documentation

adam12:
Relaying some insights from discussions in Discord: the Lato font is incredibly thin and somewhat hard to read. Considerations for using system font stack instead?

jaredcwhite:
Seconded…and even just adjusting the Lato font from weight 300 to normal 400 could help tremendously. This is a screenshot from my iPad, it's so hard for me to read. 🫣

![image](https://github.com/user-attachments/assets/891e64ab-63ab-4058-a971-3c21fd6c2665)

havenwood:
To me it seems even a bit hard to read than the screenshot on an iPad screen. Especially with grey, a thicker font would be much appreciated!

st0012:
Thanks for the feedback. I hope #1197 can address the issue 🙂 

Also, I'd appreciate if you can open new issues for reporting different categories of problems as it'll 1) make them more searchable in the future, and 2) allow maintainers to close the issue once the problem has been addressed.

PR #1176, opened 2024-09-08
Improve method source toggling

st0012:
1. Source toggling control is now always displayed
    - If we think the buttons take too much space or are distracting, I'd be happy to accept ideas/PRs for styling improvements. But hiding them by default simply made this feature underutilized.
1. Source will be displayed directly under the method method
    - This matches rubyapi.org's behaviour and makes a more natural experience. When viewing a method with long description, users won't need to click the toggle, and then scroll all the way down to see the source.
1. Clicking the method name will not toggle the source anymore. I made this change so we can later make clicking method names link to the methods instead, like rubyapi.org and many other language documentations do.

### Before

https://github.com/user-attachments/assets/ca68c0d0-147b-421f-9e04-904366856daf

### After

https://github.com/user-attachments/assets/ea6ed17f-930a-4d4e-a298-79a535c94abc

### DIFF
--- a/lib/rdoc/generator/template/darkfish/class.rhtml
+++ b/lib/rdoc/generator/template/darkfish/class.rhtml
@@ -109,7 +109,11 @@
                     gsub(/(.*)[-=]&gt;/, '\1&rarr;') %>
             </span>
             <%- if i == 0 and method.token_stream then -%>
-            <span class="method-click-advice">Toggle source</span>
+            <div class="method-controls">
+              <details class="method-click-advice">
+                <summary>Source</summary>
+              </details>
+            </div>
             <%- end -%>
           </div>
           <%-   end -%>
@@ -119,10 +123,14 @@
           </div>
           <%- else -%>
           <div class="method-heading">
-            <span class="method-name"><%= h method.name %></span><span
-              class="method-args"><%= h method.param_seq %></span>
+            <span class="method-name"><%= h method.name %></span>
+            <span class="method-args"><%= h method.param_seq %></span>
             <%- if method.token_stream then -%>
-            <span class="method-click-advice">Toggle source</span>
+            <div class="method-controls">
+              <details class="method-click-advice">
+                <summary>Source</summary>
+              </details>
+            </div>
             <%- end -%>
           </div>
           <%- end -%>
--- a/lib/rdoc/generator/template/darkfish/class.rhtml
+++ b/lib/rdoc/generator/template/darkfish/class.rhtml
@@ -130,6 +130,11 @@
 
         <%- unless method.skip_description? then -%>
         <div class="method-description">
+          <%- if method.token_stream then -%>
+          <div class="method-source-code" id="<%= method.html_name %>-source">
+            <pre><%= method.markup_code %></pre>
+          </div>
+          <%- end -%>
           <%- if method.comment then -%>
           <%= method.description.strip %>
           <%- else -%>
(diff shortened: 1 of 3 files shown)


etiennebarrie:
> 3\. Clicking the method name will not toggle the source anymore. I made this change so we can later make clicking method names link to the methods instead, like rubyapi.org and many other language documentations do.

Even if we decide to keep the source toggle on the method name, I'd love to get the URL on the method name so that at least we can right-click, copy link location and share that more easily.

st0012:
Does this look good to you? (ofc, I'm not going to make all method names green)
https://github.com/user-attachments/assets/e0ba6324-deaa-4e49-ae58-5d187796d6b1

rwstauner (approved):
These are great improvements, thank you!

p8:
Maybe move the toggle to the bottom, so the toggle and the source code are closer together?
SDoc currently does this:
image

st0012:
> I think this looks great, but I have one concern.

That's a good point. I'll think about how to avoid the confusion.

> Maybe move the toggle to the bottom, so the toggle and the source code are closer together?

Since it's been at the top for a long time, moving it to the bottom would require a huge change on users' behaviour. It's also not uncommon to see methods having very long description and examples in Ruby (e.g. `Array#[]`, `Array#fill`...etc.), and IMO being able to click and see the source toggle without scrolling further down would be a better experience.

st0012:
To help distinguishing source code blocks and example code blocks, I ended up choosing to add `Example` label to code blocks interpreted as Ruby:


Screenshot 2024-09-12 at 12 48 20

Screenshot 2024-09-12 at 12 48 50

### Notes

- It's easier to add `Source` label but I feel it a bit repetitive when you just clicked `Source` and then see the shown block marked as `Source` again.
- Styling options are limited because the html of example code blocks are NOT controlled by the Darkfish generator but directly from RDoc. So if we were to change its html structure, it could break gems that use RDoc (e.g. SDoc)
    - Therefore, I decided to just add a new class and use its `::before` element to add the label.

p8:
> Styling options are limited because the html of example code blocks are NOT controlled by the Darkfish generator but directly from RDoc. So if we were to change its html structure, it could break gems that use RDoc (e.g. SDoc)

You should be able to change the background of the source with the following:
```css
.method-source-code pre {
  background: #ebead5;
}
```

image

p8:
Looks great!

colby-swandale:
I have a couple of items: 

* The various yellows on the page doesn't seem to blend well together and makes the page distracting. Could we tweak the colors a bit to make it easier on the eyes?

Screenshot 2024-10-03 at 11 45 23 PM

* The arrow/collaspe icon next to the `Source` link is a bit confusing to me. It comes across as a play button for audio/video nearly everytime I look at It. Could we look to remove it and give the link an active/hover state?

st0012:
I made their color match and change it to fit the current color scheme better:

Screenshot 2024-10-03 at 17 25 23

I'm not happy with the color but I think further improvements on it belongs to #1164

> Could we look to remove it and give the link an active/hover state?

I think this is more subject to personal taste? To me it's a clear sign that it's a `details` element and I can expand it. It's the default triangle that comes with the element and it's in the sidebar too.

I don't mind it being redesigned later but I don't see the value of introducing yet another UI effect to remind users it's click/expandable.

st0012:
@colby-swandale Can you give this another look please?

colby-swandale (approved):

PR #1181, opened 2024-09-28
Only let browser search through source code until it's expanded

Earlopain:
Something that's been bothering me is that while the source code is not visible by default, the browser still jump to it when searching. Adding the `visible` property prevents this.

Test it out yourself:
* `bundle exec rdoc`
* open `_site/RDoc.html`
* Search for `NameError`

Before, you will get a match from `load_yaml` source code, after you only get the match when that methods source code is expanded:
image
image


colby-swandale (approved):

st0012 (approved):
It's a great improvement, thank you 👍

PR #1183, opened 2024-09-30
Display class ancestors in the sidebar

alexisbernard:
The purpose is to list all class ancestors in a new theme (see PR #1182).

### DIFF
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -711,6 +711,20 @@ class RDoc::ClassModule < RDoc::Context
     @superclass = superclass
   end
 
+  ##
+  # Get all super classes of this class in an array. The last element might be
+  # a string if the name is unknown.
+
+  def super_classes
+    result = []
+    parent = self
+    while parent = parent.superclass
+      result << parent
+      return result if parent.is_a?(String)
+    end
+    result
+  end
+
   def to_s # :nodoc:
     if is_alias_for then
       "#{self.class.name} #{self.full_name} -> #{is_alias_for}"
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -795,4 +795,22 @@ class RDoc::Generator::Darkfish
 
     extracted_text[0...150].gsub(/\n/, " ").squeeze(" ")
   end
+
+  def generate_ancestor_list(ancestors, klass)
+    return '' if ancestors.empty?
+
+    ancestor = ancestors.shift
+    content = +'<ul><li>'
+
+    if ancestor.is_a?(RDoc::NormalClass)
+      content << "<a href=\"#{klass.aref_to ancestor.path}\">#{ancestor.full_name}</a>"
+    else
+      content << ancestor.to_s
+    end
+
+    # Recursively call the method for the remaining ancestors
+    content << generate_ancestor_list(ancestors, klass)
+
+    content << '</li></ul>'
+  end
 end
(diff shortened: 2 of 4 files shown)


alexisbernard:
I believe the right name is `super_classes` but it is inconsistent with `superclass`. However `Method#super_method` exists. Let me know if you have a preference.

st0012:
Thanks for the PR. I think this is a great addition to the theme's functionalities.
I'd like to propose a simplification on the implementation though:

```diff
diff --git a/lib/rdoc/generator/darkfish.rb b/lib/rdoc/generator/darkfish.rb
index 5709fabf..cdf7fdd2 100644
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -795,4 +795,23 @@ class RDoc::Generator::Darkfish
 
     extracted_text[0...150].gsub(/\n/, " ").squeeze(" ")
   end
+
+  def generate_ancestor_list(ancestors, klass)
+    return '' if ancestors.empty?
+
+    ancestor = ancestors.shift
+    content = +'<ul><li>'
+
+    if ancestor.is_a?(RDoc::NormalClass)
+      content << "<a href=\"#{klass.aref_to ancestor.path}\">#{ancestor.full_name}</a>"
+    else
+      content << ancestor.to_s
+    end
+
+    # Recursively call the method for the remaining ancestors
+    content << generate_ancestor_list(ancestors, klass)
+
+    content << '</li></ul>'
+    content
+  end
 end
diff --git a/lib/rdoc/generator/template/darkfish/_sidebar_parent.rhtml b/lib/rdoc/generator/template/darkfish/_sidebar_parent.rhtml
index 7bd1a7ca..6808b2bf 100644
--- a/lib/rdoc/generator/template/darkfish/_sidebar_parent.rhtml
+++ b/lib/rdoc/generator/template/darkfish/_sidebar_parent.rhtml
@@ -1,42 +1,6 @@
-<%- if klass.type == 'class' && (ancestors = klass.super_classes).any? then %>
+<%- if klass.type == 'class' && (ancestors = klass.super_classes).any? -%>
 <div id="parent-class-section" class="nav-section">
   <h3>Ancestors</h3>
-  <% if ancestors.size > 1 %>
-    <% parent = ancestors.shift %>
-    <ul>
-      <li>
-        <details>
-          <summary>
-            <% if parent.is_a?(RDoc::NormalClass) %>
-              <a href="<%= klass.aref_to parent.path -%>"><%= parent.full_name -%></a>
-            <% else %>
-              <%= parent -%>
-            <% end %>
-          </summary>
-
-          <% for ancestor in ancestors %>
-            <ul>
-              <li>
-                <% if ancestor.is_a?(RDoc::NormalClass) %>
-                  <a href="<%= klass.aref_to ancestor.path -%>"><%= ancestor.full_name -%></a>
-                <% else %>
-                  <%= ancestor -%>
-                <% end %>
-          <% end %>
-
-          <% for ancestor in ancestors # Sorry, templates cannot be recursive %>
-              </li>
-            </ul>
-          <% end %>
-        </details>
-      </li>
-    </ul>
-  <% else %>
-    <%- if klass.superclass and not String === klass.superclass then -%>
-    <p class="link"><a href="<%= klass.aref_to klass.superclass.path %>"><%= klass.superclass.full_name %></a>
-    <%- else -%>
-    <p class="link"><%= klass.superclass %>
-    <%- end -%>
-  <% end %>
+  <%= generate_ancestor_list(ancestors, klass) %>
 </div>
 <%- end -%>
```

This will remove the expanding toggle, which IMO isn't necessary as expanded list will make the ancestors list instantly understandable. We can make it folded later based on user feedback.

alexisbernard:
I applied the patch and I removed the last line of `generate_ancestor_list` since `content` is already implicitly returned.

I also expanded the tree by default when I started building this feature. But after using it, it felt too repetitive because in most cases ancestors are only Object and BasicObject. Sure, let's adjust according to other user's feedback.

st0012:
>  it felt too repetitive because in most cases ancestors are only Object and BasicObject.

To be honest, I feel the same. But my thinking is that if we only display it when it's patched (in the context of gems), it may be confusing to people new to Ruby as the timing to have parent classes will vary in an implicit way.

st0012 (approved):
This is amazing, thank you!

PR #1184, opened 2024-09-30
Fix ToRdoc#accept_table

tompng:
Document of `Integer#ceil` in ruby 3.4.0dev have table. RDoc fails to show tables in text format.

Fix `ri Integer#ceil`
```
# ruby -v
ruby 3.4.0dev (2024-09-27T17:45:22Z master 027ef60500) +PRISM [x86_64-linux]
# ri Integer#ceil
/opt/ruby/lib/ruby/3.4.0+0/rdoc/markup/to_rdoc.rb:265:in 'Array#zip': wrong argument type NilClass (must respond to :each) (TypeError)

    @res << header.zip(widths, aligns) do |h, w, a|
```

Fix IRB crash
```
irb(main):001> 1.[TAB][TAB]
/opt/ruby/lib/ruby/3.4.0+0/rdoc/markup/to_rdoc.rb:265:in 'Array#zip': wrong argument type NilClass (must respond to :each) (TypeError)
```

### DIFF
--- a/lib/rdoc/markup/to_rdoc.rb
+++ b/lib/rdoc/markup/to_rdoc.rb
@@ -249,8 +249,8 @@ class RDoc::Markup::ToRdoc < RDoc::Markup::Formatter
   # Adds +table+ to the output
 
   def accept_table header, body, aligns
-    widths = header.zip(body) do |h, b|
-      [h.size, b.size].max
+    widths = header.zip(*body).map do |cols|
+      cols.map(&:size).max
     end
     aligns = aligns.map do |a|
       case a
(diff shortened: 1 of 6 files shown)


colby-swandale:
Could we separate the various fixes into their own commits, it would make reviewing each fix much easier

tompng:
> Could we separate the various fixes into their own commits

I removed The change to `lib/rdoc/markup/to_tt_only.rb` that adds `alias accept_table do_nothing`.
Now this pull request only contain one fix: `ToRdoc#accept_table`. This single fix will solve both `ri Integer#ceil` crash and `irb` completion crash which is the same cause.

st0012 (approved):
Thank you 👍

PR #1185, opened 2024-09-30
Improve RDoc pages heading levels order

antoinem:
- Add h1 heading on pages where it's missing
- Fix heading level order (h1 > h2 > …)

### DIFF
--- a/ExampleRDoc.rdoc
+++ b/ExampleRDoc.rdoc
@@ -1,3 +1,5 @@
+= Example RDoc
+
 This document contains example output to show RDoc styling.  This file was
 created from a RDoc Markup file.
 
--- a/ExampleRDoc.rdoc
+++ b/ExampleRDoc.rdoc
@@ -1,4 +1,4 @@
-= Example RDoc
+= Example \RDoc
 
 This document contains example output to show RDoc styling.  This file was
 created from a RDoc Markup file.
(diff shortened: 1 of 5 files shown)


History.rdoc
```diff
@@ -1,4 +1,6 @@
-=== 5.1.0 / 2017-02-24
+= History
```
st0012, line 1: Why do we need this change?

History.rdoc
```diff
@@ -202,7 +204,7 @@
     scheme.  See RDoc::Markup@Links for details.  Issue #93 by Tim Pease.
   * Ignore empty call-seq for methods.  Improved deduplication of C methods
     sharing the same C function.  This allows the method heading to show
-    up correctly for String#== and #===.  Bug #244 by Neurogami.
+    up correctly for String#== and #==.  Bug #244 by Neurogami.
```
st0012, line 207: The content of entries shouldn't be changed.

st0012 (changes requested):

ExampleRDoc.rdoc
```diff
@@ -1,3 +1,5 @@
+= Example RDoc
```
st0012, line 1: This needs to be escaped or it'd link to the RDoc constant

```suggestion
= Example \RDoc
```

(I'm aware that this is not intuitive but changing it would be a huge breaking change to some users)

st0012 (approved):
This is great, thank you!

st0012:
I think a rebase is needed to pass the CI.

PR #1186, opened 2024-10-01
Clean up the "toggle source" link

nithinbekal:
When the method param sequence is too long, the "toggle source" link can overlap the method params. The text for this link is also quite verbose - "click to toggle source".

### What's changed?

- Changes the link text from "click to toggle source" to "Toggle source"
- Keeps the link always visible, rather than only showing up on hover
- Removes the zoom image which takes up more space
- Makes minor changes to the typography of the text to make it easier to read

### Screenshots

Before

slice_when-before

After

slice_when-after


st0012:
Thanks for the PR, but I already included a similar change in the overhaul on method source toggle in #1176

nithinbekal:
Thanks, Stan! I should have looked at the open PRs first. 😅 

That one's much more comprehensive than this PR. Love the improvements!

PR #1188, opened 2024-10-07
Use normal `font-weight` for links

sambostock:
A weight of `600` is nearly the same as heading weight, which is `700`, putting too much _weight_ on links, and makes many sidebar links require wrapping. As RDoc is rendering API documentation, many identifiers are links, but aren't being used in a context where they should stand out.

Before|After
-|-
image|image

The default weight works nicely, and links are already identified by color and underline on hover.


st0012 (approved):
I agree that increasing the font weight for links was unnecessary. Thanks for the PR 👍

PR #1189, opened 2024-10-10
Make methods and attributes linkable

st0012:
1. Use the same highlighting color for headers (align with the change in #1176)
2. Make methods and attributes linkable 

https://github.com/user-attachments/assets/178a5561-260e-49a9-8839-b2e566c299d9


nobu (approved):

st0012:
@nobu Thank you for the prompt review!

PR #1193, opened 2024-10-19
Add release.yml

st0012:
This is the same as the one in `ruby/irb`. 

Ideally we should have an org-level setup for this but it's [not implemented](https://github.com/orgs/community/discussions/7926) yet.


colby-swandale (approved):

simi (commented):
:+1:

PR #1200, opened 2024-11-08
Extract excerpt from RDoc::Markup::Document (raw pages) correctly

st0012:
Fixes https://bugs.ruby-lang.org/issues/20862

### DIFF
--- a/test/rdoc/test_rdoc_generator_darkfish.rb
+++ b/test/rdoc/test_rdoc_generator_darkfish.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative 'helper'
 
-class TestRDocGeneratorDarkfish < RDoc::TestCase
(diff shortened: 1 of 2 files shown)


colby-swandale (approved):

test/rdoc/test_rdoc_generator_darkfish.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative 'helper'
 
-class TestRDocGeneratorDarkfish < RDoc::TestCase
+class RDocGeneratorDarkfishTest < RDoc::TestCase
```
olleolleolle, line 4: This class name change is... unconventional (I looked at the neighboring files). Is it intentional?

olleolleolle: @st0012 Thanks for providing that context. Seems like a good thing!

PR #1206, opened 2024-11-12
Split list of class and instance methods in two

alexisbernard:
Looking for a method is easier because eyes don't have to skip dashes or double colon.

Here is a screenshot where you can try to look for the instance method `collect` in order to compare:

rdoc_methods_1_vs_2_lists


st0012 (approved):
Thank you!

Issue #1199, opened 2024-11-05
Label list and note list styling should be different


nevans:
In #1157
> * Enhance table and list styling for improved readability

First off, I love the attention that's been given to rdoc over the last several months.  It's looking pretty nice!  I'm very much looking forward to the next release. 🤩

The only new issue I can think of is this: label lists and note lists now render the exact same way.  Was that intentional?  IMO, it's useful to have the two different definition list styles, even though I'd love for one or both of them to be updated from what they were in v6.7.0.

Here's what they look like in v6.7.0:
![Screenshot 2024-11-05 at 12-33-04 rdoc 6.7.0 definition lists](https://github.com/user-attachments/assets/3d29493d-010a-45d0-b889-92d99d19b1fd)

And here's what they look like today at https://ruby.github.io/rdoc/ExampleRDoc_rdoc.html
![Screenshot 2024-11-05 at 12-34-17 rdoc new definition lists](https://github.com/user-attachments/assets/3ee25b02-d4e7-4a48-a8cf-ab2740a48104)

Obviously, we can apply our own stylesheets, but IMO they should be slightly different out-of-the-box.  The `float` style that was previously used only for "note" lists is the one that is now used for both.

_Personally_, between the two I like the default styling on "note" lists less than v6.7.0's "label" lists.  IMO, the left margin's unindentation after more than two lines of text is visually jarring.

nevans:
If I have time, I'll try to submit a PR with updated styles for both lists soon.  But what do you think?  Do you agree that they _should_ be styled differently?  Should they just be returned to their 6.7.0 styles, or should they be updated to something (subjectively) nicer?

drbrain:
I don't think they need to look different when presented as HTML.

These are two types of markup that allow you to choose the most convenient way to create an HTML [description list](https://html.spec.whatwg.org/multipage/grouping-content.html#the-dl-element) based on the keys.

nevans:
@drbrain I suppose that's fair.  I can see that it's useful to have alternative representations for the exact same thing, with the only distinction that one reads more nicely in the source comments.  And I _do_ appreciate that rdoc has prioritized nice looking human-readable source comments over syntax-heavy comments-metadata.  Also, I hadn't realized (until your comments, when I tried it out and looked at the source) that `ri` _already_ formatted them both the same (and always has?).

[...]

PR #1208, opened 2024-11-15
Use the original `label` description list style

nevans:
As a default for all description lists, the original "label" style is more readable.

This is slightly different from the original `label` dl though:
* slightly increase left margin for `dd` (to 1em)
* remove right margin on `dd`
* remove `dt` bottom margin and `dd` top margin, to reduce the gap between the term and its description to only the standard line-height gap

See the discussion at #1199.  Because of that discussion, this PR gives _both_ description lists the "label" style.  A different PR (#1209) brings back distinct styling for the "note" list.

### DIFF
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -469,14 +469,15 @@ main dl {
 }
 
 main dt {
-  margin-bottom: 0.5em;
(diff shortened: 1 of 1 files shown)


nevans:
I'll make a separate PR to bring back a distinct "note" style, so that can be judged separately from this.  🙂

st0012:
Included in #1209

PR #1209, opened 2024-11-15
Use distinct styles for note lists and label lists

### LINKED ISSUE #1199: Label list and note list styling should be different

nevans:
Together with #1208, this brings back the original "note" and "label" list styles.  It also improves on the original "note" list styling in two ways: 1) the second line and every line after it share the same indentation, and 2) commas are added between multiple `dt` elements.

* #1208
  This also removes `margin-bottom` from `dt`, which is necessary for the second indentation to match subsequent lines.
* #1226
  Necessary for adding separators between `dt`.
* Bring back the `float: left` style, for note lists only
* Set the `dt` `line-height` to match `main p`.
  Otherwise, the `dt` has a larger `line-height` than the `p` tag inside the `dd`, which causes the second line to have the same indentation as the first but different from the third.
* Add commas between note list terms.

This fixes #1199.

### DIFF
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -469,14 +469,15 @@ main dl {
 }
 
 main dt {
-  margin-bottom: 0.5em;
-  margin-right: 1em;
-  float: left;
   font-weight: bold;
 }
 
 main dd {
-  margin: 0 1em 1em 0.5em;
+  margin: 0 0 1em 1em;
+}
+
+main dd p:first-child {
+  margin-top: 0;
 }
 
 /* Headers within Main */
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -469,7 +469,13 @@ main dl {
 }
 
 main dt {
+  line-height: 1.5; /* matches `main p` */
   font-weight: bold;
+}
+
+main dl.note-list dt {
+  margin-right: 1em;
+  float: left;
 }
 
 main dd {
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -478,6 +478,15 @@ main dl.note-list dt {
   float: left;
 }
 
+main dl.note-list dt:has(+ dt) {
+  margin-right: 0.25em;
+}
+
+main dl.note-list dt:has(+ dt)::after {
+  content: ', ';
+  font-weight: normal;
+}
+
 main dd {
   margin: 0 0 1em 1em;
 }
--- a/lib/rdoc/markup/to_html.rb
+++ b/lib/rdoc/markup/to_html.rb
@@ -407,7 +407,7 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
       "<li>"
     when :LABEL, :NOTE then
       Array(list_item.label).map do |label|
-        "<dt>#{to_html label}\n"
+        "<dt>#{to_html label}</dt>\n"
       end.join << "<dd>"
     else
       raise RDoc::Error, "Invalid list type: #{list_type.inspect}"
--- a/test/rdoc/test_rdoc_markup_to_html.rb
+++ b/test/rdoc/test_rdoc_markup_to_html.rb
@@ -146,7 +146,7 @@ class TestRDocMarkupToHtml < RDoc::Markup::FormatterTestCase
   end
 
   def accept_list_item_start_label
-    assert_equal "<dl class=\"rdoc-list label-list\"><dt>cat\n<dd>", @to.res.join
+    assert_equal "<dl class=\"rdoc-list label-list\"><dt>cat</dt>\n<dd>", @to.res.join
   end
 
   def accept_list_item_start_lalpha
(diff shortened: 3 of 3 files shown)


nevans:
For what it's worth, here are a few times the styles for these have been updated:
* #770
  * which mentions that ruby/ruby@33d026034e got it backwards
* 18abd1c95 is the local version of ruby/ruby@33d026034e
  * This came almost immediately after ruby/ruby@009144e8ef, which changed `NEWS` to use the new styling (or tried to).
* dd3f81ba - converted note lists from using table cells to using definition lists.
  * I think that rdoc styled them both the same from this point until 18abd1c9 and #770.  But I think documentation still indicated that note lists used something more like original table style.
  * `hanna` uses tables for note lists.  Maybe that's where `hanna` inherited the table styling from?

Anyway, this does not mean we need to style them differently by default.  If we add the closing `</dt>` tag and continue to give them different classes, then I _can_ use my own local stylesheets.  But I _do_ think it means that I'm not the only one who appreciates having distinct styles for the two list types. 🙂 

If this PR _is_ accepted, I'm willing to create a PR for similar styling for `ri`, so they match.

nevans:
FYI: I had previously played with a `display: grid` version of this, which put every `dd` into the second column, giving a somewhat table-like presentation.  While that is nice for some description lists, I think it's less generically applicable.  Also, it would've needed extra styling to be responsive to smaller screens.

The four different styles I've considered:

The "label" style restored by #1208
```
First described term
  Description starts on a new line below the term and is indented.  Lorem ipsum dolor
  sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore
  et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco
  laboris nisi ut aliquip ex ea commodo consequat.
Second described term
Riker Ipsum
  We finished our first sensor sweep of the neutral zone.  Yesterday I did not know
  how to eat gagh. And blowing into maximum warp speed, you appeared for an instant
  to be in two places at once.  They were just sucked into space.  Mr. Worf, you do
  remember how to fire phasers?
```

The "note" style in this PR
```
First described term    Description starts on same line as its term.  Subsequent
  lines are indented by one em.  Lorem ipsum dolor sit amet, consectetur adipiscing
  elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad
  minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
  commodo consequat.
Second described term, Riker Ipsum    Indentation levels are independent of the
  length of the terms.  Every description indents only one level.  We finished our
  first sensor sweep of the neutral zone.  Yesterday I did not know how to eat gagh.
  And blowing into maximum warp speed, you appeared for an instant to be in two
  places at once.  They were just sucked into space.  Mr. Worf, you do remember how
  to fire phasers?
```

An alternative "table" style, e.g as used by `hanna`
```
First described term    Definition starts on same line as its term.  All descriptions
                        share the same indentation.  This is easy to achieve with
                        display: grid.  But it does not work well with wide terms or
                        nested lists.  Lorem ipsum dolor sit amet, consectetur
                        adipiscing elit, sed do eiusmod tempor incididunt ut labore  
                        et dolore magna aliqua.
Second described term
Riker Ipsum             I chose to style this so that the description text starts on
                        the same line as the final description term.  Every
                        description indents only one level.  We finished our first
                        sensor sweep of the neutral zone.  Yesterday I did not know
                        how to eat gagh.  And blowing into maximum warp speed, you
                        appeared for an instant to be in two places at once.  They
                        were just sucked into space.  Mr. Worf, you do remember how
                        to fire phasers?
```

Indent each `dd` based on its `dt` width, mimicing how rdoc requires the comments to be formatted:

```
First term    Definition starts on same line as its term.  Each description has its
              own indentation, based on its term(s).  This still doesn't work well
              with wide terms or nested lists.  And I personally find the different
              indentation of each description to be jarring.  Lorem ipsum dolor sit
              amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut
              labore et dolore magna aliqua.
Second described term
Riker Ipsum              Indented according to its widest term.  We finished our
                         first sensor sweep of the neutral zone.  Yesterday I did
                         not know how to eat gagh. And blowing into maximum warp
                         speed, you appeared for an instant to be in two places at
                         once.  They were just sucked into space.  Mr. Worf, you do
                         remember how to fire phasers?
```

flavorjones:
I just opened a similar issue at https://github.com/ruby/rdoc/issues/1227 and I wanted to say that this PR solves my problem exactly as I'd hoped. ❤️

st0012 (commented):
Sorry for the delay. I think the updated style looks great and thanks for fixing the tag closing issue as well 👍

st0012 (approved):
Oops I was meant to approve

PR #1210, opened 2024-11-16
Stop accepting `Document` objects as `CodeObject#comment`

tompng:
`CodeObject#comment` is normally String or Comment, but Marshal dump and load creates CodeObject with Document as a comment. So there are two types of CodeObject that shouldn't be mixed:
- ParsedCodeObject: comment is a String or a Comment
- MarshalLoadedCodeObject: comment is a Document

This is implicitly doubling the number of total classes in RDoc. It is mixing huge complexity and potential bugs to RDoc codebase. Some method only accepts ParsedCodeObject. Some method only accepts MarshalLoadedCodeObject. It is difficult to know.
```ruby
def add_extension_modules_single out, store, include
  # include should be RDoc::Markup::(MarshalLoaded)Include. Does not accept RDoc::Markup::(Parsed)Include
  ...
  out << RDoc::Markup::BlankLine.new
  out << include.comment
end
```

It looks like document is assigned to comment to represent parse-result-cached comment. Alternatively, `Comment#document=` can be used to represent it.

Ideally, we should avoid mixing String with Comment, but String it is too frequently used. I think it will be easy to remove mixing String after switching to PrismRuby Parser.

I believe we should do this refactor someday but maybe not before Ruby 3.4 because the release date is pretty close.


lib/rdoc/code_object/context/section.rb
```diff
@@ -61,19 +61,13 @@ def == other
...
-    when Array then
-      @comments.concat comment
+    if comment.is_a?(Array)
+      comment.each do |c|
+        @comments << extract_comment(c)
+      end
     else
-      raise TypeError, "unknown comment type: #{comment.inspect}"
+      comment = extract_comment(comment)
+      @comments << comment unless comment.empty?
```
st0012, line 70: Can it be further simplified to something like:

```rb
comments = Array(comment)
comments.each do |c|
  extracted_comment = extract_comment(comment)
  @comments << extracted_comment unless extracted_comment.empty?
end
```

lib/rdoc/code_object/context/section.rb
```diff
@@ -135,19 +123,8 @@ def hash # :nodoc:
...
-      @comments.map do |comment|
-        comment.file
-      end
-    when RDoc::Markup::Document then
-      @comment.parts.map do |document|
-        document.file
-      end
-    else
-      raise RDoc::Error, "BUG: unknown comment class #{@comments.class}"
+    @comments.map do |comment|
```
st0012, line 126: nit: `@comments.map(&:file)`

lib/rdoc/code_object/context/section.rb
```diff
@@ -214,19 +172,8 @@ def plain_html
...
-      @comments.delete_if do |my_comment|
-        my_comment.file == comment.file
-      end
-    when RDoc::Markup::Document then
-      @comments.parts.delete_if do |document|
-        document.file == comment.file.name
-      end
-    else
-      raise RDoc::Error, "BUG: unknown comment class #{@comments.class}"
+    @comments.delete_if do |my_comment|
```
st0012, line 175: Can we also rename these locals to something like `stored_comment` and `target_comment`?

```rb
@comments.delete_if do |stored_comment|
  stored_comment.file == target_comment.file
end
```

lib/rdoc/generator/darkfish.rb
```diff
@@ -780,20 +780,13 @@ def template_for file, page = true, klass = ERB
...
     when RDoc::Comment
-      content.text
-    when RDoc::Markup::Document
-      # This case is for page files that are not markdown nor rdoc
-      # We convert them to markdown for now as it's easier to extract the text
-      formatter = RDoc::Markup::ToMarkdown.new
-      formatter.start_accepting
-      formatter.accept_document(content)
-      formatter.end_accepting
+      comment.text
```
st0012, line 787: Thank you for simplifying this part too ❤️

st0012 (commented):
Since this is a potentially risky change and we don't have much time to test it in another release, I'll wait for Ruby 3.4's release before merging it.

PR #1157, opened 2024-08-19
Modernize RDoc Darkfish template CSS

st0012:
- Update color scheme with muted green tones and improved contrast
- Enhance readability by adjusting font sizes and weights
- Improve code block styling with a light gray background
- Refactor layout for better responsiveness and sidebar presentation
- Standardize link styles across the document
- Implement CSS variables for easier theme customization
- Adjust heading styles for better visual hierarchy
- Enhance table and list styling for improved readability
- Optimize search field and navigation toggle appearance
- Improve method detail and documentation section styling
- Reorganize css and overhaul the sidebar design
- Improve code block's syntax highlighting

## Result

https://github.com/user-attachments/assets/72980420-f674-43e6-b423-54d4f8f53d65

## Comparisions

For detail comparison, I recommend clicking into images.

Left: old; Right: new

### `RDoc`

rdoc-diff-3

rdoc-diff-4
rdoc-diff-2
rdoc-diff-1

### Ruby official docs

ruby-diff-2
ruby-diff-1

PR #1211, opened 2024-11-19
Added missing _sidebar_toggle.rhtml

ishe-ua:
Add missing file


olleolleolle (approved):
This file was added in #1157 - but not added to the gemspec for inclusion, and this PR fixes that.

https://github.com/ruby/rdoc/blob/master/lib/rdoc/generator/template/darkfish/_sidebar_toggle.rhtml

st0012 (approved):

st0012:
Thank you!

zzak:
Why not just use a glob?

st0012:
@zzak doing that in https://github.com/ruby/rdoc/pull/1212

PR #1212, opened 2024-11-22
Improve how gemspec's files are defined

st0012:
Currently, the gemspec's files are defined by hand, which is error-prone.

For example: https://github.com/ruby/rdoc/pull/1211

This commit uses `Dir.glob` where possible to reduce the risk of that happening again.

- Additional files added with this approach:

    ```
    # This should have been added by only captured by this commit
    lib/rdoc/parser/prism_ruby.rb

    # These are folders and can be included/ignored either way
    # gems using Dir.glob in gemspec usually have them included
    lib/rdoc/generator/template/darkfish
    lib/rdoc/generator/template/darkfish/css
    lib/rdoc/generator/template/darkfish/fonts
    lib/rdoc/generator/template/darkfish/images
    lib/rdoc/generator/template/darkfish/js
    lib/rdoc/generator/template/json_index
    lib/rdoc/generator/template/json_index/js
    ```

- Files that are ignored after this change:

    ```
    # They make no difference on documentation generation
    # Probably can be removed
    lib/rdoc/generator/template/darkfish/.document
    lib/rdoc/generator/template/json_index/.document
    ```


rdoc.gemspec
```diff
@@ -37,7 +37,7 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
   s.require_paths = ["lib"]
   # for ruby core repository. It was generated by
   # `git ls-files -z`.split("\x0").each {|f| puts "    #{f.dump}," unless f.start_with?(*%W[test/ spec/ features/ .]) }
-  s.files = [
+  NONE_LIB_FILES = [
```
olleolleolle, line 40: Non-`lib/` files, perhaps.

```suggestion
  NON_LIB_FILES = [
```

rdoc.gemspec
```diff
@@ -50,183 +50,12 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
...
-    "lib/rdoc/code_object/top_level.rb",
-    "lib/rdoc/version.rb",
     "man/ri.1",
   ]
-  # files from .gitignore
-  s.files << "lib/rdoc/rd/block_parser.rb" << "lib/rdoc/rd/inline_parser.rb" << "lib/rdoc/markdown.rb" << "lib/rdoc/markdown/literals.rb"
+  TEMPLATE_FILES = Dir.glob("lib/rdoc/generator/template/**/*")
+  LIB_FILES = Dir.glob("lib/**/*.{rb,kpeg,ry}")
+
+  s.files = (NONE_LIB_FILES + TEMPLATE_FILES + LIB_FILES).uniq
```
olleolleolle, line 58:
```suggestion
  s.files = (NON_LIB_FILES + TEMPLATE_FILES + LIB_FILES).uniq
```

olleolleolle (commented):
(Spelling.)

st0012:
@olleolleolle Thanks, updated 👍

rdoc.gemspec
```diff
@@ -37,7 +37,7 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
   s.require_paths = ["lib"]
   # for ruby core repository. It was generated by
   # `git ls-files -z`.split("\x0").each {|f| puts "    #{f.dump}," unless f.start_with?(*%W[test/ spec/ features/ .]) }
-  s.files = [
+  NON_LIB_FILES = [
```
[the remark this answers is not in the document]
st0012: They don't have to be. I'm just used to defining static lists in constants. I've updated them to be locals instead.

colby-swandale (approved):

colby-swandale:
@hsbt Does this Pull Request impact any processes used for ruby-src?

st0012:
@colby-swandale If you're talking about [`sync_default_gems`](https://github.com/ruby/ruby/blob/14ce2443db2c97f870b6da74b613e05cab42042c/tool/sync_default_gems.rb#L153-L180), it won't be affected because it doesn't rely on the files gemspec defines.

hsbt:
I don't like `Dir.glob` in gemspec because it caused accidentally publishing debug, garbage code. 

But we can use trusted publisher now, it provides clean build and publish environments for us. I'm okay to this change.

colby-swandale:
:shipit:

PR #1213, opened 2024-11-23
Webri - a new console app for accessing Ruby online HTML docs [expermental]

BurdetteLamar:
This is _seriously_ experimental, and is best seen as a proof-of-concept.

I have not liked `ri`, so have worked up this app to get at the online docs.

### DIFF
--- a/lib/rdoc/web_ri.rb
+++ b/lib/rdoc/web_ri.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require 'rbconfig'
 require 'find'
+require 'cgi'
 
 require_relative '../rdoc'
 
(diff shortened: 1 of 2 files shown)


lib/rdoc/web_ri.rb
```diff
@@ -0,0 +1,170 @@
+# frozen_string_literal: true
+require 'open-uri'
+require 'nokogiri'
```
st0012, line 3: I don't think we can assume users would have these gems installed?

lib/rdoc/web_ri.rb
```diff
@@ -0,0 +1,170 @@
...
+# A class to display Ruby HTML documentation.
+class RDoc::WebRI
+
+  # Where the documentation lives.
+  ReleasesUrl = 'https://docs.ruby-lang.org/en/'
+
+
+  def initialize(target_name, options)
+    release_name = get_release_name(options[:release])
+    entries = get_entries(release_name)
```
st0012, line 17: So we'd have sent 2 requests by this point just to get the release and entries list. Considering both of them are rarely changed, I think we may introduce something like [rdoc-data](https://github.com/ruby/rdoc-data) used to do that packages all these information into a gem. And a tool like this can simply use it to generate the target url without making additional requests.

simi:
> I have not liked ri, so have worked up this app to get at the online docs.

Can you explain on this a little? What's problem with `ri`? Why online docs? Can't we show simliar page but locally using `ri --server` and opening related page (for `Array#join` - http://localhost:8214/ruby/Array.html#method-i-join)?

lib/rdoc/web_ri.rb
BurdetteLamar: On my machine(!), making the two requests is much faster than reading the static RI files.

So a new gem would include the data that this app gathers from the two requests?  And thus not need nokogiri?  Interesting.  But would not have master (though I'm not sure that's a drawback).

BurdetteLamar:
> > I have not liked ri, so have worked up this app to get at the online docs.
> 
> Can you explain on this a little? What's problem with `ri`? Why online docs? Can't we show simliar page but locally using `ri --server` and opening related page (for `Array#join` - http://localhost:8214/ruby/Array.html#method-i-join)?

RI output is often poorly formatted (compared to the corresponding HTML).  I'm aware of the HTML online docs, but not of whatever comes from `ri --server`.  If that's available, so much the better!

BurdetteLamar:
I'll look into building a gem that would do this work without the user's needing nokogiri.

lib/rdoc/web_ri.rb
BurdetteLamar: Agreed (for `nokogiri`);  I think `open-uri` is installed with Ruby.  But, better to avoid both.

simi:
@BurdetteLamar I think parsing of HTML will be too brittle, since any change in rdoc template will potentially break the parsing. Could be `ri` used in the background and once final target is located like method or class, run `ri --server` and open given page? I'm still not sure about the user experience, since the `ri --server` will be kept running around. :thinking:

lib/rdoc/web_ri.rb
BurdetteLamar: The more think about this idea (the gem, already containing the data for each release), the better I like it.  Will explore.

BurdetteLamar:
> @BurdetteLamar I think parsing of HTML will be too brittle, since any change in rdoc template will potentially break the parsing. Could be `ri` used in the background and once final target is located like method or class, run `ri --server` and open given page? I'm still not sure about the user experience, since the `ri --server` will be kept running around. 🤔

I think @st0012's suggestion (about a gem) avoids the problem altogether.

BurdetteLamar:
No longer interested in this (infeasible).

PR #1217, opened 2024-11-28
fix: C variables should never show up in Ancestors tree

### LINKED ISSUE #1205: In v6.8.1, C extension objects "Ancestors" list shows C symbol name instead of class name
Hi! I love the [ancestors list](https://github.com/ruby/rdoc/pull/1183), but I found an edge case with how it handles C extensions.

Take a look at the sidebar at https://nokogiri.org/rdoc/Nokogiri/HTML4/Document.html which was generated from 50dda134:

image

The actual ancestors list is:

```
Nokogiri::XML::Document
+- Nokogiri::XML::Node
   +- Object
      +- BasicObject
```

but instead it's showing `cNokogiriXmlNode` which is the C symbol that references the `Nokogiri::XML::Node` class (from `ext/nokogiri/xml_node.c`).

I haven't had time to dig into this yet, but wanted to raise an issue in case someone has the time to look.

flavorjones:
If a `NormalClass`'s superclass is a C variable ("enclosure"), then update the superclass to point to the RDoc::NormalClass.

This is done in a single pass after all files have been parsed.

Fixes #1205.

Before:

image

After:

image

### DIFF
--- a/lib/rdoc/store.rb
+++ b/lib/rdoc/store.rb
@@ -198,6 +198,18 @@ class RDoc::Store
   end
 
   ##
+  # Make sure any references to C variable names are resolved to the corresponding class.
+  #
+
+  def resolve_c_superclasses
+    @classes_hash.each_value do |klass|
+      if klass.superclass.is_a?(String) && (candidate = find_c_enclosure(klass.superclass))
+        klass.superclass = candidate
+      end
+    end
+  end
+
+  ##
   # Sets the parser of +absolute_name+, unless it from a source code file.
 
   def update_parser_of_file(absolute_name, parser)
--- a/test/rdoc/test_rdoc_store.rb
+++ b/test/rdoc/test_rdoc_store.rb
@@ -281,6 +281,26 @@ class TestRDocStore < XrefTestCase
     assert_nil @s.find_c_enclosure('cObject')
   end
 
+  def test_resolve_c_superclasses
+    # first parse a child that references an unknown parent
+    c_file1 = @s.add_file 'ext1.c'
+    c_file1.add_class RDoc::NormalClass, 'Child', 'cExternParent'
+
+    # then parse the parent and register the C variable name as a C enclosure
+    c_file2 = @s.add_file 'ext2.c'
+    parent = c_file2.add_class RDoc::NormalClass, 'Parent', 'rb_cObject'
+
+    @s.add_c_enclosure('cExternParent', parent)
+
+    # at this point, the child's superclass is still the name of the C variable
+    assert_equal("cExternParent", @s.classes_hash['Child'].superclass)
+
+    @s.resolve_c_superclasses
+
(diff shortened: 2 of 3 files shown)


lib/rdoc/store.rb
```diff
@@ -197,6 +197,19 @@ def add_file absolute_name, relative_name: absolute_name, parser: nil
...
   end
 
+  ##
+  # Make sure any references to C variable names are resolved to the corresponding class.
+  #
+
+  def resolve_c_references
+    # superclass references
+    @classes_hash.each do |_, klass|
+      if String === klass.superclass && (candidate = find_c_enclosure(klass.superclass))
```
st0012, line 207: nit:

```suggestion
    @classes_hash.each_value do |klass|
      if klass.superclass.is_a?(String) && (candidate = find_c_enclosure(klass.superclass))
```

lib/rdoc/store.rb
```diff
@@ -197,6 +197,19 @@ def add_file absolute_name, relative_name: absolute_name, parser: nil
     top_level
   end
 
+  ##
+  # Make sure any references to C variable names are resolved to the corresponding class.
+  #
+
+  def resolve_c_references
```
st0012, line 204: Will we need to resolve different types of references in the future? If we don't, `resolve_c_superclasses` may be better.

test/rdoc/test_rdoc_store.rb
```diff
@@ -281,6 +281,26 @@ def test_find_c_enclosure_from_cache_legacy
...
+
+    @s.add_c_enclosure('cExternParent', parent)
+
+    # at this point, the child's superclass is still the name of the C variable
+    assert_equal("cExternParent", @s.classes_hash['Child'].superclass)
+
+    @s.resolve_c_references
+
+    # now the ancestor tree correctly references the NormalClass objects
+    assert_equal(parent, @s.classes_hash['Child'].superclass)
```
st0012, line 301: (`#superclass` can return either string or `RDoc::NormalClass` really hurts my brain 🙈)

lib/rdoc/store.rb
flavorjones: 👍

lib/rdoc/store.rb
flavorjones: I'm not sure what else will end up with C variable names instead of classes, so I named it to be open-ended. But I'm happy to give it a narrower name (we can always rename it if we find we need to resolve more references later).

test/rdoc/test_rdoc_store.rb
flavorjones: 😬

st0012 (approved):
Thanks for the fix!

test/rdoc/test_rdoc_store.rb
flavorjones: And of course the bug in #1221 is related to this confusion. 🤦

PR #1218, opened 2024-11-29
Deprecate `main` and `title` directives

st0012:
Both main page and default title of a documentation should be global and immutable settings for the target project. Allowing users to override them in the middle of the documentation makes RDoc error-prone (see #426) and doesn't encourage the right behaviour.

We should deprecate these directives and remove them in the 7.0 release.

### DIFF
--- a/lib/rdoc/markup/pre_process.rb
+++ b/lib/rdoc/markup/pre_process.rb
@@ -223,6 +223,13 @@ class RDoc::Markup::PreProcess
     when 'title' then
       @options.default_title = param if @options.respond_to? :default_title=
 
+      warn <<~MSG
+        The :title: directive is deprecated and will be removed in RDoc 7.
+
+        Use `--title=#{param}` to specify the title displayed.
+        If you use `RDoc::Task`, use `rdoc.title = "#{param}"` instead.
+      MSG
+
       blankline
     when 'yield', 'yields' then
       return blankline unless code_object
--- a/lib/rdoc/markup/pre_process.rb
+++ b/lib/rdoc/markup/pre_process.rb
@@ -190,8 +190,10 @@ class RDoc::Markup::PreProcess
       warn <<~MSG
         The :main: directive is deprecated and will be removed in RDoc 7.
 
-        Use `--main=#{param}` to specify the initial page displayed.
-        If you use `RDoc::Task`, use `rdoc.main = "#{param}"` instead.
+        You can use these options to specify the initial page displayed instead:
+        - `--main=#{param}` via the command line
+        - `rdoc.main = "#{param}"` if you use `RDoc::Task`
+        - `main_page: #{param}` in your `.rdoc_options` file
       MSG
 
       blankline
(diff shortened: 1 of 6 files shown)


test/rdoc/test_rdoc_markup_pre_process.rb
```diff
@@ -85,18 +85,26 @@ def test_include_file_in_other_directory
 
   def test_handle
     text = "# :main: M\n"
-    out = @pp.handle text
+    output = nil
+    _, err =capture_output do
```
nobu, line 89: A nitpick:

```suggestion
    _, err = capture_output do
```

st0012: Thanks 👍

lib/rdoc/rdoc.rb
```diff
@@ -407,6 +407,7 @@ def parse_files files
 
     return [] if file_list.empty?
 
+    # This workaround can be removed after the :main: directive is removed
```
st0012, line 410:
```suggestion
    # This workaround can be removed after the :main: and :title: directives are removed
```

lib/rdoc/markup/pre_process.rb
```diff
@@ -187,6 +187,12 @@ def handle_directive prefix, directive, param, code_object = nil,
       include_file filename, prefix, encoding
     when 'main' then
       @options.main_page = param if @options.respond_to? :main_page
+      warn <<~MSG
+        The :main: directive is deprecated and will be removed in RDoc 7.
+
+        Use `--main=#{param}` to specify the initial page displayed.
```
colby-swandale, line 193: I think it's also worth mentioning `main_page` for the `.rdoc_options` file as well.

st0012:
> Does the :title: directive allow a person to customise the title element in Darkish?

Yes it's the same as the `title` option you can use in flags, `.rdoc_options`, and the rake task. [These](https://github.com/search?q=repo%3Aruby%2Frdoc+%40title+language%3AHTML%2BERB&type=code&l=HTML%2BERB) are their usages in the darkfish.

colby-swandale:
It seems (please correct me if I'm wrong) that this directive allows you to customise the title for each document generated from RDoc, which could be pretty useful for SEO. I remember that we just merged something recently that now lets us customise keyboards and meta descriptions. Would the `:title` directive fall under the same sort of functionality?

st0012:
Before https://github.com/ruby/rdoc/pull/1091, `:title:` is only used in the ToC page and index page, not each document. So it's not designed for per-page usage from the beginning yet it allows individual pages to change the global value, which is why I think this directive is not a well-designed feature.

Even after https://github.com/ruby/rdoc/pull/1091 started using it for SEO tags, the `option.title` value should still be considered a global value, not a way to customize each page's title by keep mutating it.

For SEO customization in the future, I think adding new `:meta:` or `:meta-[keywords|description]` directives which's value are attached to individual documents will be better.

colby-swandale (approved):

PR #1219, opened 2024-11-30
Sort `MethodAttr` so that names starting with symbols are before names starting with alpha ASCII

### LINKED ISSUE #1204: Method names sorted by ASCII code causes `~` and `|` to be placed at the bottom
Since the ASCII codes for `~` and `|` are greater than the alpha character codes, they are placed at the bottom of the method list compared to other operators such as `!`, `+`, or `%`. This is a bit odd as these two characters end up not grouped with the other operators.

For example, for [Integer documentation](https://docs.ruby-lang.org/en/master/Integer.html), we can see that most of the operator methods are at the top:

Screenshot 2024-11-11 at 11 18 40 AM

But there are two operators at the bottom:

Screenshot 2024-11-11 at 11 19 08 AM

flavorjones:
There are three distinct ranges of symbols in ASCII:

- the range below "A", 0..64 in decimal
- the range between "Z" and "a", 91..96 in decimal
- the range above "z", 123..127 in decimal

With this change, any method starting with a character in these "symbol ranges" will be sorted before a method starting with an alpha ASCII character. The remaining methods, all starting with alpha or 8-bit characters, will be sorted against each other exactly as before.

Specifically this addresses the issue from #1204 which is that `#[]` and `#^` were previously sorted _after_ the alpha methods. These methods will now be sorted before alpha methods.

Fixes #1204

### DIFF
--- a/lib/rdoc/code_object/method_attr.rb
+++ b/lib/rdoc/code_object/method_attr.rb
@@ -114,8 +114,8 @@ class RDoc::MethodAttr < RDoc::CodeObject
     return unless other.respond_to?(:singleton) &&
                   other.respond_to?(:name)
 
-    [     @singleton ? 0 : 1,       name] <=>
-    [other.singleton ? 0 : 1, other.name]
+    [@singleton      ? 0 : 1, name_codepoint_range,       name] <=>
+    [other.singleton ? 0 : 1, other.name_codepoint_range, other.name]
   end
 
   def == other # :nodoc:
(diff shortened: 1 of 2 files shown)


lib/rdoc/code_object/method_attr.rb
```diff
@@ -415,4 +415,16 @@ def to_s # :nodoc:
     end
   end
 
+  def name_codepoint_range # :nodoc:
+    case name.codepoints[0]
```
st0012, line 419: TIL `codepoints`

st0012 (approved):
Can confirm it fixes the issue. Thanks 👍

lib/rdoc/code_object/method_attr.rb
nobu: `case name.ord`

flavorjones: Thank you. See https://github.com/ruby/rdoc/pull/1220

PR #1225, opened 2024-12-05
Workaround JRuby's jar-dependencies error and test failures on CI

st0012:
https://github.com/mkristian/jar-dependencies/issues/86


test/rdoc/test_rdoc_generator_json_index.rb
```diff
@@ -237,6 +237,16 @@ def test_generate_gzipped
   end
 
   def test_generate_utf_8
```
st0012, line 239: @headius do you know why this test starts failing on JRuby?

PR #1226, opened 2024-12-05
Add closing tags for description list terms

nevans:
Without the closing tags, the `dt` elements contain whitespace after the text.  This normally isn't a big deal, but does mess some things up, e.g: using `::after` with `content: ", "` in stylesheets.


nevans:
n.b: this was extracted from #1209.  It allows users with custom styles to use `::after` with `content: ", "`, even if that's never added to the default styles.

st0012:
Included in #1209

PR #1228, opened 2024-12-09
Stop running CI against JRuby and some CI config cleanup

st0012:
Due to https://github.com/mkristian/jar-dependencies/issues/86, the JRuby CI hasn't been running for a while.

```
bundler: failed to load command: rake (/Users/runner/work/rdoc/rdoc/vendor/bundle/jruby/3.1.0/bin/rake)
Gem::LoadError: You have already activated jar-dependencies 0.5.0, but your Gemfile requires jar-dependencies 0.4.1. Since jar-dependencies is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports jar-dependencies as a default gem.
  check_for_activated_spec! at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/runtime.rb:30[8](https://github.com/ruby/rdoc/actions/runs/12235162661/job/34125882061#step:6:9)
                      setup at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/runtime.rb:25
                       each at org/jruby/RubyArray.java:2009
                       each at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/spec_set.rb:155
                        map at org/jruby/RubyEnumerable.java:818
                      setup at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/runtime.rb:24
                      setup at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler.rb:161
                     <main> at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/setup.rb:20
                 with_level at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/ui/shell.rb:136
                    silence at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/ui/shell.rb:88
                     <main> at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/setup.rb:20
                    require at org/jruby/RubyKernel.java:1187
           require_relative at org/jruby/RubyKernel.java:1216
                kernel_load at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/cli/exec.rb:56
                        run at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/cli/exec.rb:23
                       exec at /Users/runner/.rubies/jruby-head/lib/ruby/stdlib/bundler/cli.rb:486
[...]

PR #1229, opened 2024-12-10
Reenable JRuby

headius:
See https://github.com/ruby/rdoc/pull/1228

### DIFF
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -634,7 +634,7 @@ class RDoc::ClassModule < RDoc::Context
   # Path to this class or module for use with HTML generator output.
 
   def path
-    http_url @store.rdoc.generator.class_dir
+    http_url
   end
 
   ##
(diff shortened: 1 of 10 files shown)


headius:
This will need to be re-run after "jruby-head" builds get published overnight tonight or tomorrow.

headius:
I did not exclude Windows for now because I don't know why it wouldn't pass. If it fails again I'll look into it or exclude it again.

headius:
The JRuby 9.4.9.0 jobs are failing due to the version conflict. The head builds should be ok, but we may want to just remove the non-head jobs until I can figure out a workaround for the jar-dependencies conflict issue (or until JRuby 9.4.10.0 ships with a fix for it).

headius:
Marked as draft until we can confirm it with patched JRuby builds.

headius:
Well, the jruby-head builds have started to pass, but only on macos so far. Continuing to investigate remaining issues.

headius:
Oops, I can reopen myself....

PR #1232, opened 2024-12-12
[WIP] Centered styling rdoc

sunblaze:
Makes the desktop view less [framey](https://en.wikipedia.org/wiki/Frame_(World_Wide_Web))
Screenshot 2024-12-12 at 5 05 36 PM

Also if left nav is longer you can still scroll down.
Screenshot 2024-12-12 at 5 05 50 PM

Main caveats with this design is there's no easy way to get back to the top of the nav without scrolling to the top of the document. But I think it's a better more modern design on desktop.

Mobile view uses hamburger menu to toggle between nav and main content.
Screenshot 2024-12-12 at 5 09 08 PM

Nav
Screenshot 2024-12-12 at 5 09 15 PM

### DIFF
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -240,7 +240,7 @@ nav a:hover {
   top: 1rem;
   right: 1rem;
   cursor: pointer;
-  color: var(--highlight-color);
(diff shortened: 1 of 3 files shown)


sunblaze:
https://github.com/ruby/rdoc/pull/1235 tries to trim down the amount of changes. But it still needs changes to make the mobile view usable again.

PR #1238, opened 2024-12-13
Auto-hide navigation on link click

sunblaze:
Hide navigation sidebar when clicking anchor links on mobile devices. Previously, anchor links would change the page but the navigation sidebar would block the view.

### DIFF
--- a/lib/rdoc/generator/template/darkfish/js/darkfish.js
+++ b/lib/rdoc/generator/template/darkfish/js/darkfish.js
@@ -103,6 +103,12 @@ function hookSidebar() {
   if (isSmallViewport) {
     navigation.hidden = true;
     navigationToggle.ariaExpanded = false;
+    document.addEventListener('click', (e) => {
+      if (e.target.closest('#navigation a')) {
+        navigation.hidden = true;
+        navigationToggle.ariaExpanded = false;
+      }
+    });
   }
 }
 


lib/rdoc/generator/template/darkfish/js/darkfish.js
```diff
@@ -106,9 +106,22 @@ function hookSidebar() {
   }
 }
 
+function hookAnchorLinks() {
+  document.addEventListener('click', (e) => {
```
st0012, line 110: While this works well, is it possible to only add this listener to links under `#navigation`? I'm not familiar with JS but adding a global listener to click doesn't feel like the best practice 🤔

sunblaze: I can add the `#navigation` checking to the selection easily if we want to avoid triggering if main content is selected.

I chose this approach to add the document event listener because some content can be dynamically added (I was thinking of the search). It lets the click events bubble up to the document level and then checks if the target qualifies, which is usually preferred when there's many possible targets vs. attaching to all anchor tags individually.

sunblaze: updated to `closest` matcher to also check for `#navigation` ancestor

lib/rdoc/generator/template/darkfish/js/darkfish.js
```diff
@@ -106,9 +106,22 @@ function hookSidebar() {
...
+    const anchor = e.target.closest('#navigation a[href^="#"]');
```
st0012, line 111: We can't restrict the target to just anchors as some of the links (e.g. in search results) have relative path:
Screenshot 2024-12-17 at 19 34 04

st0012, line 111:
```suggestion
    const anchor = e.target.closest('#navigation a');
```

st0012:
I think we can simplify the implementation to this:


```diff
--- a/lib/rdoc/generator/template/darkfish/js/darkfish.js
+++ b/lib/rdoc/generator/template/darkfish/js/darkfish.js
@@ -103,19 +103,18 @@ function hookSidebar() {
   if (isSmallViewport) {
     navigation.hidden = true;
     navigationToggle.ariaExpanded = false;
-  }
-}
 
-function hookAnchorLinks() {
-  document.addEventListener('click', (e) => {
-    const anchor = e.target.closest('#navigation a[href^="#"]');
-    if (anchor && window.matchMedia("(max-width: 1023px)").matches) {
-      const navigation = document.querySelector('#navigation');
-      const navigationToggle = document.querySelector('#navigation-toggle');
-      navigation.hidden = true;
-      navigationToggle.ariaExpanded = false;
-    }
-  });
+    // hide navigation on link click when on small viewport
+    // this saves the user from having to click the toggle button again to hide the navigation
+    document.addEventListener('click', (e) => {
+      if (e.target.closest('#navigation a[href]')) {
+        const navigation = document.querySelector('#navigation');
+        const navigationToggle = document.querySelector('#navigation-toggle');
+        navigation.hidden = true;
+        navigationToggle.ariaExpanded = false;
+      }
+    });
+  }
 }
 
 document.addEventListener('DOMContentLoaded', function() {
@@ -123,5 +122,4 @@ document.addEventListener('DOMContentLoaded', function() {
   hookSearch();
   hookFocus();
   hookSidebar();
-  hookAnchorLinks();
 });
 ```

Basically, we can register the event using the same small viewport check to make sure the condition is shared. WDYT?

File: lib/rdoc/generator/template/darkfish/js/darkfish.js (hunk 2 above)
Comment (sunblaze) [reply lib/rdoc/generator/template/darkfish/js/darkfish.js:111]: I think that would still cause a refresh of the page, making listening to those click events unnecessary (switching a page hides the navigation anyways on mobile by default)

**Comment by sunblaze** [comment]
Yes, your simplification should work too. But you can also get rid of these lines as well since they're already defined
```javascript
const navigation = document.querySelector('#navigation');
const navigationToggle = document.querySelector('#navigation-toggle');
```

st0012 (approved):
This is a great quality-of-life improvement. Thank you!

PR #1239, opened 2024-12-13
Add background and drop shadow to navigation toggle

sunblaze:
Small improvement to the navigation toggle when there's overlapping text or boxes in the background. Makes it more visible and the small drop shadow shows it's floating nature better.

Screenshot 2024-12-13 at 2 11 54 PM


st0012:
Hey sorry that I tried to reduce the space between the icon and the top/bottom borders (basically have square borders) and suggest that change here, but I haven't found a way.

PR #1240, opened 2024-12-14
Enable cross reference in code

nobu:
Some people like to mark up method names in MarkDown style block quotes, like this: ruby/ruby#12333.
Currently, no links are created in the code in the RDoc, but such words most likely refer to methods.
This PR makes a word a code cross-reference if the whole word can be resolved as a reference.

### DIFF
--- a/test/rdoc/test_rdoc_markup_to_html_crossref.rb
+++ b/test/rdoc/test_rdoc_markup_to_html_crossref.rb
@@ -16,6 +16,18 @@ class RDocMarkupToHtmlCrossrefTest < XrefTestCase
     result = @to.convert 'C1'
 
     assert_equal para("<a href=\"C1.html\"><code>C1</code></a>"), result
+
+    result = @to.convert '+C1+'
+    assert_equal para("<a href=\"C1.html\"><code>C1</code></a>"), result
+
+    result = @to.convert 'FOO'
+    assert_equal para("FOO"), result
+
+    result = @to.convert '+FOO+'
+    assert_equal para("<code>FOO</code>"), result
+
+    result = @to.convert '<tt># :stopdoc:</tt>:'
+    assert_equal para("<code># :stopdoc:</code>:"), result
   end
 
   def test_convert_CROSSREF_method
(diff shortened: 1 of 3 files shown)


st0012 (changes requested):
I have a few questions/concerns:

- Given the place this changes are made, I think it'd be applied to markdown's too?
- Will these additional linking attempts cause significant slowdown when generating docs for large projects, like `ruby/ruby`
- We need to make sure when a code block doesn't link, it's still rendered as code block.
    - Currently, it breaks most of the inline code blocks in `MarkupReference.html`
        Screenshot 2024-12-14 at 12 06 25

        They should be:
        
        Screenshot 2024-12-14 at 12 09 17

nobu:
> * Given the place this changes are made, I think it'd be applied to markdown's too?

Yes, as mentioned in the description, the motivation is to mark up in markdown, but not only for it.

> * Will these additional linking attempts cause significant slowdown when generating docs for large projects, like `ruby/ruby`

There was no significant difference, 27sec -> 27sec on my machine.
Probably, already `handle_regexp_CROSSREF` does similar linking attempts, and marking up as code suppresses it.

> * We need to make sure when a code block doesn't link, it's still rendered as code block.

Fixed.

test/rdoc/test_rdoc_markup_to_html_crossref.rb
```diff
@@ -15,6 +15,12 @@ def test_convert_CROSSREF
     result = @to.convert 'C1'
 
     assert_equal para("<a href=\"C1.html\"><code>C1</code></a>"), result
+
```
st0012, line 18: When encountering a non-linkable code block, like `+FOO+`, it should be rendered as `<code>FOO</code>`, which is the current behaviour.

So there should be an assertion like:

```rb
    result = @to.convert '+FOO+'
    assert_equal para("<code>FOO</code>"), result
```

but it's failing with the current implementation.

nobu: Fixed.

st0012 (approved):
Thanks for the feature!
In `ruby/ruby`, it doesn't affect existing documentation much as far as I can tell. But now it links to many methods in news 👍 

### `master`

Screenshot 2024-12-17 at 21 45 32


### This branch

Screenshot 2024-12-17 at 21 45 35

PR #1244, opened 2024-12-15
Add attribute :force to RDoc::RubygemsHook just like RDoc::RubyGemsHook

### LINKED ISSUE #1243: `gem rdoc gem_name` fails with `undefined method 'force=' for an instance of RDoc::RubygemsHook`
Found in Reline's CI failure https://github.com/ruby/reline/actions/runs/12330703999
Failed jobs are `irb [2.7..3.3 and head] ubuntu latest`

Reproduction:
```
$ docker run --rm -ti rubylang/rubyfarm
root@bacda24563ca:/# ruby -v
ruby 3.4.0dev (2024-12-15T13:36:38Z master 366fd9642f) +PRISM [x86_64-linux]
root@bacda24563ca:/# gem rdoc readline
ERROR:  While executing gem ... (NoMethodError)
    undefined method 'force=' for an instance of RDoc::RubygemsHook

      doc.force = options[:overwrite]
         ^^^^^^^^
	/opt/ruby/lib/ruby/3.4.0+1/rubygems/commands/rdoc_command.rb:80:in 'block in Gem::Commands::RdocCommand#execute'
	/opt/ruby/lib/ruby/3.4.0+1/rubygems/commands/rdoc_command.rb:77:in 'Array#each'
	/opt/ruby/lib/ruby/3.4.0+1/rubygems/commands/rdoc_command.rb:77:in 'Gem::Commands::RdocCommand#execute'
	/opt/ruby/lib/ruby/3.4.0+1/rubygems/command.rb:326:in 'Gem::Command#invoke_with_build_args'
	/opt/ruby/lib/ruby/3.4.0+1/rubygems/command_manager.rb:253:in 'Gem::CommandManager#invoke_command'
	/opt/ruby/lib/ruby/3.4.0+1/rubygems/command_manager.rb:194:in 'Gem::CommandManager#process_args'
	/opt/ruby/lib/ruby/3.4.0+1/rubygems/command_manager.rb:152:in 'Gem::CommandManager#run'
	/opt/ruby/lib/ruby/3.4.0+1/rubygems/gem_runner.rb:57:in 'Gem::GemRunner#run'
	/opt/ruby/bin/gem:12:in '<main>'
```

`RDoc::Ruby"G"emsHook` have method `force=` but `RDoc::Ruby"g"emsHook` used by rubygems does not have it.

tompng:
Rubygems creates an instance of RDoc::RubygemsHook, sets `doc.force = overwrite`, then calls `doc.generate` the document. RDoc::RubygemsHook needs attribute `:force` just like RDoc::RubyGemsHook.

Fixes #1243

### DIFF
--- a/lib/rdoc/rubygems_hook.rb
+++ b/lib/rdoc/rubygems_hook.rb
@@ -271,7 +271,7 @@ end
 module RDoc
   class RubygemsHook
 
-    attr_accessor :generate_rdoc, :generate_ri
+    attr_accessor :generate_rdoc, :generate_ri, :force
 
     def self.default_gem?
       !File.exist?(File.join(__dir__, "..", "rubygems_plugin.rb"))
@@ -281,6 +281,7 @@ module RDoc
       @spec = spec
       @generate_rdoc = generate_rdoc
       @generate_ri   = generate_ri
+      @force = false
     end
 
     def generate
@@ -288,7 +289,9 @@ module RDoc
       return unless self.class.default_gem?
 
       # Generate document for compatibility if this is a default gem.
-      RubyGemsHook.new(@spec, @generate_rdoc, @generate_ri).generate
+      hook = RubyGemsHook.new(@spec, @generate_rdoc, @generate_ri)
+      hook.force = @force
+      hook.generate
     end
 
     def remove
--- a/test/rdoc/test_rdoc_rubygems_hook.rb
+++ b/test/rdoc/test_rdoc_rubygems_hook.rb
@@ -200,6 +200,25 @@ class TestRDocRubyGemsHook < Test::Unit::TestCase
     assert_path_exist File.join(@a.doc_dir('ri'),   'cache.ri')
   end
 
+  def test_generate_rubygems_compatible
+    original_default_gem_method = RDoc::RubygemsHook.method(:default_gem?)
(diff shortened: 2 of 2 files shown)


lib/rdoc/rubygems_hook.rb
```diff
@@ -281,14 +281,17 @@ def initialize(spec, generate_rdoc = false, generate_ri = true)
       @spec = spec
       @generate_rdoc = generate_rdoc
       @generate_ri   = generate_ri
```
st0012, line 283: I was thinking that we can initialize a `RubyGemsHook` here already, like

```rb
    def initialize(spec, generate_rdoc = false, generate_ri = true)
      @rubygems_hook = RubyGemsHook.new(spec, generate_rdoc, generate_ri)
    end
```

And then delegate methods to `@rubygems_hook`, which will handle more cases 🤔

tompng: IMO, implementing minimal things to clarify what method/attribute are needed in the compatibility layer is not a bad idea. I think it is the intention of the current implementation of RubygemsHook. (I', not sure)
This way, the new RubyGemsHook can easily change its method as long as RubygemsHook's methods works.

st0012 (approved):

PR #1250, opened 2024-12-22
Display main page's Table of Contents when viewed from the index page

st0012:
When the main page document is viewed as a standalone page, Darkfish displays the Table of Contents. When viewed from the index page, it does not.

This page changes the behaviour to also display the Table of Contents when the main page document is viewed from the index page.

### Before

Screenshot 2024-12-22 at 16 30 45

### After

Screenshot 2024-12-22 at 16 31 14

### In `ruby/ruby`

Screenshot 2024-12-22 at 16 34 10

### DIFF
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -316,11 +316,14 @@ class RDoc::Generator::Darkfish
     asset_rel_prefix = rel_prefix + @asset_rel_path
 
     @title = @options.title
+    @main_page = @files.find { |f| f.full_name == @options.main_page }
 
     render_template template_file, out_file do |io|
       here = binding
       # suppress 1.9.3 warning
       here.local_variable_set(:asset_rel_prefix, asset_rel_prefix)
+      # some partials rely on the presence of current variable to render
+      here.local_variable_set(:current, @main_page) if @main_page
       here
     end
   rescue => e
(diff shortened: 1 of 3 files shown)


lib/rdoc/generator/darkfish.rb
```diff
@@ -316,11 +316,14 @@ def generate_index
...
 
     @title = @options.title
+    @main_page = @files.find { |f| f.full_name == @options.main_page }
 
     render_template template_file, out_file do |io|
       here = binding
       # suppress 1.9.3 warning
       here.local_variable_set(:asset_rel_prefix, asset_rel_prefix)
+      # some partials rely on the presence of current variable to render
+      here.local_variable_set(:current, @main_page) if @main_page
```
vinistock, line 326: Unrelated to this PR, but do we still need the line above that says it's suppressing a 1.9.3 warning?

test/rdoc/test_rdoc_generator_darkfish.rb
```diff
@@ -134,6 +134,88 @@ def test_generate
...
+
+    assert_hard_link 'css/rdoc.css'
+    assert_hard_link 'css/fonts.css'
+
+    assert_hard_link 'fonts/SourceCodePro-Bold.ttf'
+    assert_hard_link 'fonts/SourceCodePro-Regular.ttf'
+
+    index_html = File.binread('index.html')
+
+    # If there is no main page, the index page should not have a table of contents
```
vinistock, line 214: Why not though? Is there a reason why we can't or don't want to display the table of contents every time?

vinistock (approved):

test/rdoc/test_rdoc_generator_darkfish.rb
st0012: The lack of main page here means it'll literally only have one title `This is the API documentation for My awesome Ruby project.` and nothing else.

PR #1251, opened 2024-12-23
Introduce `apply_default_exclude` config to separate default and user excluding patterns

nobu:
For ruby/ruby#12386.

### DIFF
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -355,18 +355,24 @@ class RDoc::Options
   # +--[no-]embed-mixins+ (Default is +false+.)
   attr_accessor :embed_mixins
 
+  ##
+  # Exclude the default patterns as well if true.
+  attr_reader :apply_default_exclude
+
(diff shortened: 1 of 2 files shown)


lib/rdoc/options.rb
```diff
@@ -355,18 +355,24 @@ class RDoc::Options
   # +--[no-]embed-mixins+ (Default is +false+.)
   attr_accessor :embed_mixins
 
+  ##
+  # Exclude default pattersn to exclude.
+  attr_reader :default_exclude
```
st0012, line 360: Since the set of excluded patterns is called `@exclude` now, `@default_exclude` feels like would contain the value of `DEFAULT_EXCLUDE`. I think maybe `apply_default_exclude` expresses what this value does better?

nobu: Updated.

st0012 (approved):
This is great. Thank you 👍

Issue #1231, opened 2024-12-12
Markdown links with inline code are parsed incorrectly


st0012:
Given 2 links:

```md
[Link to Foo](https://example.com)

[Link to `Foo`](https://example.com)
```

RDoc can't handle the 2nd case:

<img width="163" alt="Screenshot 2024-12-12 at 16 18 46" src="https://github.com/user-attachments/assets/a3c8cf4e-9c43-465e-849b-f24754a4514d" />

amomchilov:
Once this is implemented, we can improve `doc/ruby/options.md` like so:

````patch
From 0a7a8f7aa5ab9a20107b8808f2cada9aa4d9dfbe Mon Sep 17 00:00:00 2001
From: Alexander Momchilov <alexander.momchilov@shopify.com>
Date: Thu, 12 Dec 2024 12:18:14 -0500
Subject: [PATCH] Highlight param names as code

---
 doc/ruby/options.md | 74 ++++++++++++++++++++++-----------------------
 1 file changed, 37 insertions(+), 37 deletions(-)

diff --git a/doc/ruby/options.md b/doc/ruby/options.md
index 6e0278e143..2062fdafd5 100644
--- a/doc/ruby/options.md
+++ b/doc/ruby/options.md
@@ -61,15 +61,15 @@ nil
 
 See also:
 
-- {Option -a}[rdoc-ref:ruby/options.md@a-3A+Split+Input+Lines+into+Fields]:
+- {Option `-a`}[rdoc-ref:ruby/options.md@a-3A+Split+Input+Lines+into+Fields]:
   Split input lines into fields.
-- {Option -F}[rdoc-ref:ruby/options.md@F-3A+Set+Input+Field+Separator]:
+- {Option `-F`}[rdoc-ref:ruby/options.md@F-3A+Set+Input+Field+Separator]:
   \Set input field separator.
-- {Option -l}[rdoc-ref:ruby/options.md@l-3A+Set+Output+Record+Separator-3B+Chop+Lines]:
+- {Option `-l`}[rdoc-ref:ruby/options.md@l-3A+Set+Output+Record+Separator-3B+Chop+Lines]:
   \Set output record separator; chop lines.
-- {Option -n}[rdoc-ref:ruby/options.md@n-3A+Run+Program+in+gets+Loop]:
+- {Option `-n`}[rdoc-ref:ruby/options.md@n-3A+Run+Program+in+gets+Loop]:
   Run program in `gets` loop.
[...]

PR #1252, opened 2024-12-23
WIP: Writting test for links with code issue

skatkov:
Trying to reproduce issue mentioned in #1231 via tests.


tompng:
Original issue is fixed in #1428. Tests are also added.

PR #1258, opened 2024-12-28
Simplify switch statement in `parse_statements`.

p8:
`:on_nl`, `:on_ignored_nl`, `:on_comment`, `:on_embdoc` are handled by a single `when` branch. In this branch there are multiple `if` branches handling `:on_nl` and `:on_ignored_nl` separately from `:on_comment` and `:on_embdoc`.

By having separate `when` branches for `:on_nl`/`:on_ignored_nl` and `:on_comment`/`:on_embdoc` we can remove the `if` statements.


lib/rdoc/parser/ruby.rb
```diff
@@ -1804,67 +1804,64 @@ def parse_statements(container, single = NORMAL, current_method = nil,
       non_comment_seen = true unless (:on_comment == tk[:kind] or :on_embdoc == tk[:kind])
 
       case tk[:kind]
-      when :on_nl, :on_ignored_nl, :on_comment, :on_embdoc then
-        if :on_nl == tk[:kind] or :on_ignored_nl == tk[:kind]
-          skip_tkspace
-          tk = get_tk
```
st0012, line 1810: I'm not sure if the before/after behaviours are actually the same to be honest. 

In this condition, the `tk` variable is reassigned so the later `if` statement is comparing the new `tk` instead of the current one. And this behaviour doesn't seem to be replicated in the updated version.

p8: That line would be followed by the following:
```ruby
unget_tk tk
```
They cancel each other out, so I removed them.
The next `tk` would get picked up in the surrounding `while` loop.

There is some [extra processing](https://github.com/ruby/rdoc/blob/cda767bfc793b5f9f905a1e563f587ceb8886162/lib/rdoc/parser/ruby.rb#L1956-L1970) done at the end of the `while` loop, but this seems to always get skipped as `try_parse_comment` is always false and `keep_comment` is `true` in [both cases](https://github.com/ruby/rdoc/blob/cda767bfc793b5f9f905a1e563f587ceb8886162/lib/rdoc/parser/ruby.rb#L1869).

p8: I think the following test case for `TestRDocParserRuby` reproduces this behaviour.
It is green on both branches.

```ruby
  def test_parse_regression
    util_parser <<-CLASS
class Foo
  # ignored comment

  ##
  # :method: ghost
  # This is a ghost method

  # This is a regular method
  def regular
  end
end
    CLASS

    tk = @parser.get_tk

    @parser.parse_class @top_level, RDoc::Parser::Ruby::NORMAL, tk, @comment

    foo = @top_level.classes.first
    assert_equal 'Foo', foo.full_name

    ghost = foo.method_list.first
    assert_equal 'Foo#ghost', ghost.full_name
    assert_equal 'This is a ghost method', ghost.comment.to_s

    regular = foo.method_list.last
    assert_equal 'Foo#regular', regular.full_name
    assert_equal 'This is a regular method', regular.comment.to_s
  end
```

p8: I've added a regression test just to make sure.

tompng:
Closing because `lib/rdoc/parser/ruby.rb` (renamed to ripper_ruby.rb) is no longer a maintained default parser for `.rb` files, and will be removed in the future.

PR #1259, opened 2024-12-28
Add autolink_excluded_words option to ignore cross-references

### LINKED ISSUE #1254: Allow projects to disable auto-linking on certain words
https://github.com/ruby/ruby/pull/12451 introduced the top-level `Ruby` constant, which means that every `Ruby` in https://docs.ruby-lang.org/en/master/index.html are now linked to it:

Screenshot 2024-12-25 at 12 00 39

Instead of escaping them one by one (there are more than a thousand occurrences in `ruby/ruby`), RDoc should allow projects to list a set of words that should not be auto-linked.

st0012:
This config will be handy when the project name is the same as a class or module name, which is often the case for most of the projects.

Closes #1254

### DIFF
--- a/lib/rdoc/markup/to_html_crossref.rb
+++ b/lib/rdoc/markup/to_html_crossref.rb
@@ -83,6 +83,8 @@ class RDoc::Markup::ToHtmlCrossref < RDoc::Markup::ToHtml
   def handle_regexp_CROSSREF(target)
     name = target.text
 
+    return name if @options.autolink_excluded_words&.include?(name)
+
     return name if name =~ /@[\w-]+\.[\w-]/ # labels that look like emails
 
     unless @hyperlink_all then
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -359,6 +359,10 @@ class RDoc::Options
   # Exclude the default patterns as well if true.
   attr_reader :apply_default_exclude
 
+  ##
+  # Words to be ignored in autolink cross-references
+  attr_accessor :autolink_excluded_words
+
   def initialize loaded_options = nil # :nodoc:
     init_ivars
     override loaded_options if loaded_options
@@ -370,6 +374,7 @@ class RDoc::Options
   ]
 
   def init_ivars # :nodoc:
+    @autolink_excluded_words = []
     @dry_run = false
     @embed_mixins = false
     @exclude = []
@@ -437,7 +442,9 @@ class RDoc::Options
     @title          = map['title']
     @visibility     = map['visibility']
     @webcvs         = map['webcvs']
-    @apply_default_exclude = map['apply_default_exclude']
+
+    @apply_default_exclude   = map['apply_default_exclude']
+    @autolink_excluded_words = map['autolink_excluded_words']
 
     @rdoc_include = sanitize_path map['rdoc_include']
     @static_path  = sanitize_path map['static_path']
@@ -471,6 +478,7 @@ class RDoc::Options
     @title          = map['title']          if map.has_key?('title')
     @visibility     = map['visibility']     if map.has_key?('visibility')
     @webcvs         = map['webcvs']         if map.has_key?('webcvs')
+    @autolink_excluded_words = map['autolink_excluded_words'] if map.has_key?('autolink_excluded_words')
     @apply_default_exclude = map['apply_default_exclude'] if map.has_key?('apply_default_exclude')
 
     @warn_missing_rdoc_ref = map['warn_missing_rdoc_ref'] if map.has_key?('warn_missing_rdoc_ref')
@@ -503,7 +511,8 @@ class RDoc::Options
       @title          == other.title          and
       @visibility     == other.visibility     and
       @webcvs         == other.webcvs         and
-      @apply_default_exclude == other.apply_default_exclude
+      @apply_default_exclude == other.apply_default_exclude and
+      @autolink_excluded_words == other.autolink_excluded_words
   end
 
   ##
@@ -985,6 +994,13 @@ Usage: #{opt.program_name} [options] [names...]
              "Use --encoding instead of --charset if",
              "available.") do |value|
         @charset = value
+      end
+
+      opt.separator nil
+
+      opt.on("--autolink-excluded-words=WORDS", Array,
+             "Words to be ignored in autolink cross-references") do |value|
+        @autolink_excluded_words.concat value
       end
 
       opt.separator nil
--- a/test/rdoc/test_rdoc_markup_to_html_crossref.rb
+++ b/test/rdoc/test_rdoc_markup_to_html_crossref.rb
@@ -20,6 +20,9 @@ class RDocMarkupToHtmlCrossrefTest < XrefTestCase
     result = @to.convert '+C1+'
     assert_equal para("<a href=\"C1.html\"><code>C1</code></a>"), result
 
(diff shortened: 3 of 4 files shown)


lib/rdoc/options.rb
```diff
@@ -989,6 +998,13 @@ def parse argv
 
       opt.separator nil
 
+      opt.on("--autolink-excluded-words=WORDS", Array,
+             "Words to be ignored in autolink cross-references") do |value|
+        @autolink_excluded_words = value
```
nobu, line 1003: To give this options multiple times:

```suggestion
        @autolink_excluded_words.concat value
```

lib/rdoc/markup/to_html_crossref.rb
```diff
@@ -151,6 +151,8 @@ def link name, text, code = true, rdoc_ref: false
       label = $'
     end
 
+    return name if name == text && @options.autolink_excluded_words&.include?(name)
```
kou, line 154: Does this work with `C1[rdoc-ref:C1]`?
(I think that this should be linked.)

nobu: Isn't the last assertion in `test_convert_CROSSREF_ignored_excluded_words` that case?

test/rdoc/test_rdoc_markup_to_html_crossref.rb
```diff
@@ -30,6 +33,21 @@ def test_convert_CROSSREF
...
 
+  def test_convert_CROSSREF_ignored_excluded_words
+    @options.autolink_excluded_words = ['C1']
+
+    result = @to.convert 'C1'
+
+    assert_equal para("C1"), result
+
+    result = @to.convert '+C1+'
+    assert_equal para("<code>C1</code>"), result
```
nobu, line 44: Wouldn't it be better that this is linked?
Simple word like "Ruby", "Set" and "Rails" may not be classes/modules, but code like `+Ruby+`, `+Set+` and `+Rails+` (or `` `Ruby` ``, `` `Set` `` and `` `Rails` `` in Markdown format) are very likely classes/modules, I think.

st0012: While I understand the argument, I think it makes the behaviour harder to understand for users. If both `C1` and `+C1+` are autolinked, I'd expect the exclusion to happen to both of them as well.

Also, some people, including me, often use markdown's code to highlight & emphasize certain keywords without expecting linking. And that causes examples like [this](https://docs.ruby-lang.org/en/master/contributing/building_ruby_md.html):

Screenshot 2024-12-29 at 11 48 19

So I'd expect this config to fix that behavior too.

lib/rdoc/options.rb
st0012: Updated 👍

lib/rdoc/markup/to_html_crossref.rb
st0012: > Does this work with C1[rdoc-ref:C1]?

Yes in that case RDoc would still link it. And it's covered by a test as well.

test/rdoc/test_rdoc_markup_to_html_crossref.rb
nobu: > While I understand the argument, I think it makes the behaviour harder to understand for users. If both `C1` and `+C1+` are autolinked, I'd expect the exclusion to happen to both of them as well.

These are different.
`C1` does not need to be the whole code,  it is auto-linked in a sentence `the C1 class`.
However `+C1+` must be the whole code, `+the C1 class+` is not linked.

I mean to limit the config to plain texts.

> Also, some people, including me, often use markdown's code to highlight & emphasize certain keywords without expecting linking. And that causes examples like [this](https://docs.ruby-lang.org/en/master/contributing/building_ruby_md.html):
> Screenshot 2024-12-29 at 11 48 19

The word "Ruby" is not marked as code in that example, is it?
These "Ruby" will not be linked in my proposal.

(and I think we should discourage use of markdown's code just to highlight & emphasize non-code)

> So I'd expect this config to fix that behavior too.

I'm not against the config itself.
But a new directive at that module feels better to me.

nobu:
```diff
diff --git i/lib/rdoc/markup/to_html_crossref.rb w/lib/rdoc/markup/to_html_crossref.rb
index 76714522..678e5ee1 100644
--- i/lib/rdoc/markup/to_html_crossref.rb
+++ w/lib/rdoc/markup/to_html_crossref.rb
@@ -92,6 +92,8 @@ class RDoc::Markup::ToHtmlCrossref < RDoc::Markup::ToHtml
       return name if name =~ /\A[a-z]*\z/
     end
 
+    return name if @options.autolink_excluded_words&.include?(name)
+
     cross_reference name, rdoc_ref: false
   end
 
@@ -151,8 +153,6 @@ class RDoc::Markup::ToHtmlCrossref < RDoc::Markup::ToHtml
       label = $'
     end
 
-    return name if name == text && @options.autolink_excluded_words&.include?(name)
-
     ref = @cross_reference.resolve name, text if name
 
     case ref
diff --git i/test/rdoc/test_rdoc_markup_to_html_crossref.rb w/test/rdoc/test_rdoc_markup_to_html_crossref.rb
index cf95ebf3..e50002e8 100644
--- i/test/rdoc/test_rdoc_markup_to_html_crossref.rb
+++ w/test/rdoc/test_rdoc_markup_to_html_crossref.rb
@@ -41,7 +41,7 @@ class RDocMarkupToHtmlCrossrefTest < XrefTestCase
     assert_equal para("C1"), result
 
     result = @to.convert '+C1+'
-    assert_equal para("<code>C1</code>"), result
+    assert_equal para("<a href=\"C1.html\"><code>C1</code></a>"), result
 
     # Explicit linking with rdoc-ref is not ignored
     result = @to.convert 'Constant[rdoc-ref:C1]'
```

st0012: So in the case of `+C1+`, it's an explicit cross-ref linking syntax now. Can you update this [document](https://github.com/ruby/rdoc/blob/master/doc/rdoc/markup_reference.rb) to mention it?

st0012: I've updated the implementation to follow this expectation.

test/rdoc/test_rdoc_markup_to_html_crossref.rb
```diff
@@ -30,6 +33,21 @@ def test_convert_CROSSREF
     assert_equal para("<code># :stopdoc:</code>:"), result
   end
 
+  def test_convert_CROSSREF_ignored_excluded_words
+    @options.autolink_excluded_words = ['C1']
+
+    result = @to.convert 'C1'
+
```
kou, line 40:
```suggestion
```

kou (approved):
+1

lib/rdoc/markup/to_html_crossref.rb
kou: (`name == text` was needless. If `name == text` exists, `Constant[rdoc-ref:C1]` works but `C1[rdoc-ref:C1]` doesn't work. Because both of `name` and `text` are `C1` with `C1[rdoc-ref:C1]`.)

nobu (approved):

nobu:
Another idea: #1264

PR #1261, opened 2024-12-29
Handle ghost methods followed by undocumented methods

p8:
Docs for ghost methods would get thrown away if the following method
does not have a comment. By parsing comments on linebreaks we avoid this
problem.

Builds on: #1258

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -1806,6 +1806,8 @@ class RDoc::Parser::Ruby < RDoc::Parser
       case tk[:kind]
       when :on_nl, :on_ignored_nl then
         skip_tkspace
+        non_comment_seen = parse_comment container, tk, comment unless
+          comment.empty?
 
         keep_comment = true
         container.current_line_visibility = nil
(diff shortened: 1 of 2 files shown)


tompng:
This approach changes the way to write metaprogramming comment.
Example in https://ruby.github.io/rdoc/RDoc/Parser/Ruby.html#class-RDoc::Parser::Ruby-label-Metaprogrammed+Methods indicates that newline between metaprogramming comment and metaprogramming code is allowed, and this pull request changes it.

```ruby
# This is OK
class ActiveStorage::Blob < ActiveStorage::Record
  ##
  # :method:
  #
  # Returns the associated ActiveStorage::Attachment instances.
  has_many :attachments, autosave: false
end

# OK in master, but document is not generated with this pull request
class ActiveStorage::Blob < ActiveStorage::Record
  ##
  # :method:
  #
  # Newline between metaprogramming comment and metaprogramming code

  has_many :attachments, autosave: false
end
```

The problem is in the RDoc notation ambiguity. ghost_method+undocumented_method and metaprogramming_comment+metaprogramming_code are ambiguous.
I found a workaround in rails/rails by inserting `##`
https://github.com/rails/rails/blob/be9aa73dd72f1097be5d45a58d7912447a266bd1/activemodel/lib/active_model/attributes.rb#L90
```ruby
##
# :method: ghost

##
private # this is not a metaprogramming code
```

Disallow newline is one possible approach but is a big change that need to be carefully considered.
Another possible approach is to disallow some specific pattern like `def [different_name]`, `private`, `public` etc.
Or don't change and just document the workaround.

p8:
Thanks for the explanation @tompng !
I'll try to add a test to document this behaviour.

tompng:
Closing because `lib/rdoc/parser/ruby.rb` (renamed to ripper_ruby.rb) is no longer a maintained default parser for `.rb` files, and will be removed in the future.

PR #1265, opened 2025-01-01
[DOC] Ignore racc-generated files

nobu:
Not only `.ry` sources, generated `.rb` files seem duplicate and unnecessary.


.rdoc_options
```diff
@@ -1,8 +1,4 @@
 exclude:
```
st0012, line 1: Given that it now has no values, maybe the key can be removed too?

nobu: Updated!

st0012 (approved):
Nice! Thx

PR #1274, opened 2025-01-08
Finalize RDoc::Options before calling RDoc::RDoc#parse_files

rhenium:
Commit 6cf6e1647b97, which went to v6.5.0, changed `RDoc::Options#parse` to not call `#finish` in it. While the commit adjusted other call sites, it missed `lib/rdoc/rubygems_hook.rb`.

`RDoc::Options#finish` prepares the include paths for `:include:` directives. This has to be done before starting to parse sources.

I think this should fix https://github.com/ruby/net-http/issues/193 + https://github.com/ruby/net-http/pull/194.


st0012 (approved):
Thanks for the fix!

PR #1275, opened 2025-01-12
Enable `warn_missing_rdoc_ref` by default

st0012:
This feature has been tested in RDoc and IRB, as well as my local Ruby project for a while. It's been finding and now preventing dead rdoc-ref links effectively. So I think it's a great feature to be enabled by default.

### DIFF
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -410,7 +410,7 @@ class RDoc::Options
     @update_output_dir = true
     @verbosity = 1
     @visibility = :protected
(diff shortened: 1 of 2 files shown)


st0012:
@nobu instead of just enabling it in `ruby/ruby` with https://github.com/ruby/ruby/pull/12510, and then let different RDoc users slowly picking the configuration up, I think we can just enable it in RDoc by default. WDYT?

vinistock (approved):
This doesn't fail the RDoc generation process, so I don't see any reason why it wouldn't be turned on by default

PR #1277, opened 2025-01-15
CI: test.yml - Windows - add mswin, remove 3.4 exclude

MSP-Greg:
`ruby-core.yml` (ruby-core / RDoc under a ruby-core setup) is failing, due to recent 'default to bundled' change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,6 +11,7 @@ jobs:
     with:
       # 2.7 breaks `test_parse_statements_nodoc_identifier_alias_method`
       min_version: 3.0
+      versions: '["mswin"]'
       engine: cruby-truffleruby
 
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -25,9 +25,10 @@ jobs:
             ruby: truffleruby
           - os: windows-latest
             ruby: truffleruby-head
-          # Remove it once https://github.com/ruby/setup-ruby/issues/680 is fixed
+        include:
           - os: windows-latest
-            ruby: 3.4
+            ruby: mswin
```
kou, line 30: `ruby` -> `os` order will be better to align other job names:

```yaml
- ruby: mswin
  os: windows-latest
```

https://github.com/ruby/rdoc/actions/runs/12793372976/job/35666070365?pr=1277 : "test (3.0, ubuntu-latest)"
https://github.com/ruby/rdoc/actions/runs/12793372976/job/35666079817?pr=1277 : "test (windows-latest, mswin)"

Another approach: Add `versions: mswin` to `ruby-versions` and add `os: ubuntu-latest, ruby: mswin`/`os: macos-latest, ruby: mswin` excludes.

kou (approved):
+1

.github/workflows/test.yml
MSP-Greg: I updated to use the  `versions` input, I believe it wants a JSON array...

PR #1281, opened 2025-01-23
Add a `build:local_ruby` task to simplify local testing flow

st0012:
A common way to test RDoc changes is to build `ruby/ruby`'s documentation with the latest RDoc changes. When RDoc was a default gem, we can sync it to `ruby/ruby` with its `tool/sync_default_gems.rb` script.

Now that RDoc is a bundled gem, we need to use a different method to sync it to `ruby/ruby`. And so far building it and moving it to `ruby/ruby`'s bundled gems folder is the easiest way to do it.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -103,8 +103,9 @@ desc "Build #{Bundler::GemHelper.gemspec.full_name} and move it to local ruby/ru
 namespace :build do
   task local_ruby: :build do
     target = File.join("..", "ruby", "gems")
+
     unless File.directory?(target)
-      abort("Expected ruby to be cloned at the same level as #{Bundler::GemHelper.gemspec.full_name} to use this task")
+      abort("Expected Ruby to be cloned under the same parent directory as RDoc to use this task")
     end
 
     mv("#{path}.gem", target)
--- a/Rakefile
+++ b/Rakefile
@@ -99,6 +99,18 @@ task :clean do
   end
 end
 
+desc "Build #{Bundler::GemHelper.gemspec.full_name} and move it to local ruby/ruby project's bundled gems folder"
+namespace :build do
+  task local_ruby: :build do
+    target = File.join("..", "ruby", "gems")
+    unless File.directory?(target)
+      abort("Expected ruby to be cloned at the same level as #{Bundler::GemHelper.gemspec.full_name} to use this task")
+    end
+
+    mv("#{path}.gem", target)
+  end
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -99,6 +99,11 @@ task :clean do
   end
 end
 
+desc "Build #{Bundler::GemHelper.gemspec.full_name} and move it to local ruby/ruby project's bundled gems folder"
+task build_for_local_ruby: :build do
+  sh "mv #{path}.gem ../ruby/gems/"
```
vinistock, line 104: This has the implicit dependency of `ruby` being cloned on the same level as RDoc, right?

If so, how about we do this? Also, let's use `FileUtils` rather than `sh`, so that it's cross-platform.

```suggestion
  target = File.join("..", "ruby", "gems")

  unless File.directory?(target)
    abort("Expected ruby to be cloned at the same level as #{Bundler::GemHelper.gemspec.full_name} to use this task")
  end

  FileUtils.mv("#{path}.gem", target)
```

kou: We can use `mv` in `Rakefile`. It's better than `FileUtils.mv` because `mv` outputs log: `mv("#{path}.gem", target)`

kou (commented):
Can we use this in `.github/workflows/ruby-core.yml`?

Rakefile
```diff
@@ -99,6 +99,11 @@ task :clean do
   end
 end
 
+desc "Build #{Bundler::GemHelper.gemspec.full_name} and move it to local ruby/ruby project's bundled gems folder"
+task build_for_local_ruby: :build do
```
kou, line 103: Can we use better task name something like `local_ruby:build`?

Rakefile
vinistock: Will that work for Windows contributors though? Can we not just log manually and still ensure cross-platform support?

Rakefile
st0012: Renamed it to `build:local_ruby`

Rakefile
st0012: The `mv` here is an extension of `FileUtils`'s ([source](https://github.com/ruby/rake/blob/09a1cd41d13d0c2c23a0775f860a8c5e71d43911/lib/rake/file_utils_ext.rb#L33-L40)). So it should work cross-platform.

vinistock (approved):

PR #1289, opened 2025-02-09
Remove unnecessary coupling in comments' call_seq extraction

st0012:
There is no need to pass the method object to `RDoc::Comment#extract_call_seq` just to assign the result to the method object's `call_seq` attribute. This unnecessarily complicates the code and makes it harder to understand/refactor, which is evident in the method's tests.

This commit removes the indirection and simplifies the implementation & tests.

### DIFF
--- a/lib/rdoc/code_object/any_method.rb
+++ b/lib/rdoc/code_object/any_method.rb
@@ -110,7 +110,7 @@ class RDoc::AnyMethod < RDoc::MethodAttr
   # See also #param_seq
 
   def call_seq= call_seq
-    return if call_seq.empty?
+    return if call_seq.nil? || call_seq.empty?
 
     @call_seq = call_seq
   end
--- a/lib/rdoc/parser/c.rb
+++ b/lib/rdoc/parser/c.rb
@@ -810,7 +810,7 @@ class RDoc::Parser::C < RDoc::Parser
 
   def find_modifiers comment, meth_obj
     comment.normalize
-    comment.extract_call_seq meth_obj
+    meth_obj.call_seq = comment.extract_call_seq
 
     look_for_directives_in meth_obj, comment
   end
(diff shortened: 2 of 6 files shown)


lib/rdoc/code_object/any_method.rb
```diff
@@ -110,7 +110,7 @@ def call_seq
   # See also #param_seq
 
   def call_seq= call_seq
-    return if call_seq.empty?
+    return if call_seq.nil? || call_seq.empty?
 
     @call_seq = call_seq
```
kou, line 115: It's not related to this PR but we may want to set `@call_seq` even when `call_seq` is `nil` or `""`:

```ruby
call_seq = call_seq if call_seq&.empty?
@call_seq = call_seq
```

kou (approved):
+1

lib/rdoc/parser/c.rb
```diff
@@ -810,7 +810,7 @@ def find_const_comment(type, const_name, class_name = nil)
 
   def find_modifiers comment, meth_obj
     comment.normalize
-    comment.extract_call_seq meth_obj
+    meth_obj.call_seq = comment.extract_call_seq
```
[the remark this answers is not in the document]
st0012: I don't understand how this would change `@text.slice` as `extract_call_seq` still does it and it's still called here. Do I miss something?

lib/rdoc/code_object/any_method.rb
st0012: I thought about it but it's potentially risky if RDoc relies on this to avoid erasing the current value. Will need to do some testing to make sure it's ok to drop it.

lib/rdoc/parser/c.rb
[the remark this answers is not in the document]
vinistock: Oh, nvm, I totally read the code incorrectly.

lib/rdoc/code_object/any_method.rb
kou: It makes sense.

PR #1292, opened 2025-02-17
Remove unused attributes

st0012:
I think these attribute declarations are not necessary. Please review by commits.

### DIFF
--- a/lib/rdoc/code_object/method_attr.rb
+++ b/lib/rdoc/code_object/method_attr.rb
@@ -64,12 +64,6 @@ class RDoc::MethodAttr < RDoc::CodeObject
   attr_reader :arglists
 
   ##
-  # Pretty parameter list for this method
-
-  attr_reader :param_seq
-
-
-  ##
   # Creates a new MethodAttr from token stream +text+ and method or attribute
   # name +name+.
   #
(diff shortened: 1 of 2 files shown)


lib/rdoc/code_object/method_attr.rb
```diff
@@ -63,12 +63,6 @@ class RDoc::MethodAttr < RDoc::CodeObject
 
   attr_reader :arglists
 
-  ##
-  # Pretty parameter list for this method
-
-  attr_reader :param_seq
```
kou, line 69: It seems that `AnyMethod#arglists` uses `param_seq`: https://github.com/ruby/rdoc/blob/fa0a12bf2709e0d3ba4fa24d67962d08e4c223fe/lib/rdoc/code_object/any_method.rb#L88-L89

Can we remove the clause too?

st0012: No because `AnyMethod` defines its own `param_seq` method:

https://github.com/ruby/rdoc/blob/a253a8d163c36ede4064083d5d43f69f789d0c51/lib/rdoc/code_object/any_method.rb#L278

kou: Oh, sorry.

kou (approved):
+1

lib/rdoc/code_object/method_attr.rb
st0012: No worries 🙂 The codebase is pretty confusing now and that's why we need to refactor it first.
Please keep raising any concern you have as it's easy for me to get lost when looking at it too 😅

PR #1296, opened 2025-02-18
Refactor darkfish's classes sidebar implementation

st0012:
The current logic written in ERB template is super hard to read and understand. I think given its complexity is makes sense to  implement those logic in pure Ruby methods instead of ERB code in a view.

### DIFF
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -847,14 +847,14 @@ class RDoc::Generator::Darkfish
       if children = grouped_classes[index_klass.full_name]
         content << %(<li><details#{solo ? ' open' : ''}><summary>#{generate_class_link(index_klass, rel_prefix)}</summary>)
         content << traverse_classes(children, grouped_classes, rel_prefix)
-        content << '</ul></details>'
+        content << '</details></li>'
         solo = false
       elsif index_klass.display?
-        content << %(<li>#{generate_class_link(index_klass, rel_prefix)})
+        content << %(<li>#{generate_class_link(index_klass, rel_prefix)}</li>)
       end
     end
 
-    content
+    "#{content}</ul>"
   end
 
   def group_classes_by_namespace_for_sidebar(classes)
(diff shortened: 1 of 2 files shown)


lib/rdoc/generator/darkfish.rb
```diff
@@ -824,6 +824,50 @@ def generate_ancestor_list(ancestors, klass)
...
+        content << %(<li><details#{solo ? ' open' : ''}><summary>#{generate_class_link(index_klass, rel_prefix)}</summary>)
+        content << traverse_classes(children, grouped_classes, rel_prefix)
+        content << '</ul></details>'
+        solo = false
+      elsif index_klass.display?
+        content << %(<li>#{generate_class_link(index_klass, rel_prefix)})
+      end
+    end
+
+    content
```
tompng, line 857: Not actually related to this pull request change, but can we add a closing `</ul>` here, and closing `</li>` in `klasses.each`?
Closing `</li>` tag can be omitted but it would be much clear if the opening and closing tag corresponds.
Closing `</ul>` is mandatory (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul#technical_summary) and the html created by rdoc master seems invalid.

st0012: Good catch. Updated 👍

st0012:
BTW, I think we can later do the same refactor for `_sidebar_pages`, and it should be able to reuse some of the same logic we extracted out here.

lib/rdoc/generator/darkfish.rb
```diff
@@ -824,6 +824,50 @@ def generate_ancestor_list(ancestors, klass)
...
+  end
+
+  def traverse_classes(klasses, grouped_classes, rel_prefix, solo = false)
+    content = +'<ul class="link-list">'
+
+    klasses.each do |index_klass|
+      if children = grouped_classes[index_klass.full_name]
+        content << %(<li><details#{solo ? ' open' : ''}><summary>#{generate_class_link(index_klass, rel_prefix)}</summary>)
+        content << traverse_classes(children, grouped_classes, rel_prefix)
+        content << '</ul></details></li>'
```
tompng, line 850:
```suggestion
        content << '</details></li>'
```
and add closing `</ul>` in `traverse_classes`
```diff
# L857
-    content
+    "#{content}</ul>"
```

tompng (approved):

PR #1299, opened 2025-02-28
Relax paragraph pattern

### LINKED ISSUE #1298: Not All paragraphs start with a capital letter, as usual English text.
#1091 added excerpts.
But not all paragraphs in documentations start with a capital letter, as usual English text.

E.g., README.md in drb has the paragraph:

> dRuby is a distributed object system for Ruby.  It allows an object in one
> Ruby process to invoke methods on an object in another Ruby process on the
> same or a different machine.

`d` of `dRuby` should not be excluded, I think it should be any text, from the alphabet to the first period, excluding some symbols.

However, this is still too strict for most natural languages.

_Originally posted by @nobu in https://github.com/ruby/rdoc/pull/1091#discussion_r1973621411_

nobu:
Not all paragraphs in documentations start with a capital letter, as usual English text.

Fix #1298

### DIFF
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -700,7 +700,11 @@ class RDoc::Generator::Darkfish
     template
   end
 
-  ParagraphExcerptRegexp = /[A-Z][^\.:\/]+\./
+  # :stopdoc:
+  ParagraphExcerptRegexpOther = %r[\b\w[^./:]++\.]
+  # use \p/\P{letter} instead of \w/\W in Unicode
+  ParagraphExcerptRegexpUnicode = %r[\b\p{letter}[^./:]++\.]
+  # :startdoc:
 
   # Returns an excerpt of the comment for usage in meta description tags
   def excerpt(comment)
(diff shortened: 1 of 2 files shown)


lib/rdoc/generator/darkfish.rb
```diff
@@ -797,14 +800,22 @@ def excerpt(comment)
...
     # Match from a capital letter to the first period, discarding any links, so
     # that we don't end up matching badges in the README
-    first_paragraph_match = text.match(/[A-Z][^\.:\/]+\./)
-    return text[0...150].gsub(/\n/, " ").squeeze(" ") unless first_paragraph_match
+    pattern = ParagraphExcerptRegexpUnicode
+    begin
+      first_paragraph_match = text.match(pattern)
+    rescue Encoding::CompatibilityError
+      # The doc is non-ASCII text and encoded in other than Unicode base encodings.
+      raise unless pattern.eaual?(ParagraphExcerptRegexpUnicode)
```
vinistock, line 808: Should this be `equal?`?

lib/rdoc/generator/darkfish.rb
```diff
@@ -797,14 +800,22 @@ def excerpt(comment)
...
+      pattern = ParagraphExcerptRegexpOther
+      retry
```
vinistock, line 810: Is there no risk that this will enter an infinite loop? Is there no encoding scenario where `ParagraphExcerptRegexpOther` would fail?

nobu: Fixed at eb3a5ba6f2e38199758266124847e57174314179

lib/rdoc/generator/darkfish.rb
nobu: Fixed at eb3a5ba6f2e38199758266124847e57174314179

vinistock:
The fix looks good to me, but I don't know why only JRuby is raising a conversion error.

nobu:
Me too, and reported to JRuby (jruby/jruby#8682)

vinistock (approved):
Thank you, Nobu

PR #1300, opened 2025-03-01
Refactor Darkfish implementation

st0012:


lib/rdoc/generator/darkfish.rb
```diff
@@ -647,26 +601,28 @@ def render file_name
...
   # An io will be yielded which must be captured by binding in the caller.
 
-  def render_template template_file, out_file = nil # :yield: io
+  def render_template template_file, out_file: nil, locals: {}
     io_output = out_file && !@dry_run && @file_output
     erb_klass = io_output ? RDoc::ERBIO : ERB
 
     template = template_for template_file, true, erb_klass
 
+    @context = binding
```
nobu, line 610: Isn’t it possible that the template accidentally affects the variables of this method?

PR #1302, opened 2025-03-02
Centralize generator setup

st0012:
Discovered when working on #1300 

Currently we need to call `setup` in individual generation methods, despite having already called it in `RDoc#generate`. This is because `setup` is lazily called when we run `ri`'s server mode.

By calling `setup` eagerly in `Servlet#generator_for`, we can avoid the repeated calls to `setup` in individual generation methods.

Right now we can NOT move `setup` into `Darkfish#initialize` yet because `ri`'s server can[ initialize `Darkfish` with `nil` store](https://github.com/ruby/rdoc/blob/master/lib/rdoc/servlet.rb#L335).

### DIFF
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -301,8 +301,6 @@ class RDoc::Generator::Darkfish
   # Generate an index page which lists all the classes which are documented.
 
   def generate_index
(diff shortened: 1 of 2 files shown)


flavorjones (approved):
I think this is a good incremental improvement.

PR #1304, opened 2025-03-02
Remove unused `class_dir` and `file_dir` attributes from generators

st0012:
These generator methods are either not defined, or return a fixed `nil` value. Therefore, we can remove them and simplify some downstream logics accordingly.

### DIFF
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -634,7 +634,7 @@ class RDoc::ClassModule < RDoc::Context
   # Path to this class or module for use with HTML generator output.
 
   def path
-    http_url @store.rdoc.generator.class_dir
+    http_url(nil)
   end
 
   ##
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -634,7 +634,7 @@ class RDoc::ClassModule < RDoc::Context
   # Path to this class or module for use with HTML generator output.
 
   def path
-    http_url(nil)
+    http_url
   end
 
   ##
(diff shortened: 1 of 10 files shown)


kou (approved):
+1

jonathanhefner:
Hello! :wave:  This is a major breaking change for SDoc.  SDoc [overrides `class_dir` and `file_dir`](https://github.com/rails/sdoc/blob/805531f2620c0f0fe67039b885e7cccb6e1e1198/lib/sdoc/generator.rb#L96-L102) to prefix URL paths with `classes/` and `files/`, respectively.  For example, in the URLs for https://api.rubyonrails.org/classes/ActiveRecord/Base.html and https://api.rubyonrails.org/files/activerecord/lib/active_record/base_rb.html.

It should also be noted that RDoc < 6.13.0 _requires_ these methods when implementing a custom generator.  Omitting them will cause RDoc to raise a `NoMethodError`.

As a workaround, I monkey patched `RDoc::ClassModule` and `RDoc::TopLevel` in rails/sdoc@e2d849f48cb2ed3effe39a674ca79de7e8fb812b, but I think it would be better to revert this change.  Would it be possible to do so in RDoc 6.13.1?

st0012:
Sorry that I wasn't aware of `sdoc`'s use case of it. Can I propose an additional approach to restore customizing the prefixes than reverting the change tho as the original implementation was really not ideal IMO.

st0012:
I think customizing prefix in options is a better approach: https://github.com/ruby/rdoc/pull/1330

kou:
Oh, it may be better that we have a CI job for SDoc: https://github.com/ruby/rdoc/issues/1331

jeremyevans:
This change also broke expected file creation in the hanna gem.  I'll see if I can get the prefix option described in #1330 to work, but backwards incompatible changes like this should be reserved for major version bumps.  RDoc should not assume that Darkfish is the only generator, and just remove features that Darkfish doesn't use.

PR #1305, opened 2025-03-04
Reenable JRuby for testing

headius:
JRuby's recent issues with jar-dependencies caused installs to fail, but those issues have been resolved as of JRuby 9.4.12.0 (and by extension, jruby-head).

Reverts part of #1228.

Previous PR #1229 seemed to get messed up so I'm re-pushing here.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,6 @@ jobs:
       # 2.7 breaks `test_parse_statements_nodoc_identifier_alias_method`
       min_version: 3.0
       versions: '["mswin"]'
-      engine: cruby-truffleruby
 
   test:
     needs: ruby-versions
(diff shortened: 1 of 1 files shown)


headius:
This should be ready to merge. I filed a JRuby issue for the Windows failure: jruby/jruby#8672. When that's I'll reenable Windows CI for JRuby.

st0012:
@headius Thanks!

st0012:
There seems to be intermittent failing on ubuntu now ([example](https://github.com/ruby/rdoc/actions/runs/13748754286/job/38447004956)). Can you give it a look?

headius:
@st0012 At a glance I'd suspect the test is flawed, since execution of this code should not suddenly go non-linear after being linear previously. It seems like the harness could be made more robust, or could exercise the code more before attempting to measure tiny timeouts. At these small intervals, even the JVM JIT kicking in could cause us to slow down temporarily and unpredictably.

I will try to see if there's something actually broken.

headius:
@st0012 I have disabled the test for now and filed #1310.

PR #1308, opened 2025-03-09
Avoid accessing `RDoc` objects through `Store`

st0012:
There's an unhealthy coupling between `RDoc` and `Store`, where they hold reference of each other and access things both ways.

IMO, `RDoc` instances holding `Store` objects makes sense as `Store` is a component of it. But `Store` shouldn't need to reference back to its `RDoc` instance.

This PR is the first step of removing this coupling by dropping `Store#rdoc` usages.

### DIFF
--- a/lib/rdoc/store.rb
+++ b/lib/rdoc/store.rb
@@ -94,11 +94,7 @@ class RDoc::Store
 
   attr_accessor :path
 
-  ##
-  # The RDoc::RDoc driver for this parse tree.  This allows classes consulting
-  # the documentation tree to access user-set options, for example.
-
-  attr_accessor :rdoc
+  attr_writer :rdoc
 
   ##
   # Type of ri datastore this was loaded from.  See RDoc::RI::Driver,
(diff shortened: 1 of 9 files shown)


lib/rdoc/store.rb
```diff
@@ -94,11 +94,7 @@ def message # :nodoc:
 
   attr_accessor :path
 
-  ##
-  # The RDoc::RDoc driver for this parse tree.  This allows classes consulting
-  # the documentation tree to access user-set options, for example.
-
-  attr_accessor :rdoc
+  attr_writer :rdoc
```
tompng, line 97: I think `RDoc::Store` only needs `attr_accessor :options`, but we need to consider the initialization order of `rdoc.options` and `store.rdoc`. Maybe we can do it in another followup pull request later.

tompng (approved):

lib/rdoc/store.rb
st0012: I have a WIP PR for that (which was meant to be part of this PR). I still have a few tests to fix but can open a draft for it after this is merged.

PR #1309, opened 2025-03-09
Make `Options` a required constructor argument of `Store`

st0012:
Since `Store` relies on accessing `Options` for various tasks, it should simply hold a reference to the `Options` object itself, instead of accessing it through an `RDoc` instance and forming an unnecessary coupling.

This PR requires `Store` to be initialized with an `Options` instance.

### DIFF
--- a/test/rdoc/test_rdoc_ri_driver.rb
+++ b/test/rdoc/test_rdoc_ri_driver.rb
@@ -460,7 +460,7 @@ class RDocRIDriverTest < RDoc::TestCase
     #
     # Y is not chosen randomly, it has to be after Object in the alphabet
     # to reproduce https://github.com/ruby/rdoc/issues/814.
-    store = RDoc::RI::Store.new @home_ri
+    store = RDoc::RI::Store.new(path: @home_ri)
     store.cache[:ancestors] = { "Z" => ["Object", "Y"], "Y" => ["X"] }
     store.cache[:modules] = %W[X Y Z]
     @driver.stores = [store]
@@ -538,7 +538,7 @@ class RDocRIDriverTest < RDoc::TestCase
   end
 
   def test_complete
-    store = RDoc::RI::Store.new @home_ri
+    store = RDoc::RI::Store.new(path: @home_ri)
     store.cache[:ancestors] = {
       'Foo'      => %w[Object],
       'Foo::Bar' => %w[Object],
@@ -963,7 +963,7 @@ Foo::Bar#bother
   end
 
   def test_expand_class_2
-    @store1 = RDoc::RI::Store.new @home_ri, :home
+    @store1 = RDoc::RI::Store.new(path: @home_ri, type: :home)
 
     @top_level = @store1.add_file 'file.rb'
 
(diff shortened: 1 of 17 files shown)


kou (approved):
+1

Can we move https://github.com/ruby/rdoc/blob/c809ba889630b393c9bff5e19afe09b512c312a0/lib/rdoc/rdoc.rb#L472-L476 to `Store`?

st0012:
@kou Updated 👍

test/rdoc/test_rdoc_ri_driver.rb
```diff
@@ -1007,7 +1008,7 @@ def test_expand_name
 
     assert_equal 'Z', e.name
 
-    @driver.stores << RDoc::Store.new(nil, :system)
+    @driver.stores << RDoc::Store.new(@rdoc_options, path: nil, type: :system)
```
kou, line 1011:
```suggestion
    @driver.stores << RDoc::Store.new(@rdoc_options, type: :system)
```

test/rdoc/test_rdoc_ri_driver.rb
```diff
@@ -1084,8 +1085,8 @@ def test_filter_methods_not_found
   end
 
   def test_find_store
-    @driver.stores << RDoc::Store.new(nil,              :system)
-    @driver.stores << RDoc::Store.new('doc/gem-1.0/ri', :gem)
+    @driver.stores << RDoc::Store.new(@rdoc_options, path: nil, type: :system)
```
kou, line 1088:
```suggestion
    @driver.stores << RDoc::Store.new(@rdoc_options, type: :system)
```

kou (approved):
+1

jonathanhefner:
Hello, again! :wave:  This is a breaking change for SDoc.  As part of a performance optimization for its test suite, SDoc [manually constructs fresh `RDoc::Store` instances](https://github.com/rails/sdoc/blob/805531f2620c0f0fe67039b885e7cccb6e1e1198/spec/spec_helper.rb#L31).  The new signature of `RDoc::Store#initialize` breaks `RDoc::Store.new()`.

`RDoc::Store#initialize` appears to be a [documented public API](https://docs.ruby-lang.org/en/3.3/RDoc/Store.html#method-c-new).  Is that accurate?

st0012:
It's unfortunate that it added additional effort to upgrade, but I don't consider `Store` part of the official API. In fact, the only public components are probably `RDoc::RDoc`, `RDoc::Options`, and some of the Darkfish APIs.
As you probably are aware, RDoc is a legacy codebase and if we consider everything that's been documented as public, then it will take forever to improve its codebase enough to clear the ever-growing features backlog.

mullermp:
I ran into an issue with this change as well. I was using RDoc to parse documentation for an rspec matcher. I'm not sure if I am using it correctly, maybe I can get some guidance? I'm doing something like this now:

```
    rdoc = RDoc::RDoc.new
    options = RDoc::Options.load_options
    rdoc.options = options
    rdoc.store = RDoc::Store.new(options)
    top_level = rdoc.parse_files([file]).first
    documentation = top_level.find_class_or_module(klass)
    documentation = documentation.find_method_named(method) if method
    @actual = documentation.comment.text
    @expected = expected.chomp
    expect(@actual).to include(@expected)
```

PR #1315, opened 2025-03-16
Remove dead or unnecessary methods/attributes

st0012:
2 types of targets in this PR:
- Completely unused methods (outside of tests, ofc)
- Single-used or unnecessary methods. Like:

```rb
  def each_constant # :yields: constant
    @constants.each {|c| yield c}
  end
```

### DIFF
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -56,7 +56,6 @@ class RDoc::ClassModule < RDoc::Context
 
     klass.parent = mod.parent
     klass.section = mod.section
-    klass.viewer = mod.viewer
 
     klass.attributes.concat mod.attributes
     klass.method_list.concat mod.method_list
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -34,8 +34,6 @@ class RDoc::ClassModule < RDoc::Context
 
   attr_accessor :comment_location
 
-  attr_accessor :diagram # :nodoc:
-
   ##
   # Class or module this constant is an alias for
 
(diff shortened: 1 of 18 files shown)


test/rdoc/test_rdoc_class_module.rb
```diff
@@ -1466,66 +1466,6 @@ def test_update_includes_with_colons
...
-    b = RDoc::Extend.new 'M2', nil
-    c = RDoc::Extend.new 'C', nil
-
-    @c1.add_extend a
-    @c1.add_extend b
-    @c1.add_extend c
-    @c1.each_extend do |extend| extend.module end # cache extended modules
-
-    @m1_m2.document_self = nil
-    assert @m1_m2.remove_from_documentation?
```
st0012, line 1480: These same test assertions actually appear in other tests like `test_update_includes_with_colons` as well. So we don't need to keep them here either.

kou (approved):
+1

lib/rdoc/code_object/class_module.rb
```diff
@@ -831,22 +827,6 @@ def update_includes
     includes.uniq!
   end
 
-  ##
-  # Deletes from #extends those whose module has been removed from the
-  # documentation.
-  #--
-  # FIXME: like update_includes, extends are not reliably removed
-
-  def update_extends
```
tompng, line 840: I think we should keep this method. It should be called from method `complete` just like `update_includes`

Example: "Extended With Modules" section of https://docs.ruby-lang.org/en/master/CGI.html
```
Extended With Modules
CGI::Util
CGI::Util
```
Note that other modules (`QueryExtension`, `Html3`, etc) are just wrongly listed by `rdoc/parser/ruby.rb`

st0012: https://github.com/ruby/rdoc/pull/1317 for fixing the issue

rafaelfranca:
This change broke sdoc, because we use `last_modified` on it. I don't think rdoc should be removing public API methods in minor releases.

PR #1316, opened 2025-03-16
Fix test target code of module_function

tompng:
The test target code was wrong. `module_function` is only available inside module, not inside class.

`module_function` is `undef`ed in Class.
```ruby
Class.undefined_instance_methods
=> [:prepend_features, :extend_object, :refine, :module_function, :append_features]
```


kou (approved):
+1

PR #1317, opened 2025-03-16
Call` update_extend`s in `ClassModule#complete`

st0012:
For unknown reasons, this method was not being called in `#complete`, which led to duplicated extends in some cases.


tompng (approved):
:+1:

PR #1320, opened 2025-03-20
Image alt tag header formatting

ioquatix:
This allows us to parse the alt text from Markdown images, formatting a header text correctly (using the alt text).

Without this PR, if a Markdown file has an image instead of a title, that image will be used in the table of contents, which is unreadable and undesirable.

Used in <https://github.com/rack/rack/pull/2313>.

### DIFF
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -992,9 +992,14 @@ Strike = &{ strike? }
          "~~"
          { strike a.join }
 
-# TODO alt text support
-Image = "!" ( ExplicitLink | ReferenceLink ):a
-        { "rdoc-image:#{a[/\[(.*)\]/, 1]}" }
+Image = "!" ExplicitLink:a
+        {
+          # Extract alt text and URL
+          alt_text = a[/\{(.*?)\}/, 1] || ""
+          url = a[/\[(.*?)\]/, 1] || ""
+
+          "rdoc-image:#{url}:#{alt_text}"
+        }
 
 Link =  ExplicitLink | ReferenceLink | AutoLink
 
--- a/lib/rdoc/markup/heading.rb
+++ b/lib/rdoc/markup/heading.rb
@@ -68,8 +68,8 @@ RDoc::Markup::Heading =
   def plain_html
     text = self.text.dup
 
-    if text.match?(/rdoc-image:[^:]+:(.*)/)
-      text = text.match(/rdoc-image:[^:]+:(.*)/)[1]
+    if matched = text.match(/rdoc-image:[^:]+:(.*)/)
+      text = matched[1]
     end
 
     self.class.to_html.to_html(text)
--- a/lib/rdoc/markup/heading.rb
+++ b/lib/rdoc/markup/heading.rb
@@ -66,7 +66,13 @@ RDoc::Markup::Heading =
   # element.
 
   def plain_html
-    self.class.to_html.to_html(text.dup)
+    text = self.text.dup
+
+    if text.match?(/rdoc-image:[^:]+:(.*)/)
+      text = text.match(/rdoc-image:[^:]+:(.*)/)[1]
+    end
+
+    self.class.to_html.to_html(text)
   end
 
   def pretty_print q # :nodoc:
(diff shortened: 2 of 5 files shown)


lib/rdoc/markup/heading.rb
```diff
@@ -66,7 +66,13 @@ def label context = nil
   # element.
 
   def plain_html
-    self.class.to_html.to_html(text.dup)
+    text = self.text.dup
+
+    if text.match?(/rdoc-image:[^:]+:(.*)/)
+      text = text.match(/rdoc-image:[^:]+:(.*)/)[1]
+    end
```
st0012, line 73:
```suggestion
    if matched = text.match(/rdoc-image:[^:]+:(.*)/)
      text = matched[1]
    end
```

st0012: Can we also add a test case in `test/rdoc/test_rdoc_markup_heading.rb`?

Something like:

```rb
  def test_plain_html_using_image_alt_as_text
    h = RDoc::Markup::Heading.new 1, 'rdoc-image:foo.png:Hello World'

    assert_equal 'Hello World', h.plain_html
  end
```

lib/rdoc/markdown.kpeg
```diff
@@ -992,9 +992,14 @@ Strike = &{ strike? }
...
          { strike a.join }
 
-# TODO alt text support
-Image = "!" ( ExplicitLink | ReferenceLink ):a
-        { "rdoc-image:#{a[/\[(.*)\]/, 1]}" }
+Image = "!" ExplicitLink:a
+        {
+          # Extract alt text and URL
+          alt_text = a[/\{(.*?)\}/, 1] || ""
+          url = a[/\[(.*?)\]/, 1] || ""
```
tompng, line 999: `a` is a String that always match to these regexp because `ExplicitLink:a` is created by `"{#{l}}[#{s}]"`.
Perhaps it could be more simple if `ExplicitLink` returns structured data
```diff
  ExplicitLink =  Label:l "(" @Sp Source:s Spnl Title @Sp ")"
-                 { "{#{l}}[#{s}]" } # returns string, need to parse this string again...
+                 { { label: l, source: s } } # can return hash
```

(need to change `Link =  ExplicitLink | ReferenceLink | AutoLink` to return a string if ExplicitLink returns hash)

lib/rdoc/markup/to_html.rb
```diff
@@ -98,7 +98,12 @@ def handle_RDOCLINK url # :nodoc:
 
       gen_url CGI.escapeHTML(url), CGI.escapeHTML(text)
     when /^rdoc-image:/
-      %[<img src=\"#{CGI.escapeHTML($')}\">]
+      url, alt = $'.split(":", 2) # Split the string after "rdoc-image:" into url and alt
+      if alt
```
tompng, line 102: I think we can omit creating `alt=""` if alt is an empty string

lib/rdoc/markup/heading.rb
tompng: If the intention is to remove image and replace it with alt text, `# foo alt bar` → `foo alt bar`
I think gsub is better, and need to consider rdoc-image without alt text.

```
Text1 rdoc-image:img.jpg:alt_text text2 rdoc-image:img_without_alt.jpg text3
↓
Text1 alt_text text2 text3
```

ioquatix: @tompng I don't understand how to action your suggestion.

lib/rdoc/markdown.kpeg
ioquatix: Could we do this in a follow-up PR? I think it's out of scope for this PR.

lib/rdoc/markup/to_html.rb
ioquatix: Makes sense.

lib/rdoc/markdown.kpeg
tompng: Makes sense :+1:

st0012 (approved):
Thank you!

ioquatix:
Thanks everyone!

PR #1321, opened 2025-03-20
PoC for the event-based plugin system with YARD parsing plugin

okuramasafumi:
This PR introduces the event-based plugin system along with a YARD-parsing plugin as an example of concrete plugins.

It demonstrates how we can build practical plugins while keeping the core code small. In this case, YARD-related logic is encapsulated in a plugin and used only when users specify with `--plugins` option.
This YARD plugin successfully parses https://github.com/okuramasafumi/alba comments written with YARD, but doesn't print YARD-specific information such as types since current RDoc doesn't support type information.

This PR is completely proof of concept, and is not meant to be merged now.
I'd like to discuss if this is the right way to extend RDoc for developers, and even for core team, and improve maintainability.


lib/rdoc/options.rb
```diff
@@ -868,6 +876,12 @@ def parse argv
 
       opt.separator nil
 
+      opt.on("--plugins=PLUGINS", "-P", Array, "Use plugins") do |value|
+        @plugins.concat value
+      end
```
kou, line 881: We can use multiple `--plugin` to add multiple plugins.

```suggestion
      opt.on("--plugin=PLUGIN", "-P", "Use plugin") do |plugin|
        @plugins << plugin
      end
```

lib/rdoc/rdoc.rb
```diff
@@ -449,6 +455,9 @@ def document options
     end
     @options.finish
 
+    ::RDoc::BasePlugin.activate_with(self)
+    @options.load_plugins
```
kou, line 459: If we load plugins dynamically and use `require`, plugins will not be loaded to multiple RDoc instance.

Do we need to load plugins for each `RDoc::RDoc#document`?

lib/rdoc/event_registry.rb
kou, line 1: We already have a plugin system in `RDoc::Markup::PreProcess`.

Could you explain what is the merit of "the event-based plugin system" you suggested (and problems in the existing system)? Should existing `RDoc::Markup::PreProcess` be deprecated by this?

lib/rdoc/yard_plugin.rb
```diff
@@ -0,0 +1,228 @@
...
+
+module RDoc
+  class YardPlugin < BasePlugin
+    listens_to :rdoc_store_complete do |env, store|
+      store.all_classes_and_modules.each do |cm|
+        cm.each_method do |meth|
+          puts "Parsing #{meth.name}"
+          parsed_comment = Parser.new(meth.comment.text).parse
+          # meth.params = parsed_comment.param.map(&:to_s).join("\n")
+          meth.comment.text = parsed_comment.plain.join("\n")
```
kou, line 15: Plugins should update the given objects destructively?
How to extend RDoc by a plugin when existing RDoc objects don't have a place to store additional data? In this case, how to use the parsed type information?

lib/rdoc/yard_plugin.rb
```diff
@@ -0,0 +1,228 @@
...
+        @comment = comment
+        @parsed_comment = ParsedComment.new([], nil, [], [])
+        @mode = :initial
+        @base_indentation_level = 0 # @comment.lines.first[/^#\s*/].size
+      end
+
+      def parse
+        @comment.each_line do |line|
+          current_indentation_level = line[/^#\s*/]&.size || 0
+          if current_indentation_level >= @base_indentation_level + 2
```
kou, line 55: Does this work when the first line if `#  ...`?

lib/rdoc/yard_plugin.rb
```diff
@@ -0,0 +1,228 @@
...
+          \[(?<type1>[^\]]+)\]\s+(?<name1>\S+)\s*(?<desc1>.*)? |  # [Type] name desc
+          (?<name2>\S+)\s+\[(?<type2>[^\]]+)\]\s*(?<desc2>.*)?   # name [Type] desc
+          )
+        /x,
+          return: /@return\s+\[(?<type>[^\]]+)\]\s*(?<desc>.*)?/,
+          raise: /@raise\s+\[(?<type>[^\]]+)\]\s*(?<desc>.*)?/
+      }
+      def initialize(comment)
+        @comment = comment
+        @parsed_comment = ParsedComment.new([], nil, [], [])
```
kou, line 47: It seems that we can use a local variable not an instance variable for this.

lib/rdoc/options.rb
okuramasafumi: I'm not sure which interface is better.

```sh
# This
rdoc --plugins yard_plugin,another_plugin

# Or that
rdoc --plugin yard_plugin --plugin another_plugin
```

Correct me if I'm wrong, but I think the later way is kind of rare.

lib/rdoc/rdoc.rb
okuramasafumi: Agree, `initialize` is a better place?

lib/rdoc/event_registry.rb
okuramasafumi: I believe that `RDoc::Markup::PreProcess` requires adding directives to existing code, and that is a problem. One of the advantages of a new plugin system is that just enabling plugins works without modifying existing code.

> Should existing RDoc::Markup::PreProcess be deprecated by this?

Using directives has its own advantages, so I'd like to keep it.

lib/rdoc/yard_plugin.rb
```diff
@@ -0,0 +1,228 @@
...
+            data.append_desc(line)
+          else
+            if (tag, matchdata = matching_any_tag(line))
+              if tag == :param
+                type = matchdata[:type1] || matchdata[:type2]
+                name = matchdata[:name1] || matchdata[:name2]
+                desc = matchdata[:desc1] || matchdata[:desc2]
+                parsed_type = TypeParser.parse(type)
+                @parsed_comment[:param] << ParamData.new(type: parsed_type, name: name, desc: desc)
+                @mode = :param
```
tompng, line 67: Storing previous_tag here instead of storing `mode = :param`, appending part will be simple
```ruby
tag = ParamData.new(type: parsed_type, name: name, desc: desc)
@parsed_comment[:param] << tag # This could be `@param << tag` or `param << tag`
previous_tag = tag
```

```diff
# Append to the previous tag
- data = @mode == :param || @mode == :raise ? @parsed_comment[@mode].last : @parsed_comment[@mode]
- data.append_desc(line)
+ previous_tag.append_desc(line)
```

lib/rdoc/yard_plugin.rb
```diff
@@ -0,0 +1,228 @@
...
+    class TypeParser
+      TOKENS = {
+        collection_start: /</,
+        collection_end: />/,
+        fixed_collection_start: /\(/,
+        fixed_collection_end: /\)/,
+        type_name: /#\w+|((::)?\w+)+/,
+        literal: /(?:
+           '(?:\\'|[^'])*' |
+           "(?:\\"|[^"])*" |
```
tompng, line 163: I think some regexps are wrong.
`"\"` (escaped, quote is not closed) matches to this string literal regexp.
Integer and Float literal(integer part) matches to `type_name: /#\w+|((::)?\w+)+/`

lib/rdoc/yard_plugin.rb
okuramasafumi: True, thank you for pointing out.

lib/rdoc/yard_plugin.rb
okuramasafumi: Honestly this indentation-related implementation is not perfect and will fail for many edge cases. I'll definitely add test cases for this.

lib/rdoc/yard_plugin.rb
okuramasafumi: Awesome, I'll take this!

lib/rdoc/yard_plugin.rb
okuramasafumi: Oh I didn't notice this. I tested this code with some variations of type and it seems working. Tests are missing and we need to add them to ensure this regex working correctly since it's quite complex.

lib/rdoc/options.rb
kou: What does "rare" mean here?
The latter is used in many cases. For example, `ruby`'s `-I` is the latter interface. `curl`'s `--data`/`--header`/... are also the latter interface.

The latter style is easyer to use from a script:

```bash
args=()
if [ "$USE_X_PLUGIN" = "yes" ]; then
  args+=(--plugin X_plugin)
fi
if [ "$USE_Y_PLUGIN" = "yes" ]; then
  args+=(--plugin Y_plugin)
fi
rdoc "${args[@]}" ...
```

lib/rdoc/rdoc.rb
kou: `#initialize` is better than `#document` but do we need to load plugins for each RDoc instance?
Can we load plugins only once in one process?

lib/rdoc/event_registry.rb
kou: It seems that `RDoc::Markup::PreProcess` has the "post_process" hook too: https://github.com/ruby/rdoc/blob/894b2f1cd147a760449a52b621af7bf206452127/lib/rdoc/markup/pre_process.rb#L25-L32

Should we deprecate it by suggested plugin system?

lib/rdoc/yard_plugin.rb
okuramasafumi: Yes I consider this and don't have a confident conclusion now.
I think there are two ways, one is using `env` object and another is modifying existing objects (`meth` in his case).
Using env would be safer, and storing type information would look like:

```ruby
# In a plugin
env[meth] = DataFromYardPlugin.new(type: type)

# In a generator
type = env[meth].type
```

Here, `DataFromYardPlugin` contains various information along with type.

lib/rdoc/base_plugin.rb
```diff
@@ -0,0 +1,20 @@
...
+
+    def self.listens_to(event_name, &block)
+      rdoc.event_registry.register(event_name, block)
+    end
+
+    # Activate the plugin with the given RDoc instance
+    # Without calling this, plugins won't work
+
+    def self.activate_with(rdoc = ::RDoc::RDoc.current)
+      @@rdoc = rdoc
```
st0012, line 13: Using class variable for this means any later plugin calls `activate_with` with its own `RDoc` will apply that to all other plugins. I don't think this is what we want?


```rb
class Base
  @@name = "Base"

  def name
    @@name
  end

  def self.update_name(name)
    @@name = name
  end
end

class Foo < Base
end

class Bar < Base
end

Foo.update_name("Foo")
f = Foo.new

Bar.update_name("Bar")
b = Bar.new

puts f.name # => "Bar"
puts b.name # => "Bar"

```

lib/rdoc/options.rb
okuramasafumi: @kou Thank you for correcting me, after some investigation, I noticed that we have both styles in Ruby and other tools.
Personally I don't have a strong opinion here. For the future version we can consider YAML configuration as a primary way to enable plugins (like [rubocop](https://docs.rubocop.org/rubocop/1.74/configuration.html)).
I'm not against using multiple `--plugin` option here.

lib/rdoc/rdoc.rb
okuramasafumi: I cannot see that using multiple `RDoc::RDoc` instance is common. From CLI or Rake task, I think we have only one instance of `RDoc::RDoc`.
However, we might have multiple instances in the future, probably for concurrency.
So then, where is the ideal place to activate plugins? I through it's `RDoc::Rdoc` since it's in the execution path anyway.

lib/rdoc/event_registry.rb
okuramasafumi: As far as I can see, this `PreProcess` works with directives, and this time it's not I want to work with. The goal of this PR is to introduce a way to make RDoc understand YARD without modifying existing code.
That said, we don't have to (or should not) deprecate `post_process` here since it's about directives and removing it will be backward-incompatible, and cannot be replaced by the plugin system in this PR.

lib/rdoc/base_plugin.rb
okuramasafumi: I thought it's safe to use a class variable here since we don't have to update `rdoc` object.
However, some information such as a name of the plugin, should be stored as a class instance variables.

st0012: - If we don't expect plugins to call this method, then I don't think it should be in the base class as it is.
- It doesn't look like the yard plugin needs this attribute yet? If that's the case, let's not introduce this until we have a real use case to help us define the API better.
- If we want to pass RDoc's states down to plugins, let's avoid passing the `RDoc` object unless we have a solid use case for it. There's a lot of unnecessary coupling between RDoc's major components and I have been untangling for a while ([example](https://github.com/ruby/rdoc/pull/1308)). Exposing `RDoc` directly will potentially make it harder for such work as all of its attributes & methods will be considered public APIs.

okuramasafumi: You are right, I didn't notice that it's not used at all now. I used it at first, but `store` has enough information for the plugin so I removed it.
Then we can remove this class variable. And I agree, we should avoid `RDoc::RDoc` object (thank you for untangling it!)

st0012: Yeah I'd expect we expose `Options` and `Store` objects first if we need to share states with plugins.

okuramasafumi: https://github.com/ruby/rdoc/pull/1321/files#diff-2458b7fcdf31ccd44fa0681ba7e381360b3e4623c9f5409358222e4826c3c60dR6

Sorry it turned out that we do use `rdoc` object here because we need some global data store for event handling.
I still agree that we should avoid `RDoc::RDoc` instance, but then where should we put these event registration logic? Maybe we can introduce something like `PluginManager` and pass it to plugins?

lib/rdoc/rdoc.rb
kou: `gem install A B C` creates a `RDoc::RDoc` for each package.

lib/rdoc/yard_plugin.rb
kou: If there are multiple plugins that want to attach additional information to `meth`, the approach may be fragile. (One plugin may overwrite `env[meth]` set by another plugin.)

Can we provide more safer mechanism?

okuramasafumi: I think there are two ways of overriding: intentional and unintentional.

If the overriding is unintentional, it should be prevented. It can be done with something like below:

```ruby
require 'logger'

module RDoc
  class Environment
    def initialize
      @data = {}
      @logger = Logger.new($stderr)
      @logger.level = Logger::INFO
      @hooks = {}
    end

    def [](key)
      @data[key]
    end

    def []=(key, value)
      old_value = @data[key]
      @data[key] = value
      @logger.debug("Set env[#{key.inspect}] = #{value.inspect}")
      trigger_hooks(key, old_value, value)
    end

    def delete(key)
      @logger.debug("Delete env[#{key.inspect}]")
      old_value = @data.delete(key)
      trigger_hooks(key, old_value, nil)
    end

    def keys
      @data.keys
    end

    def fetch(key, default = nil)
      val = @data.fetch(key, default)
      @logger.debug("Fetch env[#{key.inspect}] => #{val.inspect}")
      val
    end

    def to_h
      @data.dup
    end

    def logger
      @logger
    end

    def on_key_set(key, &block)
      (@hooks[key] ||= []) << block
    end

    private

    def trigger_hooks(key, old_value, new_value)
      return unless @hooks[key]
      @hooks[key].each do |hook|
        begin
          hook.call(key, old_value, new_value, self)
        rescue => e
          @logger.error("Error in hook for key #{key.inspect}: #{e.message}")
          raise
        end
      end
    end
  end
end

env = RDoc::Environment.new
env[:foo] = 'foo'
env.on_key_set(:foo) { raise 'Do not override :foo }
env[:foo] = 'bar' # => Error
```

In this approach, preventing override is a responsibility of those who want to do so. In other words, this is not automatic. This is good because if it allows other plugins to override it, it just leaves it overridable.

lib/rdoc/event_registry.rb
okuramasafumi: @kou Yes it seems it calls hooks for every method. For YARD parsing purpose, this might work as well.

There are three issues though.

1. The naming, `PreProcess.post_process` looks so weird
2. It needs to modify `code_object` directly, resulting unexpected behaviors
3. Low extensibility, it can work only with `code_object` and not with other things

As the name suggests, this API is limited to add other markups. So, I believe we can eventually deprecate this in favor our new plugin APIs.

lib/rdoc/yard_plugin.rb
kou: I feel that attaching information to a code object is a common pattern for RDoc plugins.
(In your example, attaching type information to each method object.)

If the conflict is a rare situation, your approach may work. But if it's a common pattern, your approach will be inconvenient.

Do you think whether attaching information to a code object is a common pattern or not?

okuramasafumi: > Do you think whether attaching information to a code object is a common pattern or not?

I did some investigation, but could not find enough evidences that shows "attaching information to a code object is a common pattern".
I do think that it's potentially dangerous to modify code objects, especially because the maintainers might want to modify code object in the future. Using `env` object is safer.

kou: Ah, sorry. "attaching information to a code object" means both of `object.ADDTIONAL_INFORMATION = XXX` and `env[object] = XXX` cases. It's not related to "how to implement".

For example, if we use `object.metadata[:types] = types` (`object.metadata` returns a `Hash`) approach, we can attach multiple information to the object such as `object.metadata[:types]` and `object.metadata[:my_property]`. If we use `env[object] = XXX` approach, we can't attach multiple information to the object because it has only one space. But they are "how to implement". For example, we can use `env[[object, :types]]` with the `env` approach.

lib/rdoc/rdoc.rb
okuramasafumi: I see, that's pretty common...
Would we want separate plugins and their configurations for each `RDoc::RDoc` instance? That said, if we want to isolate each environment, loading plugins for each instance is fine. If not, then we can load plugins all at once at somewhere else (with `RDoc` module maybe?)

lib/rdoc/yard_plugin.rb
okuramasafumi: Thank you for clarification, then yes, extracting information from somewhere like comment string and attaching it to code objects would be typical for plugins.

kou: OK. Then `env[object] = XXX` and `env.on_key_set(object) { raise "Do not override #{object.inspect}" }` will be inconvenient.

lib/rdoc/rdoc.rb
kou: Who does configure RDoc plugins?

If each package have RDoc plugins configuration, we may need per package plugins configuration. If users configure RDoc plugins, we may be able to share the same plugins configuration.

lib/rdoc/yard_plugin.rb
okuramasafumi: Yes, preventing override by default is an option.

lib/rdoc/rdoc.rb
okuramasafumi: I thought only users will configure RDoc plugins, but noticed that each gem can configure RDoc along with its plugins with `rdoc_options`.

https://github.com/ruby/rdoc/blob/9f0b3adda279dd5a5c2229b65204c11b37056232/lib/rdoc/rubygems_hook.rb#L167

And I guess that's why we instantiate an RDoc instance for each package in the first place. So loading plugins for each package might be necessary.

lib/rdoc/yard_plugin.rb
kou: If the prevention happen, what should we (users? plugin developers) do as the next action?
Do we change key for `env` from `env[object] = ...` to something? What key should be used for the case?

lib/rdoc/rdoc.rb
kou: We may not need to `require` for registering a plugin. We may just need to instantiate a plugin for it. For example:

```ruby
module RDoc
  class YardPlugin < BasePlugin
    def on_rdoc_store_complete(store)
      # ...
    end
  end

  class RDoc
    def document(options)
      # ...
      plugins = options.plugins.collect do |name|
        instantiate_plugin(name) # name.capitalize.constantize.new or associating name with class or something...
      end
      # ...
      plugins.each do |plugin|
        plugin.on_rdoc_store_complete(@store) if plugin.respond_to?(:on_rdoc_store_complete)
      end
      # ...
    end
  end
end
```

lib/rdoc/event_registry.rb
kou: Could you update the PR description and clarifies the current plugin system problems and how to resolve them by the suggested plugin system in the PR description?

(I thought that I wrote this comment a few days ago but it was not written...)

lib/rdoc/rdoc.rb
okuramasafumi: Yes this is absolutely possible, in this case `on_some_event` style looks great and seems better than `listens` DSL style.

lib/rdoc/event_registry.rb
okuramasafumi: @tompng TomDoc support is definitely a candidate of plugin extraction.

> Supporting more YARD directives will will also helps it.

Plugin system makes it possible to "bring your own YARD tag". I think we need to make plugins inheritable to achieve this.

lib/rdoc/yard_plugin.rb
okuramasafumi: @kou Does https://github.com/ruby/rdoc/discussions/1257 describe problems well enough? If not please tell me so that I can add more information and details.

lib/rdoc/event_registry.rb
okuramasafumi: @kou Does https://github.com/ruby/rdoc/discussions/1257 describe problems well enough? If not please tell me so that I can add more information and details.

st0012: It'd be great if you can answer my [questions](https://github.com/ruby/rdoc/discussions/1257#discussioncomment-11677107) there too 🙂

lib/rdoc/yard_plugin.rb
okuramasafumi: My current idea is to limit Hash-like method (`[]=`) to only add value, and if existing data is there, it raises an exception. Instead we offer `update` method so that users (mostly plugin authors) can intentionally override existing values.

```ruby
    def []=(key, value)
      old_value = @data[key]
      raise if old_value

      @data[key] = value
      @logger.debug("Set env[#{key.inspect}] = #{value.inspect}")
      trigger_hooks(key, old_value, value)
      value
    end
    alias add []=

    def update(key, value)
      old_value = @data[key]
      @data[key] = value
      @logger.debug("Set env[#{key.inspect}] = #{value.inspect}")
      trigger_hooks(key, old_value, value)
      value
    end

```

lib/rdoc/event_registry.rb
okuramasafumi: @st0012 Oh, really sorry it completely slipped my mind. I added a reply there.

tompng: Currently, "event hook is enough" is just a hypothesis.
To prove it, more examples (ex: TomDoc, YARD with more implementation) are needed. I think it will reveal that something is missing, which strongly affects the design of this plugin system.

Of course one decision is to only support plugins that is possible by event hooks. But we need to know the limitation of it.
Perhaps we need to give up some tags(`@attr_reader` `@attr_writer` `@attr_accessor`) and directives(`@!group` `@!macro` `@!attribute` `@!method` `@!parse`), which might be possible to implement depending on the choice of API designs.
Does this plugin system only support implementing YARD tags? Is giving up `@attr_reader`, `@!attribute` and `@!method` OK?

kou: > Does #1257 describe problems well enough? If not please tell me so that I can add more information and details.

Oh, sorry. I missed it.

It seems that the proposal doesn't explain:

* the current plugin system such as `RDoc::Markup::PreProcess`,
* what we can do with the current plugin system,
* what we can't do with the current plugin system,
* what are the problems of the current plugin system,
* what are needed for a new plugin system,
* ...

Could you update #1257?

In addition, I think that this PR's description should have how to solve the problems of the current plugin system and how to satisfy needed features for a new plugin system.

lib/rdoc/yard_plugin.rb
kou: With the API, the following case doesn't work, right?

```ruby
class YardPlugin
  def XXX(env, ...)
    env[meth] = yard_tags
  end
end

class AnotherPlugin
  def XXX(env, ...)
    env[meth] = another_information # raise
  end
end
```

If "attaching information to a code object" is a common pattern and we recommend to use `env[meth] = ...` for it, enabling multiple plugins will be broken easily. Most plugins will use `env.update(meth, ...)` to solve it but it will lost existing information. (`yard_tags` in this case.) It will also break plugins.

We may need more useful API for the common pattern. For example: `env[meth, :tags] = yard_tags`, `env["NAMESPACE.GROUP.KEY"] = ...` (Rack style), ...

PR #1325, opened 2025-03-23
docs(README): fix the example that doesn't work

okuramasafumi:
```ruby
RDoc::RDoc.new.document(RDoc::Options.new)
```

fails with the following error:

```
/Users/okuramasafumi/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/rdoc-6.12.0/lib/rdoc/options.rb:522:in 'RDoc::Options#check_files': undefined method 'delete_if' for nil (NoMethodError)
```

I think this should be fixed, but for now modifying an example might be enough.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -53,11 +53,8 @@ To generate documentation programmatically:
   gem 'rdoc'
   require 'rdoc/rdoc'
 
-  options = RDoc::Options.new
-  # see RDoc::Options
-
   rdoc = RDoc::RDoc.new
-  rdoc.document options
+  rdoc.document([]) # This generates documentation for all files in the current directory
   # see RDoc::RDoc
 
 You can specify the target files for document generation with +.document+ file in the project root directory.


st0012:
If you think this should be fixed, which I agree, can you make a PR to address the root problem instead? It should be a rather straightforward one

tompng:
The failure is because some required setups for `RDoc::Options` are missing.
Since required things are missing, failing is not a problem. Of course the error message is not user friendly, though.

I think `rdoc.document(command_line_args)` is for command line execution and `rdoc.document(options)` is for generating document programmatically.
So README should use `rdoc.document(options)` with adding minimal option setups that succeeds without error.

okuramasafumi:
@st0012 @tompng Thank you for the quick response, I fixed the root issue in #1328 
That is still a workaround and not an ideal solution, but modifying/refactoring `RDoc::Options` class feels too much for this one problem.

README.rdoc
```diff
@@ -53,11 +53,8 @@ To generate documentation programmatically:
   gem 'rdoc'
   require 'rdoc/rdoc'
 
-  options = RDoc::Options.new
-  # see RDoc::Options
-
   rdoc = RDoc::RDoc.new
-  rdoc.document options
+  rdoc.document([]) # This generates documentation for all files in the current directory
```
tompng, line 57: How about this?
```
options = RDoc::Options.new
options.files = ['./a.rb', './b.rb']
options.setup_generator 'darkfish'
# see RDoc::Options for more 

rdoc = RDoc::RDoc.new
rdoc.document options
```

Some example of options setups are in `TestRDocRDoc#test_document`

tompng:
README fixed in #1337

PR #1326, opened 2025-03-23
Workaround `build` task issue in Ruby core CI

st0012:
The issue is described in https://github.com/rubygems/rubygems/issues/8477

The issue fails the workflow whenever RDoc's version is bumped as there will then be a mismatch between the target file name and the version that's actually built.

For example:

```
rdoc 6.13.0 built to pkg/rdoc-6.12.0.gem.
```

This commit works around the issue by manually building the gem and installing it locally so the latest installed version will match the version contained in the gemspec file.

### DIFF
--- a/.github/workflows/ruby-core.yml
+++ b/.github/workflows/ruby-core.yml
@@ -50,8 +50,13 @@ jobs:
         with:
           path: ruby/rdoc
       - name: Build RDoc locally
+        # The `build` task provided by `require 'bundler/gem_tasks'` seems to have a bug
+        # as described in https://github.com/rubygems/rubygems/issues/8477
+        # The manual `gem build` and `gem install` are used to work around this issue.
         run: |
           bundle install
+          gem build rdoc.gemspec
+          gem install rdoc-*.gem
           bundle exec rake build:local_ruby
         working-directory: ruby/rdoc
       - name: Generate Documentation with RDoc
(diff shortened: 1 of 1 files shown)


.github/workflows/ruby-core.yml
```diff
@@ -50,8 +50,13 @@ jobs:
         with:
           path: ruby/rdoc
       - name: Build RDoc locally
+        # The `build` task provided by `require 'bundler/gem_tasks'` seems to have a bug
+        # as described in https://github.com/rubygems/rubygems/issues/8477
+        # The manual `gem build` and `gem install` are used to work around this issue.
         run: |
           bundle install
+          gem build rdoc.gemspec
+          gem install rdoc-*.gem
```
kou, line 59: Do we need `bundle exec ...` here?

kou (approved):
+1

.github/workflows/ruby-core.yml
st0012: I don’t think so…? I don’t usually run `gem` with bundle as it usually doesn’t need any of the project’s dependencies. Is it a recommended practice?

kou: Ah, sorry. You're right.

Our `.gemspec` doesn't use external libraries. So we don't need `bundle exec` here.

kou, line 59: How about adding a note that the installed gem isn't used for generating Ruby documentation? (Should we see the linked issue?)

```suggestion
          # This gem isn't used for generating Ruby documentation.
          # This is just for fixing `pkg/rdoc-X.Y.Z.gem` path.
          gem install rdoc-*.gem
```

PR #1330, opened 2025-03-24
Allow customizing path prefix through options

st0012:
In https://github.com/ruby/rdoc/pull/1304, I removed the ability to set path prefix through patching Darkfish generator. But it turns out that it's used in `sdoc`.

See https://github.com/ruby/rdoc/pull/1304#issuecomment-2749251458

But the original implementation was brittle and confusing. So instead of just restoring it, I think allowing the customization through options is a better approach.

### DIFF
--- a/test/rdoc/test_rdoc_top_level.rb
+++ b/test/rdoc/test_rdoc_top_level.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative 'xref_test_case'
 
-class TestRDocTopLevel < XrefTestCase
+class RDocTopLevelTest < XrefTestCase
 
   def setup
     super
@@ -160,6 +160,13 @@ class TestRDocTopLevel < XrefTestCase
 
     other_level = @store.add_file 'path.other/level.rb'
     assert_equal 'path_other/level_rb.html', other_level.http_url
+  end
+
+  def test_path
+    assert_equal 'path/top_level_rb.html', @top_level.path
+
+    @options.file_path_prefix = 'file'
+    assert_equal 'file/path/top_level_rb.html', @top_level.path
   end
 
   def test_marshal_dump
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -630,7 +630,9 @@ class RDoc::ClassModule < RDoc::Context
   # Path to this class or module for use with HTML generator output.
 
   def path
-    http_url
+    prefix = options.class_module_path_prefix
+    return http_url unless prefix
+    File.join(prefix, http_url)
   end
 
   ##
--- a/lib/rdoc/code_object/top_level.rb
+++ b/lib/rdoc/code_object/top_level.rb
@@ -226,7 +226,9 @@ class RDoc::TopLevel < RDoc::Context
   # Path to this file for use with HTML generator output.
 
   def path
-    http_url
+    prefix = options.file_path_prefix
+    return http_url unless prefix
+    File.join(prefix, http_url)
   end
 
   def pretty_print q # :nodoc:
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -363,6 +363,16 @@ class RDoc::Options
   # Words to be ignored in autolink cross-references
   attr_accessor :autolink_excluded_words
 
+  ##
+  # The prefix to use for class and module page paths
+
+  attr_accessor :class_module_path_prefix
+
+  ##
+  # The prefix to use for file page paths
+
+  attr_accessor :file_path_prefix
+
   def initialize loaded_options = nil # :nodoc:
     init_ivars
     override loaded_options if loaded_options
@@ -417,6 +427,8 @@ class RDoc::Options
     @charset = @encoding.name
     @skip_tests = true
     @apply_default_exclude = true
+    @class_module_path_prefix = nil
+    @file_path_prefix = nil
   end
 
   def init_with map # :nodoc:
--- a/test/rdoc/test_rdoc_class_module.rb
+++ b/test/rdoc/test_rdoc_class_module.rb
@@ -1548,6 +1548,13 @@ class RDocClassModuleTest < XrefTestCase
     assert_equal ["A", "A::B", "A::B::C"], cm3.fully_qualified_nesting_namespaces
   end
 
+  def test_path
+    assert_equal 'C1.html', @c1.path
+
+    @options.class_module_path_prefix = 'class'
+    assert_equal 'class/C1.html', @c1.path
+  end
+
   class RDocClassModuleMixinsTest < XrefTestCase
     def setup
       super
--- a/test/rdoc/test_rdoc_options.rb
+++ b/test/rdoc/test_rdoc_options.rb
@@ -87,6 +87,8 @@ class TestRDocOptions < RDoc::TestCase
       'skip_tests'            => true,
       'apply_default_exclude' => true,
       'autolink_excluded_words' => [],
+      'class_module_path_prefix' => nil,
+      'file_path_prefix' => nil,
     }
 
     assert_equal expected, coder


st0012:
@jonathanhefner I think with these changes `sdoc` can adopt the branch rather easily?

(Ignore Gemfile diffs)

```diff
diff --git a/Gemfile b/Gemfile
index 0802717..8b0e609 100644
--- a/Gemfile
+++ b/Gemfile
@@ -8,9 +8,11 @@ gem "hoe"
 gem "minitest"
 gem "puma"
 
-if ENV["rdoc"] == "master"
-  gem "rdoc", :github => "ruby/rdoc"
-end
+# if ENV["rdoc"] == "master"
+#   gem "rdoc", :github => "ruby/rdoc"
+# end
 
 gem "importmap-rails"
 gem "railties"
+
+gem "rdoc", :github => "ruby/rdoc", branch: "allow-customizing-path-prefix-through-options"
diff --git a/lib/sdoc/generator.rb b/lib/sdoc/generator.rb
index 023d066..884cb53 100644
--- a/lib/sdoc/generator.rb
+++ b/lib/sdoc/generator.rb
@@ -73,6 +73,9 @@ class RDoc::Generator::SDoc
     end
     @options.pipe = true
 
+    @options.class_module_path_prefix = "classes"
+    @options.file_path_prefix = "files"
+
     @original_dir = Pathname.pwd
     @template_dir = Pathname(options.template_dir)
     @output_dir = Pathname(@options.op_dir).expand_path(options.root)
diff --git a/lib/sdoc/rdoc_monkey_patches.rb b/lib/sdoc/rdoc_monkey_patches.rb
index fac1297..c8ab744 100644
--- a/lib/sdoc/rdoc_monkey_patches.rb
+++ b/lib/sdoc/rdoc_monkey_patches.rb
@@ -1,19 +1,5 @@
 require "rdoc"
 
-RDoc::TopLevel.prepend(Module.new do
-  def path
-    File.join("files", super)
-  end
-end)
-
-
-RDoc::ClassModule.prepend(Module.new do
-  def path
-    File.join("classes", super)
-  end
-end)
-
-
 RDoc::TopLevel.prepend(Module.new do
   attr_writer :path
 
diff --git a/spec/spec_helper.rb b/spec/spec_helper.rb
index ada0ca7..a7acfb1 100644
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -28,7 +28,11 @@ def rdoc_top_level_for(ruby_code)
   # foolproof way to initialize it is by simply running it with a dummy file.
   $rdoc_for_specs ||= rdoc_dry_run("--files", __FILE__)
 
-  $rdoc_for_specs.store = RDoc::Store.new(RDoc::Options.new)
+  options = RDoc::Options.new
+  options.class_module_path_prefix = "classes"
+  options.file_path_prefix = "files"
+
+  $rdoc_for_specs.store = RDoc::Store.new(options)
 
   Dir.mktmpdir do |dir|
     path = "#{dir}/ruby_code.rb"
```

test/rdoc/test_rdoc_top_level.rb
```diff
@@ -162,6 +162,14 @@ def test_http_url
     assert_equal 'path_other/level_rb.html', other_level.http_url
   end
 
+  def test_path
+    assert_equal 'path/top_level_rb.html', @top_level.path
+
+    # Note: @options != @store.options
```
kou, line 168: Oh... Can we remove this comment by https://github.com/ruby/rdoc/pull/1332 ?

kou (approved):
+1

test/rdoc/test_rdoc_top_level.rb
st0012: I'll rebase and test once #1332 is merged 👍

kou: Merged!

st0012:
I'm open to both 6.13.1 and 6.14 but prefer the latter as it's technically a new feature.

Can you give me more context on how sdoc manages:
- Its dependency versions. Like how do you decide if it's ok to require `rdoc >= 6.14`
- The Ruby versions it needs to support

jonathanhefner:
> I'm open to both 6.13.1 and 6.14 but prefer the latter as it's technically a new feature.

Normally, I would agree that 6.14 is more appropriate.  However, I think this feature arguably should have been a part of 6.13.0, so it seems reasonable to include it in 6.13.1.  Also, 6.13.0 already deviated semantic versioning by including breaking changes; adding a feature in 6.13.1 isn't much more of a deviation.

> Can you give me more context on how sdoc manages:
>
> * Its dependency versions. Like how do you decide if it's ok to require `rdoc >= 6.14`
>
> * The Ruby versions it needs to support

To be honest, I'm not sure if there is a policy for those things.  I do not know what all is downstream of SDoc, so I am just trying to keep things in working order.

st0012:
I'm asking because we could focus on upstreaming/reducing sdoc's rdoc patches next, which I think will be beneficial to both sides. But if sdoc still needs to keep the patches around due to not being able to bump the required rdoc version, then it just won't be as effective.

jonathanhefner:
> I'm asking because we could focus on upstreaming/reducing sdoc's rdoc patches next, which I think will be beneficial to both sides. But if sdoc still needs to keep the patches around due to not being able to bump the required rdoc version, then it just won't be as effective.

Ah, I see.  In that case, I agree it would be beneficial to upstream some of the patches.  All of the monkey patches apply idempotent changes, so even if SDoc can't drop them now, it won't be impacted by RDoc adopting them.  And then SDoc could drop them at some point in the future.

List of patches:

* [Makes `RDoc::TopLevel#path` overwritable](https://github.com/rails/sdoc/blob/3fcc2bb3d93f546adcfef996e7ce99e14ed6a191/lib/sdoc/rdoc_monkey_patches.rb#L17-L23).
* [Prefixes `RDoc::Constant#aref` with `"constant-"`](https://github.com/rails/sdoc/blob/3fcc2bb3d93f546adcfef996e7ce99e14ed6a191/lib/sdoc/rdoc_monkey_patches.rb#L26-L38) so that `RDoc::Constant#path` is consistent with e.g. `RDoc::MethodAttr#path` for permalinks.
  * The patched `RDoc::Constant#path` method strips `/#.+/` before appending `"##{aref}"`, so it won't be impacted if RDoc adopts similar behavior.
* [Removes extraneous whitespace inside method signature parens](https://github.com/rails/sdoc/blob/3fcc2bb3d93f546adcfef996e7ce99e14ed6a191/lib/sdoc/rdoc_monkey_patches.rb#L41-L45) (see rails/sdoc@43384df0d6478c0277fba2d5b60c07bfe77e62b3 commit message for an example).
  * The regex `sub` calls are idempotent.
* [Prevents some unintentional auto-links and adds code styling to some explicit ref links](https://github.com/rails/sdoc/blob/3fcc2bb3d93f546adcfef996e7ce99e14ed6a191/lib/sdoc/rdoc_monkey_patches.rb#L48-L60).
  * The patch logic runs before `super` is called, so it won't be impacted if RDoc adopts similar behavior.
* [Fixes an RDoc bug](https://github.com/rails/sdoc/blob/3fcc2bb3d93f546adcfef996e7ce99e14ed6a191/lib/sdoc/rdoc_monkey_patches.rb#L63-L76) (see rails/sdoc@369f964b8066d40dffa57a0c54132d923f3cb2f9 commit message for an example).
  * The patch logic runs around the `super` calls, so it won't be impacted if RDoc adopts similar behavior.

st0012:
@jonathanhefner Thank you for the detailed list. I've created https://github.com/ruby/rdoc/issues/1333 to remind ourselves to act on them later.
In the meantime, I'll merge this and cut `6.13.1`.

st0012:
@jonathanhefner `v6.13.1` is out with this PR.

jonathanhefner:
Thank you, @st0012! :heart:

PR #1337, opened 2025-04-02
Fix RDoc::Options example in README

tompng:
Add minimal option setup(options.files and options.generator) to make the example code work.

https://github.com/ruby/rdoc/pull/1325#discussion_r2009419651


kou (approved):
+1

PR #1341, opened 2025-04-18
Move default RDoc headings to a constant

vinistock:
Instead of setting the heading values one by one, we can simply initialize the hash straight away. And since the default values never change, we can use a constant to define them and dup in case they have to be modified by the instance.

### DIFF
--- a/lib/rdoc/markup/to_rdoc.rb
+++ b/lib/rdoc/markup/to_rdoc.rb
@@ -3,6 +3,16 @@
 # Outputs RDoc markup as RDoc markup! (mostly)
 
 class RDoc::Markup::ToRdoc < RDoc::Markup::Formatter
+  DEFAULT_HEADINGS = {
+      1 => ['= ',      ''],
+      2 => ['== ',     ''],
+      3 => ['=== ',    ''],
+      4 => ['==== ',   ''],
+      5 => ['===== ',  ''],
+      6 => ['====== ', '']
+  }
+  DEFAULT_HEADINGS.default = []
+  DEFAULT_HEADINGS.freeze
 
   ##
   # Current indent amount for output in characters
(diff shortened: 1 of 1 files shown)


lib/rdoc/markup/to_rdoc.rb
```diff
@@ -3,6 +3,14 @@
...
 
 class RDoc::Markup::ToRdoc < RDoc::Markup::Formatter
+  DEFAULT_HEADINGS = {
+      1 => ['= ',      ''],
+      2 => ['== ',     ''],
+      3 => ['=== ',    ''],
+      4 => ['==== ',   ''],
+      5 => ['===== ',  ''],
+      6 => ['====== ', '']
+  }.tap { |h| h.default = [] }.freeze
```
kou, line 13: JFYI: I like the following style (no `tap`, no `freeze`) but it's just my preference. It's not required for merging.

```suggestion
  }
  DEFAULT_HEADINGS.default = []
```

kou (approved):
+1

lib/rdoc/markup/to_rdoc.rb
st0012: I think freezing a mutable structure (e.g. array, hash...etc.) constant is still a recommended practice if we don't expect it to change?

kou: I know some people recommend the style but I think that I don't like the style because:

* Many `.freeze` codes are (a bit) ugly
* Constant (`XXX = {}`) already shows our intention (we don't expect it to change)

(NOTE: I don't object the style in RDoc.)

tompng (approved):

lib/rdoc/markup/to_rdoc.rb
vinistock: I tried meeting everybody half way and split the default and freezing into separate lines.

PR #1342, opened 2025-04-18
Enable `Style/MethodDefParentheses` rule

vinistock:
The RDoc codebase has very few conventions. I think it would be nice to standardize the style of the code a bit more. I propose we start requiring parenthesis in method definitions, which in my opinion makes the code more readable.

This PR just changes the configurations and then runs `bundle exec rubocop -a`. I also excluded generated files from linting.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,6 +2,14 @@ AllCops:
   TargetRubyVersion: 3.0
   DisabledByDefault: true
   SuggestExtensions: false
+  Exclude:
+    # Exclude files that are auto generated
+    - "lib/rdoc/rd/block_parser.rb"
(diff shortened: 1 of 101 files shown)


.rubocop.yml
```diff
@@ -2,6 +2,12 @@ AllCops:
   TargetRubyVersion: 3.0
   DisabledByDefault: true
   SuggestExtensions: false
+  Exclude:
```
kou, line 5: How about adding a comment why we exclude them?

kou (approved):
+1

.rubocop.yml
vinistock: Good point. Added some comments.

PR #1348, opened 2025-04-25
Fix image tag with external source

ybiquitous:
When `rdoc-image` contains an external source, such as `https://example.com/path/to/image.svg`, the rendered `` tag is broken. This PR fixes the bug.

For example, I found the problem on the IRB website, where a Markdown image is present as a source.

Rendered HTML on the website <https://ruby.github.io/irb/>:

```html
<a href="https://badge.fury.io/rb/irb"><img src="https" alt="//badge.fury.io/rb/irb.svg:Gem Version"></a>
```

Screenshot:

image

Markdown source: <https://github.com/ruby/irb/blob/6349b03f64f173a9d31bc9042430e1134f3e1689/doc/Index.md?plain=1#L3>

```markdown
[![Gem Version](https://badge.fury.io/rb/irb.svg)](https://badge.fury.io/rb/irb)
```

Probably, this bug was introduced in the commit 894b2f1cd147a760449a52b621af7bf206452127, i.e., <https://github.com/ruby/rdoc/releases/tag/v6.13.0>.

### DIFF
--- a/lib/rdoc/markup/to_html.rb
+++ b/lib/rdoc/markup/to_html.rb
@@ -98,7 +98,7 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
 
       gen_url CGI.escapeHTML(url), CGI.escapeHTML(text)
     when /^rdoc-image:/
-     # Split the string after "rdoc-image:" into url and alt.
+      # Split the string after "rdoc-image:" into url and alt.
       #   "path/to/image.jpg:alt text" => ["path/to/image.jpg", "alt text"]
       #   "http://example.com/path/to/image.jpg:alt text" => ["http://example.com/path/to/image.jpg", "alt text"]
       url, alt = $'.split(/:(?!\/)/, 2)
--- a/lib/rdoc/markup/to_html.rb
+++ b/lib/rdoc/markup/to_html.rb
@@ -98,7 +98,11 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
 
       gen_url CGI.escapeHTML(url), CGI.escapeHTML(text)
     when /^rdoc-image:/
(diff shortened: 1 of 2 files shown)


lib/rdoc/markup/to_html.rb
```diff
@@ -98,7 +98,11 @@ def handle_RDOCLINK(url) # :nodoc:
 
       gen_url CGI.escapeHTML(url), CGI.escapeHTML(text)
     when /^rdoc-image:/
-      url, alt = $'.split(":", 2) # Split the string after "rdoc-image:" into url and alt
+      # Split the string after "rdoc-image:" into url and alt
+      url_alt = $'
+      split_pos = url_alt.index(%r{:[^/]})
+      url = split_pos ? url_alt[0, split_pos] : url_alt
+      alt = split_pos ? url_alt[split_pos + 1, url_alt.size] : nil
```
ybiquitous, line 105: It looks a bit messy, and there might be a better code here... Welcome any suggestions! 😃

kou, line 105: Positive lookahead?

```suggestion
      # Split the string after "rdoc-image:" into url and alt.
      #   "path/to/image.jpg:alt text" => ["path/to/image.jpg", "alt text"]
      #   "http://example.com/path/to/image.jpg:alt text" => ["http://example.com/path/to/image.jpg", "alt text"]
      url, alt = $'.split(/(?=:[^\/]):/, 2)
```

ybiquitous: Thank you. Negative lookahead makes the code cleaner! 👍🏼

ybiquitous: Improved via https://github.com/ruby/rdoc/pull/1348#discussion_r2060084903

st0012 (approved):
Thanks for the fix!

PR #1349, opened 2025-04-28
Polish link underline in Darkfish CSS

ybiquitous:
This aims to make sure an underline in a link doesn't overlap with underscores in a method name.

For example, method names on the following page seem a bit hard to read because underscores in a method name are hidden in an underline:
image
https://ruby.github.io/rdoc/table_of_contents.html#methods

This PR improves it a bit:

2px:
image

3px:
image

Note: The [`text-underline-offset`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-offset) CSS property is widely available now.

### DIFF
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -97,7 +97,7 @@ a {
   color: var(--link-color);
   transition: color 0.3s ease;
   text-decoration: underline;
-  text-underline-offset: 2px; /* Make sure it doesn't overlap with underscores in a method name. */
+  text-underline-offset: 0.2em; /* Make sure it doesn't overlap with underscores in a method name. */
 }
 
 a:hover {
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -96,6 +96,8 @@ main .anchor-link:target {
 a {
   color: var(--link-color);
   transition: color 0.3s ease;
+  text-decoration: underline;
+  text-underline-offset: 2px; /* Make sure it doesn't overlap with underscores in a method name. */
 }
 
 a:hover {


lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -96,6 +96,8 @@ main .anchor-link:target {
 a {
   color: var(--link-color);
   transition: color 0.3s ease;
+  text-decoration: underline;
+  text-underline-offset: 2px; /* Make sure it doesn't overlap with underscores in a method name. */
```
ybiquitous, line 100: [note] Why not more spaces like 3px? If it becomes 3px, an inline code with a link looks like (this example is on the LEGAL page): I'm a bit concerned that the underline sticks to the bottom line of the inline code box.
image

In the case of 2px:
image

But if people are in favor of 3px or more, I'm glad to change it. 😃

kou: I prefer 3px to 2px but I'm not good at visual design...

Others should decide it...

kou (approved):
+1

lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -96,6 +96,8 @@ main .anchor-link:target {
 a {
   color: var(--link-color);
   transition: color 0.3s ease;
+  text-decoration: underline;
```
kou, line 99: Do we need this?

ybiquitous: `text-decoration: underline` isn't necessarily required, but most browsers have it in their default user agent stylesheets. This is an example of Chrome:
image

So, I've added it to prevent our confusion, e.g., "Why is `text-underline-offset` specified, regardless `text-decoration` is not?"

If we want to respect browsers that don't have `text-decoration: underline` by default, we can remove it. 👌🏼 In this case, it might be helpful to add a code comment that describes the reason for not having `text-decoration`.

vinistock: I have a question. We're adding this underline offset to all links, but not every link is a method name.

Will this look okay for other links that aren't method names? Should we use a more specific selector to match just method names or is the visual consistency better?

lib/rdoc/generator/template/darkfish/css/rdoc.css
vinistock: Not a design expert either, but I find the 3px version better.

lib/rdoc/generator/template/darkfish/css/rdoc.css
tompng: I think underline offset is fine for every links.
Link text could be an URL or a filename that contains underscore. Appearance of these links also gets better.

Another example, GitHub's "Convert to draft" link and URL in pull request description text also has `text-underline-offset: .2rem; // about 3.2px`.

ybiquitous: Thanks for the feedback. For your information, it's easy to restrict only links that have `<code>`, but it might be harder for other links, e.g.

```html
<a href="..."><code>::check_modeline</code></a>

<a href="...">::check_modeline</a>
```

```css
/* This is applied into the first `<a>` only, not the second one. */
a:has(code) { text-underline-offset: 3px; }
```

ybiquitous: As GitHub does, we can also use the `rem` unit instead of `px`. `rem` has a relative value from the root font size, so it might be better to maintain.

Considering the favors on <https://github.com/ruby/rdoc/pull/1349#discussion_r2062781030>, I'll adjust the value if no other concerns:
- Use `rem` instead of `px`
- Increase the offset value to keep more space.

ybiquitous: I've increased the `text-underline-offset` value from `2px` to `0.2em` via f7304352. See the following screenshots to see the new value:

- image
- image

Indeed, `0.2em` is interpreted as `3.2px` since the `<main>` and `<nav>` have `font-size: 16px` (`16px * 0.2 == 3.2px`):

- https://github.com/ruby/rdoc/blob/8d07737ed003f512ff979c9d37f2e49c56c9db32/lib/rdoc/generator/template/darkfish/css/rdoc.css#L292-L298
- https://github.com/ruby/rdoc/blob/8d07737ed003f512ff979c9d37f2e49c56c9db32/lib/rdoc/generator/template/darkfish/css/rdoc.css#L151-L153

The reason that I chose the `em` unit instead of `rem` is because the Darkfish CSS doesn't specify `font-size` to `:root` or `html` elements (rem stands for "root em"). If this CSS has `font-size` for `:root` in the future, we may need to reconsider the `em` value at the same time.

ybiquitous: @tompng @vinistock I've pushed the commit f7304352. What do you think?

tompng (approved):

st0012 (approved):

lib/rdoc/generator/template/darkfish/css/rdoc.css
vinistock: Looks good!

vinistock (approved):

PR #1350, opened 2025-04-28
Fix link to CONTRIBUTING on README

ybiquitous:
I probably think it should be a link to the CONTRIBUTING page, instead of an inline code `CONTRIBUTING`. See https://ruby.github.io/rdoc/#label-Bugs
image

If applying this PR, it becomes:
image


kou (approved):
+1

PR #1354, opened 2025-05-01
Add support for canonical URL link tag

p8:
Currently, search engines don't know which version to show for API documentation. For example, searching for "ruby string" in Google will show the documentation for Ruby 2.0 `https://docs.ruby-lang.org/en/2.0.0/String.html` as the top result (for docs.ruby-lang.org).

The canonical URL link tag will allow us to set the preferred version:

> The canonical URL link tag defines the preferred URL for the current
> document, which helps search engines reduce duplicate content.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel#canonical

For example, for the official Ruby documentation we can add the following to `.rdoc_options`.

```yaml
canonical_root: https://docs.ruby-lang.org/en/master
```

This will add the canonical URL link tag to relevant pages:

```html
<link rel="canonical" href="https://docs.ruby-lang.org/en/master/String.html">
```

### DIFF
--- a/lib/rdoc/generator/template/darkfish/_head.rhtml
+++ b/lib/rdoc/generator/template/darkfish/_head.rhtml
@@ -25,6 +25,11 @@
   <%- end -%>
 <%- end -%>
 
+<%- if canonical_url = @options.canonical_root -%>
+<% canonical_url = current.canonical_url if defined?(current) %>
+<link rel="canonical" href="<%= canonical_url %>">
+<%- end -%>
+
 <script type="text/javascript">
   var rdoc_rel_prefix = "<%= h asset_rel_prefix %>/";
   var index_rel_prefix = "<%= h rel_prefix %>/";
--- a/lib/rdoc/generator/markup.rb
+++ b/lib/rdoc/generator/markup.rb
@@ -55,6 +55,18 @@ module RDoc::Generator::Markup
     end
   end
 
+  ##
+  # The preferred URL for this object.
+
+  def canonical_url
+    options = @store.options
+    if path
+      File.join(options.canonical_root, path.to_s)
+    else
+      options.canonical_root
+    end
+  end
+
 end
 
 class RDoc::CodeObject
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -373,6 +373,11 @@ class RDoc::Options
 
   attr_accessor :file_path_prefix
 
+  ##
+  # The preferred root URL for the documentation
+
+  attr_accessor :canonical_root
+
   def initialize(loaded_options = nil) # :nodoc:
     init_ivars
     override loaded_options if loaded_options
@@ -429,6 +434,7 @@ class RDoc::Options
     @apply_default_exclude = true
     @class_module_path_prefix = nil
     @file_path_prefix = nil
+    @canonical_root = nil
   end
 
   def init_with(map) # :nodoc:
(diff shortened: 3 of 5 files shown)


lib/rdoc/generator/template/darkfish/_head.rhtml
```diff
@@ -25,6 +25,12 @@
   <%- end -%>
 <%- end -%>
 
+<%- if @options.canonical_root -%>
+  <% context = klass if defined?(klass) %>
+  <% context = file if defined?(file) %>
+  <link rel="canonical" href="<%= canonical_url(context) %>">
```
st0012, line 31: Can this be included in the if statements above? Then we don't need the context conditions separately.

test/rdoc/test_rdoc_generator_darkfish.rb
```diff
@@ -514,6 +514,29 @@ def test_meta_tags_for_empty_document
...
   end
 
+  def test_canonical_url_for_index
+    @options.canonical_root = "https://docs.ruby-lang.org/en/master/"
+    @g.generate
+
+    content = File.binread("index.html")
+
+    assert_include(content, '<link rel="canonical" href="https://docs.ruby-lang.org/en/master/">')
+
```
st0012, line 524:
```suggestion
```

test/rdoc/test_rdoc_generator_darkfish.rb
```diff
@@ -514,6 +514,29 @@ def test_meta_tags_for_empty_document
...
+    top_level.add_class(@klass.class, @klass.name)
+    inner = @klass.add_class(RDoc::NormalClass, "Inner")
+
+    @options.canonical_root = "https://docs.ruby-lang.org/en/master/"
+    @g.generate
+
+    content = File.binread("Klass/Inner.html")
+
+    assert_include(content, '<link rel="canonical" href="https://docs.ruby-lang.org/en/master/Klass/Inner.html">')
+  end
```
st0012, line 538: Can we have another test for file pages?

lib/rdoc/generator/darkfish.rb
```diff
@@ -729,6 +729,10 @@ def excerpt(comment)
     extracted_text[0...150].tr_s("\n", " ").squeeze(" ")
   end
 
+  def canonical_url(context)
+    File.join(@options.canonical_root, context&.path.to_s)
```
st0012, line 733: I prefer making `canonical_url` taking the path already instead of a nilable context. It'll likely be easier to maintain and debug.

st0012 (changes requested):
Love this idea ❤️  Just some suggestions.

lib/rdoc/generator/darkfish.rb
p8: I've changed the implementation to take the path instead.

lib/rdoc/generator/template/darkfish/_head.rhtml
p8: I've changed the implementation to use `current` instead of `klass` and `file` to simplify.
If I'd use the `if` statement I would need to add an `else` as well, resulting in adding the tag 4 times.

test/rdoc/test_rdoc_generator_darkfish.rb
p8: Done!

kou (approved):
+1

test/rdoc/test_rdoc_generator_darkfish.rb
st0012: Hmm sorry I don't see it?

p8:
I’ve changed the implementation to have less logic in the template and moved it RDoc::Markup. This will allow other templates to use it as well.

st0012 (approved):
This is great, thank you!

st0012:
@p8 can you rebase and push again? I'll merge it then, thx

p8:
Thanks @st0012 ! I've rebased to master.

st0012:
I've applied this to Ruby's [master doc](https://github.com/ruby/ruby/pull/13410) and [3.4 doc](https://github.com/ruby/ruby/pull/13410).

p8:
@st0012 Thanks!

PR #1359, opened 2025-05-04
Rename test classes from Test* to *Test

st0012:
Since class names are usually nouns, test classes should be named like `ClassNameTest` instead of `TestClassName`.

This convention is now commonly used in projects (e.g. Rails) and tools like Ruby LSP use it to check if a class is a test class. (From tools' perspective, checking `Test` as a postfix is more accurate than checking `Test` as a prefix as `TestHelper`, `TestCase`...etc. are not test classes.)

I've previously renamed some classes this way. This commit renames all remaining classes to follow the convention.


simi:
It will make sense to rename files as well IMHO to keep it consistent.

kou (approved):
+1

st0012:
I've renamed the test files too. It'd require almost all opened PRs to rebase, but it should be rather easy to do.

st0012:
Windows head CI are failing for unrelated reasons. Merging now

PR #1361, opened 2025-05-05
Set language in HTML

p8:
Setting the language helps with search engines, screen readers and text rendering. The language defaults to `en` as that is probably most common (it's also the default for navigation in Darkfish).

Setting the `locale` option will override the language to the locale.

### DIFF
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -595,7 +595,7 @@ class RDoc::Generator::Darkfish
     <<-TEMPLATE
 <!DOCTYPE html>
 
-<html>
+<html lang="#{@options.locale&.name || 'en'}">
 <head>
 #{head_file.read}
 
--- a/test/rdoc/rdoc_generator_darkfish_test.rb
+++ b/test/rdoc/rdoc_generator_darkfish_test.rb
@@ -388,6 +388,21 @@ class RDocGeneratorDarkfishTest < RDoc::TestCase
     assert_include File.binread('index.html'), %Q[href="./#{base}"]
   end
 
+  def test_html_lang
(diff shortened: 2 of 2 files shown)


test/rdoc/test_rdoc_generator_darkfish.rb
```diff
@@ -388,6 +388,21 @@ def test_template_stylesheets
...
 
+  def test_html_lang
+    @g.generate
+
+    content = File.binread("index.html")
+    assert_include(content, '<html lang="en">')
+  end
+
+  def test_html_lang_from_locale
+    @options.locale = 'ja'
```
vinistock, line 399: I think this should be `locale_name`, no? Based on my understanding, `@locale` is only set [here](https://github.com/ruby/rdoc/blob/fdaa5a605a93cb274556587b9152f94dff2b2fb1/lib/rdoc/options.rb#L639) based on the name.

In fact, I don't know why `Options` is exposing an `attr_accessor`, you shouldn't be able to override this.

Did you verify if this works with manual testing? I think this may be bypassing the way options are initialized and parsed because we're setting `locale` directly to a string, but I think it should be a [Locale object](https://github.com/ruby/rdoc/blob/master/lib/rdoc/i18n/locale.rb) outside of tests.

p8: Ah sorry, you are correct. Manual testing resulted in an inspect of the Locale object.
I'm setting the locale to the Locale object for now, as the `locale_name` doesn't have an accessor.

kou (approved):
+1

vinistock (approved):

p8:
Thanks @vinistock !

PR #1363, opened 2025-05-07
Remove border-left from method details

JuanitoFatas:
It reads better without the border left 😅 E.g. [Process#kill](https://docs.ruby-lang.org/en/master/Process.html#method-c-kill)

|Before|After|
|------|------|
|CleanShot 2025-05-07 at 22 08 10@2x|CleanShot 2025-05-07 at 22 08 20@2x|


lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -581,7 +581,6 @@ main .method-detail {
 
 main .method-detail:target {
   margin-left: -10px;
-  border-left: 10px solid var(--border-color);
```
tompng, line 584: The value `margin-left: -10px` should be `"-#{border_left_width}px"`. If border-left is `0`, margin-left should also be `0`.

However, this border-left is a highlight of the focused method. Could you suggest a better highlight instead of completely removing it?

JuanitoFatas: Ah, that makes sense now. Given that the anchor link reliably brings the method to the top of the viewport, I think we could remove both `border-left` and `margin-left`. 

The method name is easy to spot without extra highlighting. It will look like this if we remove both the border and margin:

https://github.com/user-attachments/assets/6c44c3da-97bd-4235-aa7e-7c981d753794

tompng: This is one of the features of RDoc. I don't want to remove it as possible if it is fixable.

This border-left exists in other links to heading tags, not only to method links.
heading_borderleft_style


Differences are:
Border-left to method links are applied to the whole `div.method-detail` content while border-left to other links are only applied to heading tags.

How about changing the border-left target to heading of the method detail?
It will be more consistent between method links and other links to heading tags.

```diff
- main .method-detail:target {
+ main .method-detail:target .method-heading {
```

p8:
In SDoc we show a small highlight next to the method as well:
image

lib/rdoc/generator/template/darkfish/css/rdoc.css
JuanitoFatas: I like this! Pushed up [212b284](https://github.com/ruby/rdoc/pull/1363/commits/212b284363ab79ff4f447075a015587d88e10c52) and updated the screenshot in PR body.

tompng: Can you leave the margin-left and border style as is?

There is a layout shift. On the other hand, `h2:target` and the original `.method-detail:target` doesn't have this layout shift.
layout_shift

Of course we can discuss adding a padding to `h2:target` and/or `.method-detail:target .method-heading`, with or without this layout shift. 
But it's easy to merge if the point of this pull request is clear.

JuanitoFatas: Sure thing. Sorry for all the back and forth. Updated.

lib/rdoc/generator/template/darkfish/css/rdoc.css
```diff
@@ -579,8 +579,8 @@ main .method-detail {
   margin-bottom: 2.5em;
 }
 
-main .method-detail:target {
-  margin-left: -10px;
+main .method-detail:target .method-heading {
+  margin-left: 10px;
```
tompng, line 583:
```suggestion
  margin-left: -10px;
```
(to revert the change)

PR #1366, opened 2025-05-12
Load `cgi/escape` instead of `cgi/util`

Earlopain:
`cgi/util` will warn on Ruby 3.5. `cgi/escape` is available since Ruby 2.4, but to use CGI.unescape* methods on Ruby 3.4 and earlier, `cgi/util` must still be required.

There's only one place in rdoc that uses such a method, so only one place needs to handle earlier rubies.

https://bugs.ruby-lang.org/issues/21258

### DIFF
--- a/lib/rdoc/markup/to_html.rb
+++ b/lib/rdoc/markup/to_html.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
-require 'cgi/util'
+require 'cgi/escape'
+# For CGI.unescape on earlier rubies
+require 'cgi/util' if RUBY_VERSION < '3.5'
 
 ##
 # Outputs RDoc markup as HTML.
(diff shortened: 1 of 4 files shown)


lib/rdoc/markup/to_html.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
-require 'cgi/util'
+require 'cgi/escape'
+# For CGI.unescape on earlier rubies
+require 'cgi/util' if RUBY_VERSION < '3.5'
```
tompng, line 4:
```ruby
ruby -rcgi/escape -e "CGI.unescape ''"
# Ruby 3.5: OK
# Ruby 3.4: uninitialized class variable @@accept_charset in #<Class:CGI> (NameError)
ruby -rcgi/escape -rcgi/util -e "CGI.unescape ''"
# Ruby 3.4: OK
```

tompng (approved):
Thank you :+1:

PR #1371, opened 2025-05-19
Bump version to 6.14.0

st0012:
https://github.com/ruby/rdoc/compare/v6.13.1...master

We can then bump `ruby/ruby`'s RDoc to [set canonical url](https://github.com/ruby/rdoc/pull/1354).

### DIFF
--- a/lib/rdoc/version.rb
+++ b/lib/rdoc/version.rb
@@ -5,6 +5,6 @@ module RDoc
   ##
   # RDoc version you are using
 
-  VERSION = '6.13.1'
+  VERSION = '6.14.0'
 
 end


tompng (approved):
The ci failure is related to https://github.com/ruby/rdoc/pull/1366

Passed test (head windows-latest) is using `3.5.0dev (2025-05-17 master aa0f689bf4)`
failed test (mswin, windows-latest) is using `3.5.0dev (2025-02-27T07:55:54Z master 2dff416ff2)`
The failure is because the version is old and does not contain recent CGI updates. I think we can ignore it.

PR #1374, opened 2025-05-31
Add version selector

p8:
To make it easier to switch to documentation of other versions, add a select box to select other versions.

For example, the versions can be defined in `.rdoc_options`:
```yaml
versions:
  latest: https://docs.ruby-lang.org/en/
  master: https://docs.ruby-lang.org/en/master/
  3.4: https://docs.ruby-lang.org/en/3.4/
```

image

This would allow `https://docs.ruby-lang.org/en/` to just show RDoc documentation for the last released version instead of a list of documentation version.

### DIFF
--- a/lib/rdoc/generator/markup.rb
+++ b/lib/rdoc/generator/markup.rb
@@ -67,6 +67,22 @@ module RDoc::Generator::Markup
     end
   end
 
+  ##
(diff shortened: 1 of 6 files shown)


st0012:
Thanks for the PR!
Can this feature benefit https://api.rubyonrails.org/ too?

PR #1376, opened 2025-06-12
Allow rdoc-ref to link to non-text files

jeremyevans:
It appears the current behavior of disallowing links to non-text files is deliberate, as it was explicitly added in 3628e196cbbac4767f5b845c5b795c635a671b04.  While the commit message explains the change, it doesn't provide a justification for excluding non-text TopLevels. The issue mentioned in the commit message is also unrelated to the change.

It's just as useful to link to a non-text file as it is to link to a text file, so I think it should be allowed.  I was surprised when it didn't work. I want to use this feature in tilt's documentation.

If there is a reason to disallow it by default, I think this limitation should be documented.  I also I think we should add an option to allow it in that case.

### DIFF
--- a/lib/rdoc/store.rb
+++ b/lib/rdoc/store.rb
@@ -723,7 +723,7 @@ class RDoc::Store
   # Returns the RDoc::TopLevel that is a text file and has the given +name+
 
   def page(name)
-    @text_files_hash.each_value.find do |file|
+    @files_hash.each_value.find do |file|
       file.page_name == name or file.base_name == name
     end
   end
--- a/lib/rdoc/code_object/top_level.rb
+++ b/lib/rdoc/code_object/top_level.rb
@@ -115,14 +115,6 @@ class RDoc::TopLevel < RDoc::Context
   alias name base_name
 
   ##
-  # Only a TopLevel that contains text file) will be displayed.  See also
-  # RDoc::CodeObject#display?
-
-  def display?
-    text? and super
-  end
-
-  ##
   # See RDoc::TopLevel::find_class_or_module
   #--
   # TODO Why do we search through all classes/modules found, not just the
--- a/lib/rdoc/cross_reference.rb
+++ b/lib/rdoc/cross_reference.rb
@@ -200,7 +200,7 @@ class RDoc::CrossReference
     ref = resolve_method name unless ref
 
     # Try a page name
-    ref = @store.page name if not ref and name =~ /^[\w.]+$/
+    ref = @store.page name if not ref and name =~ /^[\w.\/]+$/
 
     ref = nil if RDoc::Alias === ref # external alias, can't link to it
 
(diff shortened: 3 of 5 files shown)


st0012:
I agree with the goal but I think we need a bit more changes to achieve that.

To test the PR, I added `{Link}[rdoc-ref:lib/rdoc/cross_reference.rb]` to `readme.rdoc`. But clicking it doesn't bring me to the right page. The reason seems to be because `cross_reference.rb`'s uri path is `/RDoc/CrossReference.html` but it's not successfully generated in the html link.
And can you add some test cases to `test/rdoc/rdoc_markup_to_html_crossref_test.rb` as well? Thanks

jeremyevans:
> I agree with the goal but I think we need a bit more changes to achieve that.
> 
> To test the PR, I added `{Link}[rdoc-ref:lib/rdoc/cross_reference.rb]` to `readme.rdoc`. But clicking it doesn't bring me to the right page. The reason seems to be because `cross_reference.rb`'s uri path is `/RDoc/CrossReference.html` but it's not successfully generated in the html link. 

The link is correct, the issue is that darkfish doesn't generate the page.  Added a commit to fix darkfish, and updated the darkfish tests to test that it is generated. When I was testing earlier versions of this, I was using hanna (as I do for all my projects), which has always generated doc files for non-text files.

> And can you add some test cases to `test/rdoc/rdoc_markup_to_html_crossref_test.rb` as well? Thanks

The first commit already tests resolution of non-text files:

```
  assert_ref @top_level, 'xref_data.rb'
```

Can you let me know which additional tests you would like?

lib/rdoc/store.rb
```diff
@@ -723,7 +723,7 @@ def modules_hash
   # Returns the RDoc::TopLevel that is a text file and has the given +name+
 
   def page(name)
-    @text_files_hash.each_value.find do |file|
+    @files_hash.each_value.find do |file|
```
st0012, line 726: Can you update the comment too?

st0012 (commented):
Ok I think I understand the issue here better now:
- Themes can decide whether they want to display non-text files. `hanna` displays them but `darkfish` doesn't
- But RDoc's cross referencing feature doesn't allow referencing to non-text files (e.g. `lib/foo.rb`) regardless if the theme generates them or not

Is my understanding correct?
If it is:
- I think it's better to revert the 2nd commit as it makes darkfish generate a lot of new pages while most of them are empty. Sorry for the back and forth.
- Can you also update `rdoc-ref`'s documentation in `doc/rdoc/markup_reference.rb` to mention how to link to files?

jeremyevans:
I removed the darkfish commit, and added commits to updated the comment and show a usage example. Probably best to squash merge these, but I kept them separate to facilitate review.

st0012 (approved):
Thank you!

PR #1379, opened 2025-06-18
Added gemspec file to gem package

hsbt:
The original `gemspec` need to install `rdoc` as bundled gems at the official Ruby package.

Fix https://bugs.ruby-lang.org/issues/21312

### DIFF
--- a/rdoc.gemspec
+++ b/rdoc.gemspec
@@ -54,9 +54,7 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
     "rdoc.gemspec",
   ]
   template_files = Dir.glob("lib/rdoc/generator/template/**/*")
-  lib_files = Dir.chdir(File.expand_path('..', __FILE__)) do
-    Dir.glob("lib/**/*.{rb,kpeg,ry}")
-  end
+  lib_files = Dir.glob("lib/**/*.{rb,kpeg,ry}", base: File.expand_path('..', __FILE__))
 
   s.files = (non_lib_files + template_files + lib_files).uniq
 
--- a/rdoc.gemspec
+++ b/rdoc.gemspec
@@ -54,7 +54,9 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
     "rdoc.gemspec",
   ]
   template_files = Dir.glob("lib/rdoc/generator/template/**/*")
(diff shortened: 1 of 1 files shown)


hsbt:
I'm checking this change will fix to install the broken gemspec with the current Ruby HEAD.

rdoc.gemspec
```diff
@@ -38,6 +38,7 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
   # for ruby core repository. It was generated by
   # `git ls-files -z`.split("\x0").each {|f| puts "    #{f.dump}," unless f.start_with?(*%W[test/ spec/ features/ .]) }
   non_lib_files = [
+    "rdoc.gemspec",
```
kou: Could you keep this list in alphabetical order?

hsbt: Sure!

hsbt:
I'm investigating another issue for `make install` with rdoc installation.

rdoc.gemspec
```diff
@@ -51,9 +51,12 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
     "exe/rdoc",
     "exe/ri",
     "man/ri.1",
+    "rdoc.gemspec",
   ]
   template_files = Dir.glob("lib/rdoc/generator/template/**/*")
-  lib_files = Dir.glob("lib/**/*.{rb,kpeg,ry}")
+  lib_files = Dir.chdir(File.expand_path('..', __FILE__)) do
+    Dir.glob("lib/**/*.{rb,kpeg,ry}")
+  end
```
kou: How about using `base:` instead of `Dir.chdir`?

```suggestion
  lib_files = Dir.glob("lib/**/*.{rb,kpeg,ry}", base: File.expand_path('..', __FILE__))
```

hsbt: 👍  This seems more smart.

hsbt:
I confirmed https://github.com/ruby/rdoc/pull/1379/commits/783adb8ba6a03f219edc36bb618d11ddf8bda159 is working fine with `make install` of ruby.

rdoc.gemspec
nobu: Why not `base: __dir__`?

deivid-rodriguez:
Thank you for fixing this @hsbt!

PR #1380, opened 2025-06-18
Use GITHUB_TOKEN instead of admin credential

hsbt:
`MATZBOT_GITHUB_WORKFLOW_TOKEN` have admin grant. It's unnecessary for `gh release create`.


hsbt:
https://github.com/ruby/rdoc/actions/runs/15725230955/job/44313652328?pr=1380 need to new release of rdoc because `lib/rdoc/markdown.rb` is not generated with development branch.

PR #1389, opened 2025-07-12
Standardize table and heading markup elements

vinistock:
I noticed that Markup elements are currently not standardized in any way - even though the code does make some assumptions.

For example, there are visitors that depend on all elements implementing `accept`, but no parent class or interface that will help guarantee every element implements it. Some elements are implemented as structs, while other are classes. Some have parent classes for shared functionality and APIs and others don't.

I propose that we standardize the markup elements a bit more, so that we can have a common interface for them. It will make it easier for the RDoc to Markdown translator if we can enforce that every markup element has to be able to print itself in Markdown - and I believe the code will be easier to understand too.

This PR creates a base `Markup::Element` class and I standardized `Heading` and `Table` as an example. I'm also finding it challenging to understand the logic with no clue about what types are and what is expected from each method, so I also propose we annotate some of these as documentation - even if we're not adopting type checking.

If folks agree, I will continue to standardize more markup elements.

### DIFF
--- a/lib/rdoc/markup/heading.rb
+++ b/lib/rdoc/markup/heading.rb
@@ -1,84 +1,100 @@
 # frozen_string_literal: true
-##
-# A heading with a level (1-6) and text
 
-RDoc::Markup::Heading =
-  Struct.new :level, :text do
+module RDoc
+  class Markup
+    # A heading with a level (1-6) and text
+    #
+    #  RDoc syntax:
+    #   = Heading 1
+    #   == Heading 2
+    #   === Heading 3
+    #
+    #  Markdown syntax:
+    #   # Heading 1
+    #   ## Heading 2
+    #   ### Heading 3
+    class Heading < Element
+      #: String
+      attr_reader :text
 
-  @to_html = nil
-  @to_label = nil
+      #: Integer
+      attr_accessor :level
 
-  ##
-  # A singleton RDoc::Markup::ToLabel formatter for headings.
+      # A singleton RDoc::Markup::ToLabel formatter for headings.
+      #: () -> RDoc::Markup::ToLabel
+      def self.to_label
+        @to_label ||= Markup::ToLabel.new
+      end
 
-  def self.to_label
-    @to_label ||= RDoc::Markup::ToLabel.new
-  end
+      # A singleton plain HTML formatter for headings. Used for creating labels for the Table of Contents
+      #: () -> RDoc::Markup::ToHtml
+      def self.to_html
+        @to_html ||= begin
+          markup = Markup.new
+          markup.add_regexp_handling CrossReference::CROSSREF_REGEXP, :CROSSREF
 
-  ##
-  # A singleton plain HTML formatter for headings.  Used for creating labels
-  # for the Table of Contents
+          @to_html = Markup::ToHtml.new nil
 
-  def self.to_html
-    return @to_html if @to_html
+          def @to_html.handle_regexp_CROSSREF(target)
+            target.text.sub(/^\\/, '')
+          end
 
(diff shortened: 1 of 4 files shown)


lib/rdoc/markup/heading.rb
```diff
@@ -1,84 +1,100 @@
...
+      end
+
+      # A singleton plain HTML formatter for headings. Used for creating labels for the Table of Contents
+      #: () -> RDoc::Markup::ToHtml
+      def self.to_html
+        @to_html ||= begin
+          markup = Markup.new
+          markup.add_regexp_handling CrossReference::CROSSREF_REGEXP, :CROSSREF
+
+          @to_html = Markup::ToHtml.new nil
```
tompng: I think it should be a local variable here `@to_html` → `to_html`
It will be assigned to `@to_html` at `@to_html ||= begin`

tompng (approved):
Standardizing markup element classes makes sense :+1:

lib/rdoc/markup/heading.rb
```diff
@@ -1,84 +1,100 @@
...
+
+          @to_html
+        end
+      end
+
+      #: (Integer, String) -> void
+      def initialize(level, text)
+        @level = level
+        @text = text
+        super()
```
kou: Initializing the base object (`Markup::Element`) before this object may be better:

```suggestion
        super()
        @level = level
        @text = text
```

Or we may want to remove `super()` here because the current `Markup::Element#initialize` is empty.

lib/rdoc/markup/heading.rb
```diff
@@ -1,84 +1,100 @@
...
+        label = +""
+        label << "#{context.aref}-" if context&.respond_to?(:aref)
+        label << aref
+        label
+      end
+
+      # HTML markup of the text of this label without the surrounding header element.
+      #: () -> String
+      def plain_html
+        text = self.text.dup
```
kou: It seems that we don't need `dup` here.

lib/rdoc/markup/heading.rb
```diff
@@ -1,84 +1,100 @@
...
+
+        if matched = text.match(/rdoc-image:[^:]+:(.*)/)
+          text = matched[1]
+        end
+
+        self.class.to_html.to_html(text)
```
kou: How about using more descriptive local variable name and remove `self.` from `self.text` something like:

```suggestion
        no_image_text = text

        if matched = no_image_text.match(/rdoc-image:[^:]+:(.*)/)
          no_image_text = matched[1]
        end

        self.class.to_html.to_html(no_image_text)
```

kou (approved):
+1

st0012 (commented):
I think this is a good idea 👍 But how is this base class compared to `RDoc::Markup::Raw`? Are they complementary or it could be a replacement?
Also, it may be worth to create a list of elements to update to parallelize/track the work?

lib/rdoc/markup/heading.rb
vinistock: Yeah, good idea

lib/rdoc/markup/heading.rb
vinistock: I moved the `super()` call above. I would keep it as if we end up adding shared state to the parent class, then we don't need to update all of the children.

vinistock:
> how is this base class compared to RDoc::Markup::Raw

`Raw` will also inherit from `Element` as it represents an element that gets inserted with no modifications in the document.


> it may be worth to create a list of elements to update to parallelize/track the work?

I can put it in an issue if you'd like. My understanding is that it would be everything under `lib/rdoc/markup`.

PR #1391, opened 2025-07-21
Fix backslash in RDoc monofont causing unwanted linking

### LINKED ISSUE #1390: Backslash in RDoc monofont causes linking
Bug can be reproduced by generating the RDoc for this script:

```ruby
class Foo
  # This is not linked:
  # <tt>.bar.hello()</tt>
  #
  # But these are linked:
  # <tt>.bar.hello(\)</tt>
  # <tt>.bar.hello(\\\\)</tt>
  def foo

  # Hello from bar
  def bar
  end
end
```

The generated RDoc HTML looks like this. Notice how the backslash causes the monofont to generate a link to `Foo#bar`:

Image

I cannot find a way to prevent this link from generating. I think this is a bug, but I might be missing something from the syntax.

st0012:
## Summary
- Fixes issue where backslashes in `<tt>` tags caused unwanted cross-reference links
- Adds validation to ensure escaped characters don't trigger false positive cross-references
- Preserves normal cross-reference functionality for valid method/class references

## Problem
When text in `<tt>` tags contains backslashes (e.g., `<tt>.bar.hello(\)</tt>`), RDoc was incorrectly creating cross-reference links. This happened because the backslash matched the escaped character pattern in `CROSSREF_REGEXP`, causing the entire string to be processed as a potential cross-reference.

## Solution
Added checks in both `handle_regexp_CROSSREF` and `convert_flow` to detect when a match is solely due to escaped character patterns that don't represent valid method/class references. These strings are now rendered as code without creating links.

## Test plan
- [x] Added new test case `test_convert_CROSSREF_backslash_escapes_in_monofont` to verify the fix
- [x] Confirmed existing tests pass
- [x] Tested with the reproduction case from the issue

Fixes #1390

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/lib/rdoc/markup/to_html_crossref.rb
+++ b/lib/rdoc/markup/to_html_crossref.rb
@@ -94,6 +94,18 @@ class RDoc::Markup::ToHtmlCrossref < RDoc::Markup::ToHtml
       return name if name =~ /\A[a-z]*\z/
     end
 
+    # Check if the match is solely due to an escaped character pattern.
+    # If the text contains backslash escapes but doesn't match any
+    # valid cross-reference pattern (class, method, etc.), return as-is
+    if name =~ /\\/ && 
+       name !~ /\A#{CLASS_REGEXP_STR}(?:[.#]|::)#{METHOD_REGEXP_STR}\z/o &&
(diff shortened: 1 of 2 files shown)


st0012:
Sorry for the noise. I ran the wrong claude command.

Issue #1319, opened 2025-03-18
Add support for reading search term from query param


nobodywasishere:
It would be really useful to be able to perform a search of the documentation using a query parameter, such as `https://docs.ruby-lang.org/en/3.4/?query=URI`, and have that search be automatically performed. It would allow for the creation of a bang without needing to resort to a google cse. See: https://github.com/kagisearch/bangs/pull/195

PR #1396, opened 2025-07-29
Search automatically on ?q= URL query parameter

jhawthorn:
Fixes https://github.com/ruby/rdoc/issues/1319 (I wanted this for exactly that reason 😅)

Automatically start searching when ?q= parameter is given in the URL, selecting and visiting the first match if available.

If not available, the query will be filled into the search box but we won't navigate away.

This should enable configuring a generated rdoc page as a search engine in one's web browser, or as a bang command in a search engine. For reference Rails' sdoc (which I think rdoc's search is originally based on) does this: https://api.rubyonrails.org/?q=changed%3F

### DIFF
--- a/lib/rdoc/generator/template/darkfish/js/darkfish.js
+++ b/lib/rdoc/generator/template/darkfish/js/darkfish.js
@@ -72,7 +72,15 @@ function hookSearch() {
   }
 
   search.select = function(result) {
-    window.location.href = result.firstChild.firstChild.href;
+    var href = result.firstChild.firstChild.href;
+    var query = this.input.value;
+    if (query) {
+      var url = new URL(href, window.location.origin);
+      url.searchParams.set('q', query);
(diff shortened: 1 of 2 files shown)


st0012:
Accepting `?q=` will be a big improvement ❤️ 
But is it possible to keep the search results open even after jumping like SDoc does?
If that's too complicated (I tried it a bit but failed) due to how darkfish works, I'd prefer not jumping immediately (or only jump when there's exactly one result), so we can show the results dropdown.

jhawthorn:
> But is it possible to keep the search results open even after jumping like SDoc does?

SDoc does this using turbolinks, which I think works extremely well in the docs, but is a bit more than I wanted to do in this PR.

How's this? I made it so that the page we jump to is given `?q=<search>&nav=0`, and we use that second parameter to disable the "feeling lucky".

st0012:
CI failure is unrelated.

PR #1397, opened 2025-08-03
searcher.js: Escape characters when building regexps

byroot:
Otherwise characters with regexp meaning ~like `.` won't behave as expected~.

Actually `.` when inside `[]` doesn't have a special meaning, however `\` and a few other break the search.

### DIFF
--- a/lib/rdoc/generator/template/json_index/js/searcher.js
+++ b/lib/rdoc/generator/template/json_index/js/searcher.js
@@ -58,7 +58,12 @@ Searcher.prototype = new function() {
 
   function buildRegexps(queries) {
     return queries.map(function(query) {
-      return new RegExp(query.replace(/(.)/g, '([$1])([^$1]*?)'), 'i');
+      var pattern = [];
+      for (var i = 0; i < query.length; i++) {
+        var char = RegExp.escape(query[i]);
+        pattern.push('([' + char + '])([^' + char + ']*?)');
+      }
+      return new RegExp(pattern.join(''), 'i');
     });
   }
 


st0012:
Sorry can you give me an example of the issue that you're trying to fix?

byroot:
For now this doesn't fix much of anything, because nothing really match these special characters.

But my initial plan was to allow to search by fully qualified name, e.g. `Array#map`, or `Array.new`, and this small bug was in the way. I didn't get around to implement the feature I wanted (yet?) because the JSON blob with the data that is being searched doesn't have fully qualified names. So it's much more work.

Hence I figured I'd just submit that small fix independently.

If you want to check it works, go on the current version of RDoc, and search for `\`:

```
Uncaught SyntaxError: unterminated character class
    buildRegexps https://ruby.github.io/rdoc/js/searcher.js:61
    buildRegexps https://ruby.github.io/rdoc/js/searcher.js:60
    find https://ruby.github.io/rdoc/js/searcher.js:17
    search https://ruby.github.io/rdoc/js/search.js:54
    observer https://ruby.github.io/rdoc/js/search.js:23
2 [searcher.js:61:14](https://ruby.github.io/rdoc/js/searcher.js)
```

On this branch, it doesn't crash.

glaszig:
quote mdn:

> Since ⁨May 2025⁩, this feature works across the latest devices and browser versions.

and that's why search broke in safari 17.6:

> TypeError: RegExp.escape is not a function. (In 'RegExp.escape(query[i])', 'RegExp.escape' is undefined) in searcher.js:63

i don't think only and exclusively browsers from 2025 should be required for a documentation page.

would you accept an alternative implementation even if it means "bundling" a polyfill?

i think even this should work for the purpose:

```js
if (!RegExp.escape) {
    RegExp.escape = function(string) {
        return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
    };
}
```

byroot:
> i don't think only and exclusively browsers from 2025 should be required for a documentation page.

Why the snark?

Of course it's an oversight, and of course a fix is welcome. It never occurred to me `Regexp.escape` might be a new thing, sorry my JS is rusty.

I'll be looking for a polyfill.

byroot:
https://github.com/ruby/rdoc/pull/1442

PR #1399, opened 2025-08-08
Fix incomplete string escaping or encoding

ptrgits:
fix the problem, replace `text.sub('\\', '')` with `text.gsub('\\', '')` so that all occurrences of the backslash character are removed from the string, not just the first one. This change should be made only on line 297 in the `handle_regexp_SUPPRESSED_CROSSREF` method in `lib/rdoc/markup/to_rdoc.rb`. No additional imports or method definitions are required, as `gsub` is a standard Ruby `String` method. This change preserves the intended functionality while ensuring that all backslashes are removed, not just the first.

### References
[RailsActiveRecord::Base::sanitize_sql](https://api.rubyonrails.org/classes/ActiveRecord/Sanitization/ClassMethods.html)

### DIFF
--- a/lib/rdoc/markup/to_rdoc.rb
+++ b/lib/rdoc/markup/to_rdoc.rb
@@ -294,7 +294,7 @@ class RDoc::Markup::ToRdoc < RDoc::Markup::Formatter
 
   def handle_regexp_SUPPRESSED_CROSSREF(target)
     text = target.text
-    text = text.sub('\\', '') unless in_tt?
+    text = text.gsub('\\', '') unless in_tt?
     text
   end
 


kou:
Could you also share an input that reproduces this problem and the actual result and expected result with the input?

It seems that `RailsActiveRecord::Base::sanitize_sql` https://api.rubyonrails.org/classes/ActiveRecord/Sanitization/ClassMethods.html#method-i-sanitize_sql doesn't use backslack.

tompng:
This part is registered on line 59
```
@markup.add_regexp_handling(/\\\S/, :SUPPRESSED_CROSSREF)
```
A string that matches `/\\\S/` (two length string starts with \) is passed to `handle_regexp_SUPPRESSED_CROSSREF`.

Example rdoc format:
```
== See \RDoc and \Array and \Hash. Use backslash(\\)
```

Example target passed to handle_regexp_SUPPRESSED_CROSSREF: `"\\R"`, `"\\A"`, `"\\H"`, `"\\\\"`

Expected output:
```
== See RDoc and Array and Hash. Use backslash(\)
```

`target` is always two-length string that starts with `\` and only the first backslash should be removed. If second char is `\`, it should remain.
`text.sub` is enough.  `text.gsub` will wrongly removes escaped backslashes.

PR #1404, opened 2025-08-11
Escape unescaped backslash at the end of code

### LINKED ISSUE #1402: Markdown: code ends with a backslash
RDoc markdown renders `` `\` `` as `<code></code>`.
This is because the markdown code is translated to `<code>\</code>`, and the backslash just escapes the following close tag and the open tag becomes invalid.

Other implementations (Github markdown at least) renders it as code only contains a backslash, `` \ ``.

nobu:
Fix #1402

### DIFF
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -498,6 +498,15 @@
   end
 
   ##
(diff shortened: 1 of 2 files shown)


kou (approved):
+1

PR #1408, opened 2025-08-13
[DOC] Fix `warn_missing_rdoc_ref` default value in doc

st0012:
It's actually enabled by default now. Also removes the setting from `.rdoc_options` as it's not needed.


kou (approved):
+1

vinistock (approved):

PR #1409, opened 2025-08-14
Fix warnings

nobu:


Gemfile
```diff
@@ -11,3 +11,4 @@ gem 'rubocop', '>= 1.31.0'
 gem 'gettext'
 gem 'prism', '>= 0.30.0'
 gem 'webrick'
+gem 'tsort'
```
tompng: RDoc explicitly use tsort, so I think this dependency should be added in gemspec

tompng (approved):

PR #1410, opened 2025-08-14
Commit generated files

st0012:
Closes https://github.com/ruby/rdoc/issues/1107 and will likely prevent issues like https://bugs.ruby-lang.org/issues/21541 in the future

With ignoring generated files in the source control, we always need one extra step to make the gem usable, which limits how the gem can be distributed (#1107) and make post-distribution issues easier to happen (which are in general harder to fix too).

If gems like ~~`prism` and~~ `rbs` and `debug` can avoid this issue by committing generated files and verifying them on CI, I think we should do that too.


Rakefile
```diff
@@ -68,28 +54,49 @@ parsed_files = PARSER_FILES.map do |parser_file|
     puts "Generating #{parsed_file}..."
     case ext
     when '.ry' # need racc
-      racc = Gem.bin_path 'racc', 'racc'
       rb_file = parser_file.gsub(/\.ry\z/, ".rb")
-      ruby "#{racc} -l -E -o #{rb_file} #{parser_file}"
+      sh "bundle exec racc -l -E -o #{rb_file} #{parser_file}"
```
kou: In general, `sh("command", "arg1", "arg2", ...)` is safer than `sh("command line")`:

```suggestion
      sh "bundle", "exec", "racc", "-l", "-E", "-o", rb_file, parser_file
```

Rakefile
```diff
@@ -68,28 +54,49 @@ parsed_files = PARSER_FILES.map do |parser_file|
...
       File.open(rb_file, 'r+') do |f|
         newtext = "# frozen_string_literal: true\n#{f.read}"
         f.rewind
         f.write newtext
       end
     when '.kpeg' # need kpeg
-      kpeg = Gem.bin_path 'kpeg', 'kpeg'
       rb_file = parser_file.gsub(/\.kpeg\z/, ".rb")
-      ruby "#{kpeg} -fsv -o #{rb_file} #{parser_file}"
+      sh "bundle exec kpeg -fsv -o #{rb_file} #{parser_file}"
```
kou:
```suggestion
      sh "bundle", "exec", "kpeg", "-fsv", "-o", rb_file, parser_file
```

Rakefile
```diff
@@ -68,28 +54,49 @@ parsed_files = PARSER_FILES.map do |parser_file|
...
+    unless File.exist?(file)
+      abort "Generated file #{file} does not exist!"
+    end
+  end
+
+  diff_output = `git diff --exit-code #{parsed_files.join(' ')} 2>&1`
+  unless $?.success?
+    puts "Generated parser files are out of date!"
+    puts "Please run 'rake generate' and commit the changes."
+    puts "\nDifferences found:"
```
kou: No argument `puts` may be more (a bit) readble than `\n`: 

```suggestion
    puts
    puts "Differences found:"
```

kou (approved):
+1

tompng:
I agree for adding generated files :+1:

CI seems to pass even if generated files are out of date.
Diff to markdown.kpeg: https://github.com/tompng/rdoc/actions/runs/16994310600 All tests passes
Diff to block_parser.ry: https://github.com/tompng/rdoc/actions/runs/16995015009 macos-latest always fails, windows-latest always passes.

Reproduce locally:
1. Add a diff to `markdown.kpeg` and commit. verify_generated correctly fails.
2. Switch to another branch and go back again. verify_generated wrongly passes.
3. `touch markdown.kpeg` and try `rake verify_generated` again. It correctly fails.

vinistock (approved):

PR #1411, opened 2025-08-20
[DOC] Tweaks for markup reference intro

BurdetteLamar:
In working on a markdown reference, I've noticed that the markup reference could use some work, too.

This adds some clarity to the intro, and has a couple of other small tweaks.

Much more to come.

### DIFF
--- a/doc/rdoc/markup_reference.rb
+++ b/doc/rdoc/markup_reference.rb
@@ -1,5 +1,10 @@
 require 'rdoc'
 
+# This page is about RDoc's +rdoc+ input format,
+# which here we'll call "RDoc markup" or simply "markup."
+#
+# == About This \Class
+#
 # \Class \RDoc::MarkupReference exists only to provide a suitable home
 # for a reference document for \RDoc markup.
 #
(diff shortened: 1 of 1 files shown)


doc/rdoc/markup_reference.rb
```diff
@@ -21,43 +26,44 @@
...
 #   - Documentation may be derived from the Ruby code itself;
-#     see {Documentation Derived from Ruby Code}[rdoc-ref:RDoc::MarkupReference@Documentation+Derived+from+Ruby+Code].
+#     see {Derived Documentation}[rdoc-ref:RDoc::MarkupReference@Derived+Documentation].
 #
 # - <tt>.c</tt> (C code file): markup is parsed from C comments.
 #   A comment that immediately precedes
 #   a function that implements a Ruby method,
 #   or otherwise immediately precedes the definition of a Ruby object,
 #   becomes the documentation for that object.
-# - <tt>.rdoc</tt> (\RDoc markup text file) or <tt>.md</tt> (\RDoc markdown text file):
```
peterzhu2118, line 49: It seems like we're losing the information here that we generate docs from markdown files.

BurdetteLamar: My thinking is that this is the markup reference.  I don't know whether a markup file can declare itself as markdown.  But will restore if you like.

peterzhu2118: Ah I see. Might be good to include it in a separate bullet point and make it clear that the rest of this guide doesn't apply since it's markdown. This section seems to list all of the sources that RDoc generates documentation from.

BurdetteLamar: Info added.

peterzhu2118 (approved):

Issue #1344, opened 2025-04-20
Adding YARD document support to RDoc


okuramasafumi:
## Background

I implemented a plugin system for RDoc in https://github.com/ruby/rdoc/pull/1321 and YARD parsing plugin is implemented.
In the discussion with @st0012, @kou and @vinistock we decided that we should implement YARD parsing feature as a standalone feature without plugin.
The note exists in https://github.com/ruby/rdoc/discussions/1257#discussioncomment-12882912

## Steps

There are some steps to have effective YARD parsing feature in RDoc.

1. Adding support of YARD tags that already works with current RDoc such as `@yield` and `@private`
2. Adding support of YARD tags that needs simple modification of RDoc such `@deprecated`
3. Adding features such as type support to utilize information from YARD
4. Adding support of YARD directives and macros

I will implement these step by step.

## Note

The idea of plugin system is not completely abandoned but postponed.

kou:
@st0012 Sorry. I forgot to confirm whether YARD style tag feature should be orthogonal to markups such as RDoc format and Markdown or not.

If YARD style tag feature is orthogonal to markups, we may implement it in `RDoc::Markup::PreProcess`, `RDoc::Parser::*` or somewhere.
If YARD style tag feature is NOT orthogonal to markups, we will implement it in `RDoc::{Markdown,Markup,RD,TomDoc}`.

(I'm not familiar with RDoc code base yet but I feel that YARD style tag feature is orthogonal to markups. It seems that YARD also uses orthogonal approach.)

@okuramasafumi How about using check list markup for the `Steps` section for easy to track?

```markdown
* [ ] Adding support of YARD tags that already works with current RDoc such as `@yield` and `@private`
* [ ] Adding support of YARD tags that needs simple modification of RDoc such `@deprecated`
* [ ] Adding features such as type support to utilize information from YARD
* [ ] Adding support of YARD directives and macros
```

okuramasafumi:
@kou At first I thought I can use sub-issues feature, but I couldn't. So I agree, I'll converted them into a list.
And I was going to implement YARD parsing feature in `RDoc::TomDoc` class using `RDoc::Markup::PreProcess` just like current `RDoc::TomDoc` class since YARD is similar to TomDoc from RDoc perspective (another documentation format).
But if we want to add just tag support to RDoc without specifying `markup: yard` directive, we might want to find another approach, maybe without `RDoc::TomDoc` class.

okuramasafumi:
[...]

PR #1416, opened 2025-08-27
feat(yard): add initial support of YARD tags

okuramasafumi:
At this time, `@yield`, `@private` and `@api private` are supported.
Ref: #1344

### DIFF
--- a/lib/rdoc/parser/prism_ruby.rb
+++ b/lib/rdoc/parser/prism_ruby.rb
@@ -510,13 +510,21 @@ class RDoc::Parser::PrismRuby < RDoc::Parser
 
     receiver = receiver_name ? find_or_create_module_path(receiver_name, receiver_fallback_type) : @container
     meth = RDoc::AnyMethod.new(nil, name, singleton: singleton)
+
     if (comment = consecutive_comment(start_line))
       handle_consecutive_comment_directive(@container, comment)
       handle_consecutive_comment_directive(meth, comment)
 
       comment.normalize
       meth.call_seq = comment.extract_call_seq
+
+      # Save visibility before comment assignment (where YARD processing happens)
+      visibility_before = meth.visibility
       meth.comment = comment
+      # Check if YARD tags changed the visibility
+      if meth.visibility != visibility_before
+        visibility = meth.visibility
+      end
     end
     handle_modifier_directive(meth, start_line)
     handle_modifier_directive(meth, args_end_line)
@@ -527,7 +535,7 @@ class RDoc::Parser::PrismRuby < RDoc::Parser
       receiver,
       meth,
       line_no: start_line,
-      visibility: visibility,
+      visibility: visibility,  # Use YARD visibility if set, otherwise use Ruby visibility
       params: params,
       calls_super: calls_super,
       block_params: block_params,


lib/rdoc/yard.rb
```diff
@@ -0,0 +1,135 @@
...
+      # Remove the @api private line from comment
+      text.gsub!(/^\s*#?\s*@api\s+private\s*$\n?/, '')
+    end
+  end
+
+  ##
+  # Extract parameter names from YARD type specification
+  # e.g., "[String, Integer]" -> "value1, value2"
+  # e.g., "[item, index]" -> "item, index"
+  def extract_param_names(params_string)
```
okuramasafumi, line 114: This part is kind of too much, but YARD's `@yield` might cause confusion to use type name instead of param name. Here it replaces type name with temporary param name,

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+
+      # Remove the @api private line from comment
+      text.gsub!(/^\s*#?\s*@api\s+private\s*$\n?/, '')
+    end
+  end
+
+  ##
+  # Extract parameter names from YARD type specification
+  # e.g., "[String, Integer]" -> "value1, value2"
+  # e.g., "[item, index]" -> "item, index"
```
tompng, line 113: Example and implementation does not match.
"String, Integer" -> "arg1, arg2"
"item, index" -> "item, index"

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+  ##
+  # Process @yield tags to extract block parameters
+  def process_yield_tags
+    return unless code_object.respond_to?(:block_params=)
+
+    # Skip if already has block_params from :yields: directive
+    return if code_object.block_params && !code_object.block_params.empty?
+
+    # Match @yield [params] description
+    if text =~ /^\s*#?\s*@yield\s*(?:\[([^\]]*)\])?\s*(.*?)$/
```
tompng, line 70: YARD tag have a common format `# @tagname free_format_argument_part` `# @!tagname free_format_argument_part`
First scan tags in a generic way, and pass it to each tag-process part is better.
This way, parsing multiline tags would be easier.

lib/rdoc/code_object/method_attr.rb
```diff
@@ -156,6 +156,20 @@ def store=(store)
     @file = @store.add_file @file.full_name if @file
   end
 
+  ##
+  # Process comment as YARD comment
+  def comment=(comment)
+    # Process YARD tags while we still have the RDoc::Comment object
+    if comment.is_a?(RDoc::Comment)
+      RDoc::YARD.process(comment, self)
```
tompng, line 164: In the comment below, `@foo` is not a YARD tag.
```ruby
# @foo &&= bar
# ^^^^^^^^^^^^
def visit_instance_variable_and_write_node(node)
end
```
Ruby's comment frequently contains code example that uses instance variable. It's the same format as YARD tags.

So we need to make YARD feature opt-in, or only accept specific yard tags. (opt-in is better)

```ruby
# This is just an idea. Specify that this file uses YARD format.
# :markup: yard

class A
  # @yard-tag
  def foo; end
end
```

```ruby
# found 37 files in github code search
# @api = ...
```

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -527,7 +535,7 @@ def add_method(name, receiver_name:, receiver_fallback_type:, visibility:, singl
       receiver,
       meth,
       line_no: start_line,
-      visibility: visibility,
+      visibility: visibility,  # Use YARD visibility if set, otherwise use Ruby visibility
```
tompng, line 538: This comment should be where YARD actually overrides visibility, near `visibility = meth.visibility` line.

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+      code_object.visibility = :private
+
+      # Remove the @private line from comment
+      text.gsub!(/^\s*#?\s*@private\s*$\n?/, '')
+    end
+  end
+
+  ##
+  # Process @api tags (specifically @api private)
+  def process_api_tags
```
tompng, line 99: I think `@api private` sets documenting level visibility which RDoc doesn't have. It's not for making a method private_method.
One idea is to let it act as `nodoc` for now.

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+    @comment = comment
+    @code_object = code_object
+  end
+
+  def process
+    return comment unless comment && code_object
+
+    # Only process RDoc::Comment objects to avoid modifying plain strings
+    # that aren't actually documentation comments
+    return comment unless comment.is_a?(RDoc::Comment)
```
tompng, line 34: These guards are not needed. Method signature will be simple.
```rbs
(String | RDoc::Comment | nil, RDoc::CodeObject | nil) -> void
# ↓
(RDoc::Comment, RDoc::CodeObject) -> void
```

`RDoc::Parser::PrismRuby` don't set String as comment, and it will be the default in the future.

lib/rdoc/code_object/constant.rb
```diff
@@ -174,6 +174,20 @@ def store=(store)
     @file = @store.add_file @file.full_name if @file
   end
 
+  ##
+  # Process comment as YARD comment
+  def comment=(comment)
```
tompng, line 179: Handling YARD in `comment=` seems too aggressive. It can be triggered from many different situation and places.
```ruby
def add_alias
  # Maybe this is ok, but I can't confidently say that this is safe
  method.comment = an_alias.comment
end
```

How about explicitly processing it in RDoc::Parser::PrismRuby?
```ruby
def add_comment
  ...
  comment.normalize
  override_visibility, some_other_information = handle_yard(meth, comment)
  if override_visibility
    do_something
  end
  meth.comment = comment
  ...
end
```

`RDoc::Parser::Ruby` is too complicated (and buggy) to maintain, so the system design of YARD-support shouldn't be affected by RDoc::Parser::Ruby implementation detail. I think supporting YARD only in RDoc::Parser::PrismRuby is enough.

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -510,13 +510,21 @@ def add_method(name, receiver_name:, receiver_fallback_type:, visibility:, singl
...
       comment.normalize
       meth.call_seq = comment.extract_call_seq
+
+      # Save visibility before comment assignment (where YARD processing happens)
+      visibility_before = meth.visibility
       meth.comment = comment
+      # Check if YARD tags changed the visibility
+      if meth.visibility != visibility_before
+        visibility = meth.visibility
+      end
```
kou, line 527: It seems that this shows that the proposed approach isn't integrated with the current implementation clearly. Can we use another more integrated approach?

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+##
+# YARD compatibility module for RDoc
+#
+# This module provides support for parsing YARD tags in RDoc comments.
+# Currently supports:
+# - @yield [params] - Describes block parameters
+# - @private - Marks method/class as private
+# - @api private - Alternative private marking
+
+class RDoc::YARD
```
kou, line 12: Can we improve this name?

I think that "YARD" isn't suitable here because we don't support full YARD features. We only focus on YARD tags here.

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+
+  ##
+  # Processes YARD tags in the given comment text for a code object
+  #
+  # Returns modified comment text with YARD tags processed
+  def self.process(comment, code_object)
+    new(comment, code_object).process
+  end
+
+  attr_reader :comment, :code_object, :text
```
kou, line 22: Do we need them?

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+
+class RDoc::YARD
+
+  ##
+  # Processes YARD tags in the given comment text for a code object
+  #
+  # Returns modified comment text with YARD tags processed
+  def self.process(comment, code_object)
+    new(comment, code_object).process
+  end
```
kou, line 20: Do we need this shortcut?

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+
+  def process
+    return comment unless comment && code_object
+
+    # Only process RDoc::Comment objects to avoid modifying plain strings
+    # that aren't actually documentation comments
+    return comment unless comment.is_a?(RDoc::Comment)
+
+    # Skip processing if comment is in TomDoc format
+    return comment if comment.format == 'tomdoc'
```
kou, line 37: Can we use `comment.tomdoc?`

BTW, is this a responsibility of this class? It seems that this is a responsibility of callers.

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+    @text = comment.text.dup
+
+    # Process @yield tags
+    process_yield_tags
+
+    # Process @private tags
+    process_private_tags
+
+    # Process @api tags
+    process_api_tags
```
kou, line 48: In general, we should not parse the target multiple times by "start parsing -> extract a tag -> process the tag -> remove the tag -> finish parsing" one by one destructively (`gsub!`). It may require additional cares in following processes.
See also: https://magazine.rubyist.net/articles/0010/0010-CodeReview.html#pukipastr_parse (Japanese)
 
Can we parse the comment only once?

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    # Only update comment.text if we actually modified it
+    # This preserves the @document for comments created from parsed documents
+    if text != comment.text
+      comment.text = text
+    end
```
kou, line 54: Do we need to update `comment` destructively in this class?

It seems that we can receive `comment.text` in `#initialize` and returns (not destructively) processed text. Updating `comment.text` is a responsibility of the callers.

lib/rdoc/yard.rb
kou, line 70: `@yield` can use multiple lines: https://github.com/lsegal/yard/blob/5b93b3a4f15a52f3165edef112e718149d294c38/lib/yard/parser/source_parser.rb#L223-L224

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+      text.gsub!(/^\s*#?\s*@yield.*?$\n?/, '')
+    end
+  end
+
+  ##
+  # Process @private tags to set visibility
+  def process_private_tags
+    return unless code_object.respond_to?(:visibility=)
+
+    if text =~ /^\s*#?\s*@private\s*$/
```
kou, line 89: Why is `?` needed after `#`?

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+      text.gsub!(/^\s*#?\s*@api\s+private\s*$\n?/, '')
+    end
+  end
+
+  ##
+  # Extract parameter names from YARD type specification
+  # e.g., "[String, Integer]" -> "value1, value2"
+  # e.g., "[item, index]" -> "item, index"
+  def extract_param_names(params_string)
+    return '' if params_string.nil? || params_string.empty?
```
kou, line 115: Do we need these checks?

I feel that it's responsibility of callers.

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+
+    # If params look like variable names (lowercase start), use them
+    if params_string =~ /^[a-z_]/
```
kou, line 118: We should use `\A` not `^` for the start of string. `^` is for the start of line.
See also: https://www.clear-code.com/blog/2015/3/25.html (Japanese)

lib/rdoc/yard.rb
```diff
@@ -0,0 +1,134 @@
...
+      params_string.split(',').map(&:strip).join(', ')
```
kou, line 119:
```suggestion
      params_string.split(/\s*,\s*/).join(', ')
```

Issue #1414, opened 2025-08-20
Can a paragraph contain blocks?


BurdetteLamar:
In the markup reference, I've written that a paragraph may contain other blocks.  Is this true?  Would it be better to say that a paragraph may contain only embedded text markup?

Issue #1413, opened 2025-08-20
Is text markup a block?


BurdetteLamar:
In the markup reference, I've included text markup as a type of block (along with paragraph, list, heading, etc.).  But is it really useful to think of text markup in this way?  Would it be better to treat text markup as embed-able in certain other blocks?

PR #1417, opened 2025-08-27
[DOC] Revision of Paragraphs section

BurdetteLamar:
Although technically it's possible to nest some types of blocks in a paragraph (by eschewing newlines), it's poor practice.  Is the language here too strong?  Or should there be explanation and warnings?

### DIFF
--- a/doc/rdoc/markup_reference.rb
+++ b/doc/rdoc/markup_reference.rb
@@ -124,14 +124,8 @@ require 'rdoc'
 #
 #   You'll love it.
 #
-# A paragraph may contain nested blocks, including:
+# A paragraph may contain only one type of nested block:
 #
-# - {Verbatim text blocks}[rdoc-ref:RDoc::MarkupReference@Verbatim+Text+Blocks].
-# - {Code blocks}[rdoc-ref:RDoc::MarkupReference@Code+Blocks].
-# - {Block quotes}[rdoc-ref:RDoc::MarkupReference@Block+Quotes].
-# - {Lists}[rdoc-ref:RDoc::MarkupReference@Lists].
-# - {Headings}[rdoc-ref:RDoc::MarkupReference@Headings].
-# - {Horizontal rules}[rdoc-ref:RDoc::MarkupReference@Horizontal+Rules].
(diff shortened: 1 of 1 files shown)


doc/rdoc/markup_reference.rb
```diff
@@ -124,14 +124,8 @@
...
-# A paragraph may contain nested blocks, including:
+# A paragraph may contain only one type of nested block:
 #
-# - {Verbatim text blocks}[rdoc-ref:RDoc::MarkupReference@Verbatim+Text+Blocks].
-# - {Code blocks}[rdoc-ref:RDoc::MarkupReference@Code+Blocks].
-# - {Block quotes}[rdoc-ref:RDoc::MarkupReference@Block+Quotes].
-# - {Lists}[rdoc-ref:RDoc::MarkupReference@Lists].
-# - {Headings}[rdoc-ref:RDoc::MarkupReference@Headings].
-# - {Horizontal rules}[rdoc-ref:RDoc::MarkupReference@Horizontal+Rules].
 # - {Text Markup}[rdoc-ref:RDoc:MarkupReference@Text+Markup].
```
peterzhu2118, line 129: IMO a list with one bullet point shouldn't be a list.

I also don't know what the point of this change is, what is considered a "nested block"? The "Text Markup" section just talks about how to format the text with bold, italics, and monofont. Is that even considered a block or just formatting?

At any rate, I think we either keep the original (and maybe reword it) or just remove all of this completely.

BurdetteLamar: I've sometimes used a 1-item list to make treatments parallel, which I've thought helpful.  (The other sections put nestings in lists.)

About your questions:  I've asked them myself:

- https://github.com/ruby/rdoc/issues/1414
- https://github.com/ruby/rdoc/issues/1413

BurdetteLamar:
@peterzhu2118, might it be better to close this PR and instead explore the idea of demoting text markup from its status as a block?

peterzhu2118:
> might it be better to close this PR and instead explore the idea of demoting text markup from its status as a block?

Maybe? I'm not sure how we want to define what a block is.

BurdetteLamar:
Closing, pending other changes.

PR #1419, opened 2025-09-02
[DOC] New doc 'RDoc Documents'

BurdetteLamar:


BurdetteLamar:
> I think there is a lot of duplicate information in the ["Writing Documentation" section in the README.md](https://github.com/ruby/rdoc?tab=readme-ov-file#label-Writing+Documentation). I think we should consolidate this with that (or link to this from that).

I agree.  I'd like to go ahead and merge this (redundancy is relatively harmless), and work on the README later on.

BurdetteLamar:
@peterzhu2118, I also have in-work `rdoc_sources.md`, which the README can also link to.

st0012:
I think this is a good high level overview that should be put into the readme instead of a separate doc.

BurdetteLamar:
> I think this is a good high level overview that should be put into the readme instead of a separate doc.

I wrote this to formally establish the term _document_, so that I can link to it whenever I write "top of the document" or "end-of-document."

I'd prefer its remaining a separate doc, and linking to it from the readme.  I have some other things in-work, and had planned on re-working the readme in the near future.

PR #1421, opened 2025-09-06
End to end tests

BurdetteLamar:
I think there's been some discussion about end-to-end testing.  Here's a start.

There are two questions:

- Is the community interested in end-to-end testing (and, crucially, willing to review PRs)?
- Is the structure offered here workable?  If not, what would be better?

### DIFF
--- a/test/rdoc/end_to_end/heading_test.rb
+++ b/test/rdoc/end_to_end/heading_test.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+require_relative 'helper'
+
+class HeadingTest < XrefTestCase
+
+  def test_headings
+    markup = <<MARKUP
+= Section 1
+== Section 1.1
+=== Section 1.1.1
+==== Section 1.1.1.1
+===== Section 1.1.1.1.1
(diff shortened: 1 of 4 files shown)


BurdetteLamar:
@peterzhu2118 and @st0012:  I see that there are many failing checks.  I'd like to wait for opinions about the test code before addressing the failures  (If no one likes this PR, it needs to be closed, not fixed.)

BurdetteLamar:
@st0012 and @peterzhu2118:  The checks now all pass, so I think this is ready for review.

st0012 (commented):
By end to end, I do mean running `rdoc` executable directly in tests 👍 
But if we do that, we probably don't need most or all the helpers and stuff from `XrefTestCase`. Instead, we can create a new `IntegrationTestCase` that provides the `run_rdoc` method directly.

BurdetteLamar:
To be replaced by https://github.com/ruby/rdoc/pull/1430.

BurdetteLamar:
Revised and moved to [PR 1430](https://github.com/ruby/rdoc/pull/1430).

PR #1422, opened 2025-09-07
Migrate README to markdown

st0012:


rdoc.gemspec
```diff
@@ -60,7 +60,7 @@ RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentat
 
   s.files = (non_lib_files + template_files + lib_files).uniq
 
-  s.rdoc_options = ["--main", "README.rdoc"]
+  s.rdoc_options = ["--main", "README.md"]
   s.extra_rdoc_files += s.files.grep(%r[\A[^\/]+\.(?:rdoc|md)\z])
```
kou, line 64: Can we use `.rdoc_options` instead of duplicated configurations here?

nobu: This option is in `.rdoc_options` already, so we can drop it here.

st0012: `.rdoc_options` is not packaged in the gem tho. I also don't think we should use it for gem's doc options anyway because as of now `.rdoc_options` usually specify configs specifically for web docs.
We should probably provide clearer instructions or even APIs to help distinguish options for different environments. But I don't have any good ideas on this yet.

kou: I see.

PR #1427, opened 2025-09-27
Fix incorrect method linking in `tt` tags

### LINKED ISSUE #1390: Backslash in RDoc monofont causes linking

st0012:
Fix #1390

### DIFF
--- a/lib/rdoc/markup/to_html_crossref.rb
+++ b/lib/rdoc/markup/to_html_crossref.rb
@@ -228,12 +228,12 @@ class RDoc::Markup::ToHtmlCrossref < RDoc::Markup::ToHtml
   # When the candidate occupies the whole span (aside from trailing
   # punctuation), the tt markup is replaced by the resolved cross-reference.
 
-  def convert_tt_crossref(flow, index)
-    opener = flow[index]
+  def convert_tt_crossref(flow_items, index)
+    opener = flow_items[index]
     return unless tt_tag?(opener.turn_on)
 
-    string = flow[index + 1]
-    closer = flow[index + 2]
+    string = flow_items[index + 1]
+    closer = flow_items[index + 2]
 
     return unless String === string
     return unless RDoc::Markup::AttrChanger === closer
--- a/lib/rdoc/markup/to_html_crossref.rb
+++ b/lib/rdoc/markup/to_html_crossref.rb
@@ -197,14 +197,24 @@ class RDoc::Markup::ToHtmlCrossref < RDoc::Markup::ToHtml
         if tt_tag?(item.turn_on) and
           String === (str = flow[i]) and
           RDoc::Markup::AttrChanger === flow[i+1] and
-          tt_tag?(flow[i+1].turn_off, true) and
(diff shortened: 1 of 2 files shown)


lib/rdoc/markup/to_html_crossref.rb
```diff
@@ -184,43 +184,73 @@ def link(name, text, code = true, rdoc_ref: false)
...
   end
+
+  private
+
+  ##
+  # Detects <tt>...</tt> spans that contain a single cross-reference candidate.
+  # When the candidate occupies the whole span (aside from trailing
+  # punctuation), the tt markup is replaced by the resolved cross-reference.
+
+  def convert_tt_crossref(flow, index)
```
kou: Should we use `flow_items` here too?

test/rdoc/rdoc_markup_to_html_crossref_test.rb
```diff
@@ -33,6 +33,33 @@ def test_convert_CROSSREF
...
+    result = formatter.convert '<tt>.bar.hello(\)</tt>'
+
+    assert_equal para('<code>.bar.hello(\)</code>'), result
+
+    formatter = RDoc::Markup::ToHtmlCrossref.new @options, 'C9.html', @c9_b
+    result = formatter.convert '<tt>.bar.hello(\\\\)</tt>'
+
+    assert_equal para('<code>.bar.hello(\\)</code>'), result
+  ensure
+    @options.hyperlink_all = hyperlink_all
```
kou: It seems that we don't need this because we always set `@options.hyperlink_all` in `setup`.

test/rdoc/rdoc_markup_to_html_crossref_test.rb
```diff
@@ -33,6 +33,33 @@ def test_convert_CROSSREF
     assert_equal para("<code># :stopdoc:</code>:"), result
   end
 
+  def test_convert_CROSSREF_backslash_in_tt
+    hyperlink_all = @options.hyperlink_all
+    @options.hyperlink_all = false
+
+    formatter = RDoc::Markup::ToHtmlCrossref.new @options, 'C9.html', @c9_b
+    result = formatter.convert '<tt>C1</tt>'
+
```
kou, line 40: How about removing this empty line because `format = ...; result = ...; assert_equal ...` are related?

```suggestion
```

kou (approved):
+1

test/rdoc/rdoc_markup_to_html_crossref_test.rb
st0012: I've reduced the lines this test takes

PR #1428, opened 2025-09-27
Preserve inline styling inside tidy link labels

### LINKED ISSUE #1231: Markdown links with inline code are parsed incorrectly

st0012:
Fixes #1231


lib/rdoc/markup/to_html.rb
```diff
@@ -458,4 +462,104 @@ def to_html(item)
     super convert_flow @am.flow item
   end
 
+  private
+
+  def append_to_tidy_label(fragment)
+    return unless @tidy_link_buffer
+
```
kou: It seems that we can remove this check here because it seems that callers are responsible to check `tidy_link_capturing?` before calling this.

lib/rdoc/markup/to_html.rb
```diff
@@ -458,4 +462,104 @@ def to_html(item)
...
+    link = gen_url(url, label_html)
+
+    return link if trailing.empty?
+
+    link + CGI.escapeHTML(trailing)
+  end
+
+  def extract_tidy_link_parts(text)
+    if text =~ /^(.*?)\}\[(.*?)\](.*)$/
```
kou:
```suggestion
    if text =~ /\A(.*?)\}\[(.*?)\](.*)\z/
```

lib/rdoc/markup/to_html.rb
```diff
@@ -458,4 +462,104 @@ def to_html(item)
...
+
+  def append_to_tidy_label(fragment)
+    return unless @tidy_link_buffer
+
+    @tidy_link_buffer << fragment
+  end
+
+  def convert_complete_tidy_link(text)
+    return text unless
+      text =~ /^\{(.*)\}\[(.*?)\]$/ or text =~ /^(\S+)\[(.*?)\]$/
```
kou:
```suggestion
      text =~ /\A\{(.*)\}\[(.*?)\]\z/ or text =~ /\A(\S+)\[(.*?)\]\z/
```

kou (approved):
+1

lib/rdoc/markup/to_html.rb
```diff
@@ -359,7 +356,14 @@ def accept_table(header, body, aligns)
   # CGI-escapes +text+
 
   def convert_string(text)
-    CGI.escapeHTML text
+    html = CGI.escapeHTML text
+
+    if tidy_link_capturing?
+      append_to_tidy_label html
+      ''
```
tompng: I think we need to handle crossref and also append it to tidy_link_buffer.
Currently, if styling is considered crossref or crossref-suppress, it is not hanlded correctly.
```md
[Link to `RDoc` and `Foo` and `\RDoc` and `Bar`](https://example.com)
```
Generate with `rdoc file.md`
Generate with `rdoc file.md`
Generate with `rdoc file.md lib/rdoc/rdoc.rb`
Generate with `rdoc file.md lib/rdoc/rdoc.rb`

How about changing `res << text` to `append_text(res, text)` in `convert_flow` and define
`def append_text(res, text)` as an alias of `emit_tidy_link_fragment` here?
Then we don't need to return `''` to suppress `res << text`. Cross-ref in label problem might be solved easier.

st0012: Good catch 👍 I think it should be fixed now.

PR #1432, opened 2025-10-12
New theme: "Aliki"

st0012:
This theme's "backend" (Ruby part) is mostly based on Darkfish. But design and frontend features is (IMO) a lot better than it.

The work was assisted by AI coding tools through a few weeks. I cleaned up the history because they're not helping reviewing anyway.

(Aliki is the name of my cat, which was not involved in this work)

### Key features

- 3 column design (which matches most API docs nowadays)
- Switchable dark mode
- Right ToC column
- Better search UI
- Slightly better mobile design
- Code pasting support

### Future work

- Configurable top navbar items (at least link to GH and gem?)
- Configurable footer (same as the above?)
- Social media preview image support (it's not possible have nice static preview cards without an image)
- ???

## Preview

Link in https://github.com/ruby/rdoc/pull/1432#issuecomment-3395208790

### DIFF
--- a/.rdoc_options
+++ b/.rdoc_options
@@ -4,6 +4,8 @@ main_page: README.md
 autolink_excluded_words:
 - RDoc
 
+generator_name: aliki
+
 exclude:
 - AGENTS.md
 - CLAUDE.md


hsbt:
I prefer Aliki theme. I'm +1 to make this theme as default.

colby-swandale:
Same!

kou:
+1

mame:
Does this URL stay the same? If existing links, including fragments, continue to work well, it looks good :+1:

st0012:
@mame Thanks. And yes, the theme should not change any RDoc behaviour other than the visuals.

nobu:
I think it's better to make the blue letters in dark mode a little brighter.
In general, less contrast combination does not look good.

st0012:
@nobu I've updated the blue color in the dark mode so it has more contrast with the background. Let me know if it works, I'd also be happy to open/accept follow up PRs to improve it.

rmosolgo:
👋 I just noticed this new theme on the Ruby docs website. It is really nice, thank you for working on it!

st0012:
@rmosolgo Thank you for the kind words. I really appreciate it 😊

PR #1433, opened 2025-10-13
Improve search result by prioritizing exact matches

### LINKED ISSUE #1194: curious search result order on docs.ruby-lang.org
If you search for "String" on https://docs.ruby-lang.org/en/3.3/ the first four results returned are obscure:

Gem::SafeMarshal::Elements::String
Prism::StringFlags
Prism::StringNode
Reline::Terminfo::StringWithTiparm
String

Would it be possible to change it so that if one types an exact class name, the class of that name is first on the list? This would make it easier to use the search feature. More generally, the prioritization of these results is surprising.

I reported this at https://bugs.ruby-lang.org/issues/20722 and was directed here.  Apologies if this is the wrong place.

st0012:
I know adding 2 new phases for exact matches feels like a workaround, but without a more thorough review on the entire searching algorithm this is the best solution I have.

I also added basic tests for `search.js` using `mini_racer` gem (dev dependency) to simulate JS evaluation. I chose this approach because it should be simpler than maintaining a whole set of JS dependencies and setups.

Fixes #1194

### DIFF
--- a/lib/rdoc/generator/template/json_index/js/searcher.js
+++ b/lib/rdoc/generator/template/json_index/js/searcher.js
@@ -101,6 +101,7 @@ Searcher.prototype = new function() {
   function matchPassExactMethod(index, longIndex, queries, regexps) {
     var indexWithoutParens = index.replace(/\(\)$/, '');
     if (indexWithoutParens != queries[0]) return false;
+    if (index === indexWithoutParens) return false; // Not a method (no parens to remove)
     for (var i=1, l = regexps.length; i < l; i++) {
       if (!index.match(regexps[i]) && !longIndex.match(regexps[i]))
         return false;
@@ -215,12 +216,15 @@ Searcher.prototype = new function() {
     var togo = CHUNK_SIZE;
     var matchFunc, hltFunc;
 
+    var isLowercaseQuery = queries[0] === queries[0].toLowerCase();
+
     while (state.pass < 6 && state.limit > 0 && togo > 0) {
+      // When query is lowercase, prioritize methods over classes
       if (state.pass == 0) {
-        matchFunc = matchPassExact;
+        matchFunc = isLowercaseQuery ? matchPassExactMethod : matchPassExact;
         hltFunc = highlightQuery;
       } else if (state.pass == 1) {
-        matchFunc = matchPassExactMethod;
+        matchFunc = isLowercaseQuery ? matchPassExact : matchPassExactMethod;
         hltFunc = highlightQuery;
       } else if (state.pass == 2) {
         matchFunc = matchPassBeginning;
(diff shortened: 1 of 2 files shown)


lib/rdoc/generator/template/json_index/js/searcher.js
```diff
@@ -85,6 +85,35 @@ Searcher.prototype = new function() {
...
+  /*
+   * This record matches if both the index and longIndex exactly equal queries[0]
+   * and the record matches all of the regexps. This ensures top-level exact matches
+   * like "String" are prioritized over nested classes like "Gem::Module::String".
+   */
+  function matchPassExact(index, longIndex, queries, regexps) {
+    if (index != queries[0]) return false;
+    if (longIndex != queries[0]) return false;
+    for (var i=1, l = regexps.length; i < l; i++) {
+      if (!index.match(regexps[i]) && !longIndex.match(regexps[i]))
```
vinistock: If we're searching for an exact match, why do we still need to check these regexes?

test/rdoc/rdoc_generator_json_index_searcher_test.rb
```diff
@@ -0,0 +1,321 @@
...
+
+return if RUBY_DESCRIPTION =~ /truffleruby/ || RUBY_DESCRIPTION =~ /jruby/
+
+begin
+  require 'mini_racer'
+rescue LoadError
+  return
+end
+
+return unless defined?(MiniRacer::Context)
```
vinistock: Do we need this return considering we have the return above on `LoadError`?

lib/rdoc/generator/template/json_index/js/searcher.js
st0012: Yeah good catch. There can't be multiple regexps all matching both index and longIndex at the same time.

KieranP:
Another great improvement would be to order results by most common classes first. Consider https://docs.ruby-lang.org/en/master?q=empty%3F - The top 5 results are Array, Dir, ENV, File, and FileTest. Arguably, the most useful ordering would be String, Array, Hash, Set, and Symbol.

But I imagine that rdoc doesn't have a way to weight certain classes over others. And then there would probably be some contested discussions around how things should be weighted., so this may never happen, but I wanted to throw the idea out there, just in case :-)

st0012:
> But I imagine that rdoc doesn't have a way to weight certain classes over others.

Yeah the generated site is completely static and there's no way we can collect that info. We may get some insight if docs.ruby-lang.org collects some statistics. But we'd still have no ways to insert those weights without significantly changing RDoc as you said.

colby-swandale:
@st0012 I have list of boosts that I use for Ruby API, I would be interested to see if this is something RDoc could use in its search. 

https://github.com/rubyapi/rubyapi/blob/50cf34d2b96ff799a3cb75994edbf06172305ec9/app/models/ruby_object.rb#L2-L34

st0012:
@colby-swandale Thanks that's definitely helpful! I don't think this should be baked in RDoc directly given that it's technically a static site generator for all Ruby projects. 
But it may be possible to take a yaml in `.rdoc_options` list as weights 🤔

PR #1434, opened 2025-10-16
Add Herb linter and fix HTML and ERB linting issues

st0012:
- Added `Herb LSP` to suggested VS Code extension, which will surface ERB/HTML issues in `.rthml` templates.
- Fixed all surfaced issues
- Added ERB lint step on CI

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,10 +27,7 @@ jobs:
         bundler-cache: true
     - name: Run rubocop
       run: bundle exec rubocop
-    - name: Set up Node.js
-      uses: actions/setup-node@v6
-      with:
-        node-version: '20'
     - name: Run herb linter
+      # npx should be available by default so we don't need `setup-node` action
       # Don't lint erb files outside of lib as the current directory also includes vendor/bundle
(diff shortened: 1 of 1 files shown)


marcoroth:
FWIW, I think you could run `npx @herb-tools/linter "**/*.rhtml"` since the linter CLI accepts a glob pattern.

st0012:
@marcoroth Thanks! I've updated the PR to just use the linter

.github/workflows/lint.yml
```diff
@@ -27,3 +27,9 @@ jobs:
         bundler-cache: true
     - name: Run rubocop
       run: bundle exec rubocop
+    - name: Set up Node.js
+      uses: actions/setup-node@v4
```
olleolleolle: Perhaps the latest version of this Action's good, too? https://github.com/actions/setup-node/releases/tag/v6.0.0

marcoroth: I think you can even omit the `actions/setup-node` step, since `npx` is available by default in the base runner.

st0012:
Tests are failing because of various unrelated reasons 💀 
Merging this one first. Thanks again @marcoroth

PR #1435, opened 2025-10-19
Fix missing indentation on targeted .method-detail elements

dkibui:
This patch improves the spacing of highlighted method details in generated RDoc pages.

When a method anchor (for example, #method-i-force) is targeted, a left border is applied to the `.method-detail` element. Previously the border caused the inner content of the selected element to touch the left border, reducing readability.

This change adds a small left padding to the targeted element to maintain consistent spacing between the border and its contents.

Screenshot 2025-10-19 at 16 05 07
Screenshot 2025-10-19 at 16 04 43

### DIFF
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -582,6 +582,7 @@ main .method-detail {
 main .method-detail:target {
   margin-left: -10px;
   border-left: 10px solid var(--border-color);
+  padding-inline-start: 1em;
 }
 
 main .method-header {


kou:
If we use this approach, method entries aren't aligned:

image

How about this?

```diff
diff --git a/lib/rdoc/generator/template/darkfish/css/rdoc.css b/lib/rdoc/generator/template/darkfish/css/rdoc.css
index c84a604c..456d2a03 100644
--- a/lib/rdoc/generator/template/darkfish/css/rdoc.css
+++ b/lib/rdoc/generator/template/darkfish/css/rdoc.css
@@ -580,8 +580,9 @@ main .method-detail {
 }
 
 main .method-detail:target {
-  margin-left: -10px;
-  border-left: 10px solid var(--border-color);
+  margin-left: -15px;
+  border-left: 5px solid var(--border-color);
+  padding-left: 10px;
 }
 
 main .method-header {
```

image

tompng:
There is a related pull request https://github.com/ruby/rdoc/pull/1363 that restrict the highlight only to method signature part.
It solve method entries not-aligned problem in another way.

I think it's both fine adding indent or move highlight left, it's a design matter. (I think RDoc needs a designer)

kou:
Ah, sorry. I saw the pull request but I couldn't remember it.

st0012:
Thanks for the PR! Unfortunately I've addressed the issue in the upcoming new theme so this won't be necessary anymore 😅 

Screenshot 2025-11-18 at 22 34 24

PR #1442, opened 2025-11-01
searcher.js: handle missing RegExp.escape

byroot:
That API is more recent than I expected, so there is still a small minority of browsers that may be missing it.

But this escaping is only needed when the search context special chars, which too is a minority of searches.

Given that, I think it's better to just not escape if support is missing.

cc @st0012


st0012 (approved):
Thanks!

PR #1447, opened 2025-11-07
Add tooltip to attribute type

amomchilov:
1. Move the `[R]` part outside the underlined and linked `<a>` tag
2. Add a `title` field to explain what `[R]`, `[W]` or `[RW]` mean.

     It's fairly self-explanatory if you see `[RW]`, but on many APIs' docs, you'll only have a bunch of `[R]`s, so you'll be missing that context clue.

    On hover, it shows one of:
    * `attr_reader`
    * `attr_writer`
    * `attr_accessor`

| Before | After |
|--------|-------|
| Before | After |

### DIFF
--- a/lib/rdoc/generator/template/darkfish/class.rhtml
+++ b/lib/rdoc/generator/template/darkfish/class.rhtml
@@ -94,10 +94,8 @@
       <%- attributes.each do |attrib| %>
       <div id="<%= attrib.aref %>" class="method-detail anchor-link">
         <div class="method-heading attribute-method-heading">
-          <a href="#<%= attrib.aref %>" title="Link to this attribute">
-            <span class="method-name"><%= h attrib.name %></span>
-            <span class="attribute-access-type">[<%= attrib.rw %>]</span>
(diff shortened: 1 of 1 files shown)


lib/rdoc/generator/template/darkfish/class.rhtml
```diff
@@ -96,8 +96,8 @@
         <div class="method-heading attribute-method-heading">
           <a href="#<%= attrib.aref %>" title="Link to this attribute">
             <span class="method-name"><%= h attrib.name %></span>
```
amomchilov: FIXME: remove whitespace after this element

michaelherold:
The behavior of the `¶` is quite confusing with this change. It pops up when you hover over anything in the `div`, but appears to be associated with the `a`. I'm not really sure what its purpose is as there is no click/select behavior attached to it. My _guess_ is that it's there to make hovering the link more obvious.

I like the change, but think we should address the pilcrow's (TIL its name) behavior as part of it.

PR #1454, opened 2025-11-17
[Aliki] Set scroll-margin-top to all heading inside <main/>

tompng:
Before (header's opacity is set to 0.9 to describe)
before
After
after


kou (approved):
+1

PR #1455, opened 2025-11-17
[Aliki] Remove nav detail's nested padding-right

tompng:
Before:
padding-right is nested.
Nested `<details>` have narrow width.
before_1
before_2
before_3

After:
after

### DIFF
--- a/lib/rdoc/generator/template/aliki/css/rdoc.css
+++ b/lib/rdoc/generator/template/aliki/css/rdoc.css
@@ -864,11 +864,6 @@ nav a:hover {
   }
 }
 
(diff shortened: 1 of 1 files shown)


lib/rdoc/generator/template/aliki/css/rdoc.css
```diff
@@ -866,7 +866,6 @@ nav a:hover {
 
 nav ul li details {
   position: relative;
```
st0012: I discovered that this may not be needed either when reviewing this 🤔

kou (approved):
+1

PR #1456, opened 2025-11-17
[Aliki] Stable calculation of active toc

tompng:
Fix this unstable active table-of-contents link calculation shown in this video:
Clicking "Plugin types", active toc link will be "Option Parsing" or "File Parsing", depend on previous scroll position.
https://github.com/user-attachments/assets/bfd5c92a-9f11-4c9b-9267-86ae55607541

IntersectionObserver only notifies changed intersections.
We need to track which heading tag currently intersects with viewport. (I use Set in this pull request)
Use the top-most intersecting heading to make user-clicked toc links match with active toc links.

### DIFF
--- a/lib/rdoc/generator/template/aliki/js/aliki.js
+++ b/lib/rdoc/generator/template/aliki/js/aliki.js
@@ -242,7 +242,6 @@ function hookTocActiveHighlighting() {
 
     // Add active class to current link
     correspondingLink.classList.add('active');
-    activeLink = correspondingLink;
 
     // Scroll link into view if needed
     var tocNav = document.querySelector('#toc-nav');


lib/rdoc/generator/template/aliki/js/aliki.js
```diff
@@ -209,54 +209,66 @@ function generateToc() {
...
   };
 
-  var activeLink = null;
+  var intersectingHeadings = new Set();
+  function update() {
+    var firstIntersectingHeading = targetHeadings.find(function(heading) {
+      return intersectingHeadings.has(heading);
+    });
+    if (!firstIntersectingHeading) return;
+    var correspondingLink = document.querySelector('.toc-link[data-target="' + firstIntersectingHeading.id + '"]');
```
kou, line 235: This may be simpler:

```suggestion
    var correspondingLink = document.querySelector(`.toc-link[data-target="${firstIntersectingHeading.id}"]`);
```

lib/rdoc/generator/template/aliki/js/aliki.js
```diff
@@ -209,54 +209,66 @@ function generateToc() {
...
+    if (!correspondingLink) return;
+
+    // Remove active class from all links
+    tocLinks.forEach(function(link) {
+      link.classList.remove('active');
+    });
+
+    // Add active class to current link
+    correspondingLink.classList.add('active');
+    activeLink = correspondingLink;
```
kou: It seems that we don't need `activeLink`:

```suggestion
```

kou (approved):
+1

lib/rdoc/generator/template/aliki/js/aliki.js
tompng: I want to keep it for now.
Currently this file doesn't use ES6 feature: `const` `let` `()=>{}` and `` `${}` ``
I think we can use newer syntaxes, but I think it's better to do it in a separate pull request.

PR #1458, opened 2025-11-18
Fix rb_define_method(singleton_class_of_module, ...) not to change nested module name

### LINKED ISSUE #1440: Duplicate entries in "Class and Module Index"
These screenshots are from ruby/resolv.

Image

The opened trees are same too.
Image

I don't know what the trigger is, but I've sometimes seen the same problem for years.

tompng:
Fixes #1440
Always track class and singleton class stored to a variable to `@classes`.
Otherwise, `find_class var_name, class_name` called from `handle_method` will add a new class.

Before:
before
After:
after

Minimal reproduction:
```c
VALUE mFoo = rb_define_module("Foo");
VALUE mBar = rb_define_module_under(mFoo, "Bar");
VALUE mBarS = rb_singleton_class(mBar);
rb_define_method(mBarS, "baz", baz, 0);
```

`RDoc::Parser::C#handle_method` calls `find_class 'mBarS', 'Foo::Bar'`.
`RDoc::Parser::C#find_class` calls `@top_level.add_class RDoc::NormalClass, 'Foo::Bar'`. This will create `::Bar` for some reason and `Foo::Bar` disappears from generated document.

### DIFF
--- a/test/rdoc/parser/c_test.rb
+++ b/test/rdoc/parser/c_test.rb
@@ -781,6 +781,40 @@ void Init_Blah(void) {
     assert               methods.first.singleton
   end
 
+  def test_do_methods_nested_module_singleton_class
+    parser = util_parser <<~EOF
+      VALUE baz(VALUE klass, VALUE year) {
+      }
+      void Init_Foo(void) {
+        VALUE mFoo = rb_define_module("Foo");
+        VALUE mBar = rb_define_module_under(mFoo, "Bar");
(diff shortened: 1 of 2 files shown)


test/rdoc/parser/c_test.rb
```diff
@@ -781,6 +781,41 @@ def test_do_methods_singleton_class
     assert               methods.first.singleton
   end
 
+
```
kou:
```suggestion
```

kou (approved):
+1

PR #1459, opened 2025-11-19
[Aliki] Adjust table of content scrollable area height

tompng:
`aside.table-of-contents .toc-sticky nav` had `height: calc(100vh - var(--layout-header-height))`
but height should exclude padding and height of `<h3>On This Page</h3>` element.
This calculation is hard, so I use display:flex instead.

Left: before, Right: after
before_after

Scenario: scroll to page bottom
Left: before, Right: after
beforeafter_scrollbottom

### DIFF
--- a/lib/rdoc/generator/template/aliki/css/rdoc.css
+++ b/lib/rdoc/generator/template/aliki/css/rdoc.css
@@ -1226,6 +1226,18 @@ aside.table-of-contents * {
   outline: none !important;
 }
 
+aside.table-of-contents .toc-sticky {
+  display: flex;
+  flex-direction: column;
+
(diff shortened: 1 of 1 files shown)


lib/rdoc/generator/template/aliki/css/rdoc.css
```diff
@@ -1226,6 +1226,16 @@ aside.table-of-contents * {
   outline: none !important;
 }
 
+aside.table-of-contents .toc-sticky {
+  display: flex;
+  flex-direction: column;
+  height: calc(100vh - var(--layout-header-height) - var(--space-8) * 2);
```
st0012, line 1234: Can you add a comment explaining this calculation?

PR #1460, opened 2025-11-19
[Aliki] Remove unnecessary css variable declaration & usages

st0012:
Mostly removing unnecessary intermediate css variable declarations.

### DIFF
--- a/lib/rdoc/generator/template/aliki/css/rdoc.css
+++ b/lib/rdoc/generator/template/aliki/css/rdoc.css
@@ -59,10 +59,9 @@
   --color-background-primary: #ffffff;
   --color-background-secondary: var(--color-neutral-50);
   --color-background-tertiary: var(--color-neutral-100);
-  --color-background-elevated: #ffffff;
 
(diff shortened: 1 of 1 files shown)


lib/rdoc/generator/template/aliki/css/rdoc.css
```diff
@@ -151,22 +158,6 @@
   --z-fixed: 300;
   --z-modal: 400;
   --z-popover: 500;
-
-  /* Legacy Compatibility (maps to new tokens) */
-  --sidebar-width: var(--layout-sidebar-width);
```
tompng, line 156: `var(--sidebar-width)` is still used in L707. I think it should be changed to `var(--layout-sidebar-width)`
```css
@media (max-width: 1023px) {
  nav {
    ...
    width: var(--sidebar-width);
  }
  ...
}
```

st0012:
FWIW, I'm looking to add a linter for CSS, which should make similar cleanup easier in the future.

PR #1463, opened 2025-11-19
[Aliki] Add stylelint and update Aliki's css according to suggestions

st0012:


package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "rdoc",
+  "version": "1.0.0",
+  "description": "- GitHub: [https://github.com/ruby/rdoc](https://github.com/ruby/rdoc) - Issues: [https://github.com/ruby/rdoc/issues](https://github.com/ruby/rdoc/issues)",
```
kou: Do we need link markup here?

```suggestion
  "description": "- GitHub: https://github.com/ruby/rdoc - Issues: https://github.com/ruby/rdoc/issues",
```

package.json
```diff
@@ -0,0 +1,33 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "lint:css": "stylelint 'lib/rdoc/generator/template/**/*.css'"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ruby/rdoc.git"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
kou: Ruby?

```suggestion
  "license": "Ruby",
```

st0012:
@kou Sorry that I forgot to clean up `package.json`. Since we only need it to manage dev dependencies, we don't need to include those fields anyway.

PR #1465, opened 2025-11-22
[Aliki] Improve header links

st0012:
1. Remove `top` links next to the headers. This is a bad way to provide go-to-top feature. If users want it, we should do it differently in Aliki.
2. Make headers directly linkable, instead of using the pilcrow icon. This will make the code simpler and IMO more natural.


tompng (approved):
:+1:

PR #1468, opened 2025-11-23
Small refactors

st0012:


lib/rdoc/code_object/top_level.rb
```diff
@@ -40,6 +50,14 @@ def initialize(absolute_name, relative_name = absolute_name)
     @relative_name = relative_name
     @parser        = nil
 
+    if relative_name
+      @base_name = File.basename(relative_name)
+      @page_name = @base_name.sub(/\.(rb|rdoc|txt|md)$/i, '')
```
kou: `\z` is better because `$` matches end of new lines:

```suggestion
      @page_name = @base_name.sub(/\.(rb|rdoc|txt|md)\z/i, '')
```

PR #1470, opened 2025-11-23
Use Prism instead of Ripper for Ruby source tokenization

st0012:


lib/rdoc/parser/ripper_state_lex.rb
```diff
@@ -1,27 +1,59 @@
...
+  EXPR_LABEL = 0x400  # 1024 - Label in hash literal
+
+  REDEFINABLE_OPERATORS = %w[! != !~ % & * ** + +@ - -@ / < << <= <=> == === =~ > >= >> [] []= ^ ` | ~].freeze
 
-  class InnerStateLex < Ripper::Filter
-    def initialize(code)
-      super(code)
+  # Returns tokens parsed from +code+.
+  def self.parse(code)
+    lex = self.new(code)
```
kou, line 21: `self.` is redundant here.

```suggestion
    lex = new(code)
```

lib/rdoc/parser/ripper_state_lex.rb
```diff
@@ -1,27 +1,59 @@
...
+    tokens = []
+    begin
+      while tk = lex.get_squashed_tk
+        tokens.push tk
+      end
+    rescue StopIteration
     end
```
kou, line 28: It seems that we can remove `begin`/`rescue StopIteration`/`end` here because 2b96a8b617c8d3438ccf9bdb302353c66142afca removed `Enumerator`.

lib/rdoc/parser/ripper_state_lex.rb
```diff
@@ -232,36 +239,31 @@ def get_squashed_tk
...
-  private def get_op_tk(tk)
-    redefinable_operators = %w[! != !~ % & * ** + +@ - -@ / < << <= <=> == === =~ > >= >> [] []= ^ ` | ~]
-    if redefinable_operators.include?(tk[:text]) and tk[:state] == EXPR_ARG then
-      tk[:state] = Ripper::Lexer::State.new(EXPR_ARG)
+  def get_op_tk(tk)
+    if REDEFINABLE_OPERATORS.include?(tk[:text]) and tk[:state] == EXPR_ARG
+      tk[:state] = EXPR_ARG
       tk[:kind] = :on_ident
-    elsif tk[:text] =~ /^[-+]$/ then
+    elsif tk[:text] =~ /^[-+]$/
```
kou, line 263:
```suggestion
    elsif tk[:text] =~ /\A[-+]\z/
```

kou (approved):
+1

tompng:
I think creating a new syntax highlighter based on `Prism.lex` is another option.
Syntax highlighting doesn't need state bits. These are used in `rdoc/parser/ruby.rb` to parse ruby code.
I think `ripper_state_lex.rb` is a bit complicated to be used as a prototype of prism based syntax highlighter

PR #1467, opened 2025-11-22
Highlight C code

st0012:
This PR adds C code highlighting in both code blocks and method source code:

Screenshot 2025-11-22 at 21 37 21

Screenshot 2025-11-22 at 21 15 24

PR #1471, opened 2025-11-23
[Aliki] Add C syntax highlighting with a custom JS highlighter

st0012:
Does exactly the same as #1467 but using JS as this way it doesn't slowdown `ruby/ruby`'s doc generation.

Screenshot 2025-11-25 at 11 16 18

### DIFF
--- a/lib/rdoc/generator/template/aliki/js/c_highlighter.js
+++ b/lib/rdoc/generator/template/aliki/js/c_highlighter.js
@@ -1,0 +1,299 @@
+/**
+ * Client-side C syntax highlighter for RDoc
+ */
+
+(function() {
+  'use strict';
+
+  // C control flow and storage class keywords
+  const C_KEYWORDS = new Set([
+    'auto', 'break', 'case', 'continue', 'default', 'do', 'else', 'extern',
+    'for', 'goto', 'if', 'inline', 'register', 'return', 'sizeof', 'static',
+    'switch', 'while',
+    '_Alignas', '_Alignof', '_Generic', '_Noreturn', '_Static_assert', '_Thread_local'
+  ]);
+
+  // C type keywords and type qualifiers
+  const C_TYPE_KEYWORDS = new Set([
+    'bool', 'char', 'const', 'double', 'enum', 'float', 'int', 'long',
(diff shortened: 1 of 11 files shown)


lib/rdoc/generator/template/aliki/js/c_highlighter.js
```diff
@@ -0,0 +1,293 @@
...
+
+  // C keywords
+  const C_KEYWORDS = new Set([
+    'auto', 'break', 'case', 'char', 'const', 'continue', 'default', 'do',
+    'double', 'else', 'enum', 'extern', 'float', 'for', 'goto', 'if',
+    'inline', 'int', 'long', 'register', 'restrict', 'return', 'short',
+    'signed', 'sizeof', 'static', 'struct', 'switch', 'typedef', 'union',
+    'unsigned', 'void', 'volatile', 'while',
+    '_Alignas', '_Alignof', '_Atomic', '_Bool', '_Complex', '_Generic',
+    '_Imaginary', '_Noreturn', '_Static_assert', '_Thread_local'
```
tompng: I think it's worth adding `bool` `true` `false`

lib/rdoc/generator/template/aliki/js/c_highlighter.js
```diff
@@ -0,0 +1,293 @@
...
+  const C_TYPES = new Set([
+    'VALUE', 'ID', 'size_t', 'ssize_t', 'ptrdiff_t', 'uintptr_t', 'intptr_t',
+    'uint8_t', 'uint16_t', 'uint32_t', 'uint64_t',
+    'int8_t', 'int16_t', 'int32_t', 'int64_t',
+    'FILE', 'DIR', 'va_list'
+  ]);
+
+  // Common Ruby VALUE macros
+  const RUBY_MACROS = new Set([
+    'Qtrue', 'Qfalse', 'Qnil', 'Qundef', 'NULL'
```
tompng: `TRUE` and `FALSE` are also frequently used macro.
(It's already handled in `/^[A-Z][A-Z0-9_]*$/.test(word)`, but perhaps it could be added just like NULL)

Issue #1479, opened 2025-11-29
Aliki Feedback


Earlopain:
I'm playing around a bit with the new theme on https://docs.ruby-lang.org/en/master/ and have some thoughts I wanted to share. I like the theme toggle (dark mode specifically) and how the right side of my monitor is being used now. The search has more space to breath and doesn't take over the sidebar, with is nice.

Still, some complains/bugs:

Nested sections in the TOC are flattenend: 

<img width="289" height="602" alt="Image" src="https://github.com/user-attachments/assets/56a4df42-89b5-4f25-ad56-f83a35ce556d" />

<img width="266" height="513" alt="Image" src="https://github.com/user-attachments/assets/1b539002-af41-42e9-927b-87e392a1dc0e" />

I guess it looses formatting since String is supposed to be `code`.

I'm also not particularly liking how the left sidebar now always contains the top-level pages. I will pretty much always scroll past that to get to the more important information. Perhaps that can be moved above "Class and Module Index", so that class-specific stuff comes first?

On mobile, the hamburger menu is selectable which is annoying when rapidly opening/closing it. I do that expecting the search to be there but it's actually always visible now (which I like!)

On mobile, the "On This Page" section is missing, or I at least didn't manage to find it.

On mobile, the hamburger menu starts out expanded and closes on page load so it always flashes briefly for me. Depends on network speed I guess. Sometimes it also closes immediatly after I click, although I'm not sure how to reproduce.

All that said, I'm a fan of how it looks now

st0012:
Thanks for the clear feedback 🙂 

> Nested sections in the TOC are flattenend:

Yeah it'd be great to have it nested. I just don't have time to prioritize it yet. Do you mind opening an issue for this? PR is welcome too

> I'm also not particularly liking how the left sidebar now always contains the top-level pages. I will pretty much always scroll past that to get to the more important information. Perhaps that can be moved above "Class and Module Index", so that class-specific stuff comes first?

Yeah I'm working on a solution for it. Currently I plan to make it collapse-able, and automatically collapse it outside of the index page.

I want to make the page docs more useful overtime, so I want to make it easily accessible.

> On mobile, the hamburger menu is selectable which is annoying when rapidly opening/closing it.

Can you open a PR to help address this 🙂 

> On mobile, the "On This Page" section is missing, or I at least didn't manage to find it.

[...]

PR #1480, opened 2025-11-30
Aliki: Make hamburger toggle non-selectable

Earlopain:
It's implemented via an utf8 character but should not be selected by double-clicking

Ref https://github.com/ruby/rdoc/issues/1479

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -7,6 +7,7 @@
   "rules": {
     "csstools/value-no-unknown-custom-properties": true,
     "custom-property-no-missing-var-function": true,
-    "no-descending-specificity": null
+    "no-descending-specificity": null,
+    "property-no-vendor-prefix": null
   }
 }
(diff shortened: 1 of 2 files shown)


st0012:
I'm not sure if I misunderstood the desired behavior, but the preview site's hamburger icon is still selectable?

image

Earlopain:
Ugh, it needs to be prefixed for some browsers. I will fix this tomorrow

Earlopain:
I added fallback for safari, which I guess you were using. I don't know what your browser targets are, here's the browser support: https://caniuse.com/user-select-none. Some older versions of firefox also have their own prefix but usage is very low, so probably not worth it? Same for internet explorer.

Seems like stylelint doesn't like it though?

Earlopain:
> If the prefix is unavoidable

Yeah, it is ( think). I guess stylelint expects you to do some postprocessing on the files so this happens automatically. I disabled the rule

PR #1481, opened 2025-11-30
Aliki: Add padding to TOC

Earlopain:
I'm unsure if rdoc uses this somewhere but here is how it looks like on https://docs.ruby-lang.org/en/master/String.html

image

In https://github.com/ruby/rdoc/issues/1479 I wrote about it not being formatted. Since the TOC is generated via javascript, that seems a bit difficult to fix.

But I'm unsure if that is even needed. What was bothering me was that there was no connection between the "What's Here" and its subsections. This fixes that in a pretty simple way.

### DIFF
--- a/lib/rdoc/generator/template/aliki/css/rdoc.css
+++ b/lib/rdoc/generator/template/aliki/css/rdoc.css
@@ -1316,6 +1316,14 @@ aside.table-of-contents .toc-sticky nav {
   height: auto;
 }
 
+aside.table-of-contents .toc-list > .toc-h2 {
+  margin-left: var(--space-4);
+}
+
+aside.table-of-contents .toc-list > .toc-h3 {
+  margin-left: var(--space-8);
+}
+
 /* Hide TOC on mobile/tablet */
 @media (width <= 1279px) {
   aside.table-of-contents {


lib/rdoc/generator/template/aliki/css/rdoc.css
```diff
@@ -1316,6 +1316,10 @@ aside.table-of-contents .toc-sticky nav {
   height: auto;
 }
 
+aside.table-of-contents .toc-list > .toc-h3 {
+  margin-left: var(--space-5);
+}
+
```
st0012, line 1326: Since ToC shows h1~h3, I think we can have indentation for h2 too

Image

```suggestion
aside.table-of-contents .toc-list > .toc-h2 {
  margin-left: var(--space-4);
}

aside.table-of-contents .toc-list > .toc-h3 {
  margin-left: var(--space-8);
}

```

Earlopain: darkfish doesn't add the h1 from `class String` on https://docs.ruby-lang.org/en/master/String.html. I don't know by which rules, on https://docs.ruby-lang.org/en/3.4/standard_library_md.html the first h1 is present. Maybe simple prefix check?

That confused me a bit. Anyways, I'm not opposed to this, I think it looks fine.

PR #1483, opened 2025-12-03
Apply markup in table cell

ksss:
In the documentation for methods like `test`, backtick are used within table cells.
While each parser handles backquotes within table cells, the generator does not.
As a result, the tags are displayed literally without being processed.

To resolve this, I propose that tag expressions should be properly handled within table cells as well.

### Before
ri --format=bs test

```
 Character  |Test
------------|---------------------------------------------------------------
<tt>'<'</tt>|Whether the `mtime` at `path0` is less than that at `path1`.
<tt>'='</tt>|Whether the `mtime` at `path0` is equal to that at `path1`.
<tt>'>'</tt>|Whether the `mtime` at `path0` is greater than that at `path1`.
```

ri --format=markdown test

```
 Character  |Test
------------|---------------------------------------------------------------
<tt>'<'</tt>|Whether the `mtime` at `path0` is less than that at `path1`.
<tt>'='</tt>|Whether the `mtime` at `path0` is equal to that at `path1`.
<tt>'>'</tt>|Whether the `mtime` at `path0` is greater than that at `path1`.
```

ri --format=ansi

image

### After

ri --format=bs test

```
Character|Test
---------|---------------------------------------------------------------
   '<'   |Whether the `mtime` at `path0` is less than that at `path1`.
   '='   |Whether the `mtime` at `path0` is equal to that at `path1`.
   '>'   |Whether the `mtime` at `path0` is greater than that at `path1`.
```

ri --format=markdown test

```
Character|Test
---------|---------------------------------------------------------------
  `'<'`  |Whether the `mtime` at `path0` is less than that at `path1`.
  `'='`  |Whether the `mtime` at `path0` is equal to that at `path1`.
  `'>'`  |Whether the `mtime` at `path0` is greater than that at `path1`.
```

ri --format=ansi

image

### DIFF
--- a/lib/rdoc/markup/to_ansi.rb
+++ b/lib/rdoc/markup/to_ansi.rb
@@ -81,6 +81,10 @@ class RDoc::Markup::ToAnsi < RDoc::Markup::ToRdoc
     end
   end
 
+  def calculate_text_width(text)
+    text.gsub(/\e\[[\d;]*m/, '').size
+  end
+
   ##
   # Starts accepting with a reset screen
 
--- a/lib/rdoc/markup/to_bs.rb
+++ b/lib/rdoc/markup/to_bs.rb
@@ -65,6 +65,10 @@ class RDoc::Markup::ToBs < RDoc::Markup::ToRdoc
     end
   end
 
+  def calculate_text_width(text)
+    text.gsub(/_\x08/, '').gsub(/\x08./, '').size
+  end
+
   ##
   # Turns on or off regexp handling for +convert_string+
 
--- a/lib/rdoc/markup/to_rdoc.rb
+++ b/lib/rdoc/markup/to_rdoc.rb
@@ -250,13 +250,10 @@ class RDoc::Markup::ToRdoc < RDoc::Markup::Formatter
   # Adds +table+ to the output
 
   def accept_table(header, body, aligns)
-    body = body.map do |row|
-      row.map do |cell|
-        attributes cell
-      end
-    end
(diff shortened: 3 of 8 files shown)


ksss:
@kou 
You are absolutely right.
How can I implement `ToAnsi#accept_table`?

tompng:
I think `RDoc::Markup::ToRdoc` can have a method like `calculate_text_width(text)` and use it in `accept_table`.

```ruby
widths = header.zip(*body).map do |cols|
  cols.map(&:size).max # Use `calculate_text_width instead of size
end
...
# a is :rjust | :ljust | :center, w is width
t.__send__(a, w)
↓
t.__send__(a, w + calculate_text_width(t) - t.size)
```

And each formatter can override it if it needs special width calculation logic.

### ToRDoc#calculate_text_width
Simply return `text.size` for now. Maybe there's a room for improvement in the future such as treating some multibyte character as 2-width.

### ToAnsi#calculate_text_width
Remove escape sequences before calculating width

### ToBs#calculate_text_width
I think ToBS also need another custom width calculation logic.
`"<b>dd</b>"` becomes `"d\bdd\bd"` and its width should be 2.
`"d" backspace "d"` will print `"d"` twice in the same position. (Printing a character to a paper by ink twice makes the text darker/bold)

ksss:
@tompng 
Thank you for your wonderful suggestion!
I also learned about the BS format, which I was unaware of.
I have reimplemented it using the proposed method and updated the description.

PR #1487, opened 2025-12-07
Fix source code styling issues

st0012:
- By setting width to 100%, it ensures the source code container takes up
  the full width of its parent container.
- By setting box-sizing to border-box, it ensures when the source code block
  expands, it doesn't overflow the container.
- By limiting the method header's width, it ensures that the source code toggle
  doesn't get pushed down to a weird position.

### DIFF
--- a/lib/rdoc/generator/template/aliki/css/rdoc.css
+++ b/lib/rdoc/generator/template/aliki/css/rdoc.css
@@ -1119,6 +1119,8 @@ main .method-source-code {
 
 main .method-source-code pre {
   border-color: var(--color-accent-hover);
+  width: 100%;
+  box-sizing: border-box;
 }
 
 main .method-source-code.active-menu {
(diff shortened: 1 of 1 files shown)


lib/rdoc/generator/template/aliki/css/rdoc.css
```diff
@@ -1143,6 +1145,7 @@ main .method-detail:target {
 
 main .method-header {
   display: inline-block;
+  max-width: 85%;
```
tompng: Width of `"▶︎ Source"` depends on font size.
Does it make sense to use something like `calc(100% - 6em)`?

tompng (approved):

lib/rdoc/generator/template/aliki/css/rdoc.css
st0012: Thanks 👍

Issue #1335, opened 2025-03-30
Source link does not work for Hash@has_value?


BurdetteLamar:
At https://docs.ruby-lang.org/en/master/Hash.html#method-i-has_value-3F, clicking the `Source` link only re-orients the arrow;  does not display the source code.

tompng:
There are two separate problems.
1. Inconsistency of source link and actual source DOM existence in the generated HTML file
2. `Hash#value?` and `Hash#has_value?` are both considered `skip_description? == true`, thus, document is completely ignored.

PR #1488, opened 2025-12-08
Fix method source code not created if a method's description is skipped

extern-c:
This PR resolves https://github.com/ruby/rdoc/issues/1335.

### DIFF
--- a/lib/rdoc/generator/template/aliki/class.rhtml
+++ b/lib/rdoc/generator/template/aliki/class.rhtml
@@ -156,15 +156,13 @@
               <summary>Source</summary>
             </details>
           </div>
+          <div class="method-source-code" id="<%= method.html_name %>-source">
+            <pre class="<%= method.source_language %>" data-language="<%= method.source_language %>"><%= method.markup_code %></pre>
+          </div>
         <%- end %>
 
         <%- unless method.skip_description? then %>
         <div class="method-description">
-          <%- if method.token_stream then %>
-          <div class="method-source-code" id="<%= method.html_name %>-source">
-            <pre class="<%= method.source_language %>" data-language="<%= method.source_language %>"><%= method.markup_code %></pre>
-          </div>
-          <%- end %>
           <%- if method.mixin_from then %>
             <div class="mixin-from">
               <%= method.singleton ? "Extended" : "Included" %> from <a href="<%= klass.aref_to(method.mixin_from.path) %>"><%= method.mixin_from.full_name %></a>


lib/rdoc/generator/template/aliki/class.rhtml
```diff
@@ -156,15 +156,13 @@
               <summary>Source</summary>
             </details>
           </div>
+          <div class="method-source-code" id="<%= method.html_name %>-source">
+            <pre class="<%= method.source_language %>" data-language="<%= method.source_language %>"><%= method.markup_code %></pre>
+          </div>
```
tompng, line 161: Moving this will change this HTML structure:
```html
<div class="method-heading"></div>
<div class="method-controls"></div>
<div class="method-description"></div>
```

How about moving the whole `<%- if method.token_stream %> method controls <% -end %>` to 

```erb
<%- unless method.skip_description? then %>
(HERE)
<div class="method-description">
  ...
</div>
...
<%- end %>
```

st0012: I'm aware of this but it also feels like source code div actually shouldn't be placed under description, if we want their display to be separate?

tompng: > if we want their display to be separate

I think there is no need to separate. If there's a reason to skip description, (example: it's an alias and the description/source are both visible in the original method), we don't need to show method source.
~This will reduce generated HTML size especially when many aliases are defined.~ (edited. this part seems wrong)

For `Hash#has_value?`, it shouldn't skip description. #1491 will fix it from the backend side.

tompng: Looks like `method.skip_description?` and `method.token_stream.nil?` is almost identical (especially after some bug fixes of backend),
I reconsidered it's OK moving method-description outside of `<%- unless method.skip_description? then %>` :+1:

extern-c: Hi @st0012 and @tompng,

Thanks for the code review! For my own understanding, would you mind confirming when  `method.skip_description?` is supposed to return `true`?

tompng: > when method.skip_description? is supposed to return true?

In the source code document
```ruby
# Whether to skip the method description, true for methods that have
# aliases with a call-seq that doesn't include the method name.
def skip_description?
```

Here's an example:
```c
/*
 * call-seq:
 *   foo() -> integer
 *   bar() -> integer
 * Returns 42
 */
VALUE rb_obj_foo(VALUE self){return INT2FIX(42);}
void Init_FooBar() {
  rb_define_method(rb_cObject, "foo", rb_obj_foo);
  rb_define_method(rb_cObject, "bar", rb_obj_foo);
  rb_define_method(rb_cObject, "baz", rb_obj_foo);
}
```
The same method is defined as `foo`, `bar` and `baz`. But call-seq only have `foo` and `bar`.
In this case, description for `baz` is skipped.

Actual description skipped method: https://docs.ruby-lang.org/en/master/Complex.html#method-i-angle

In other words, minor aliases that is not so worth documented, is skipped.
On the other hand, `token_stream` seems to be linked to the most major method.
If the most major method is skipped == treated as most minor, something should be wrong.

tompng:
Thank you :+1:

lib/rdoc/generator/template/aliki/class.rhtml
extern-c: @tompng thanks for the detailed explanation. I had noticed some inconsistencies which you clarified and documented  here: https://github.com/ruby/rdoc/issues/1492.

PR #1491, opened 2025-12-08
Fix call-seq deduplicate for name overlap in aliases

tompng:
Backend fix of #1335
Fix call-seq deduplicate of `Hash#value?` and `Hash#has_value?`.
When alias names have overlap, deduplicate didn't work correctly.

### DIFF
--- a/lib/rdoc/code_object/any_method.rb
+++ b/lib/rdoc/code_object/any_method.rb
@@ -351,7 +351,6 @@ class RDoc::AnyMethod < RDoc::MethodAttr
     return call_seq unless is_alias_for || !aliases.empty?
 
     method_name = self.name
-    method_name = method_name[0, 1] if method_name =~ /\A\[/
 
     entries = call_seq.split "\n"
 
@@ -360,15 +359,24 @@ class RDoc::AnyMethod < RDoc::MethodAttr
       ignore << is_alias_for.name
       ignore.concat is_alias_for.aliases.map(&:name)
     end
-    ignore.map! { |n| n =~ /\A\[/ ? /\[.*\]/ : n}
-    ignore.delete(method_name)
-    ignore = Regexp.union(ignore)
 
-    matching = entries.reject do |entry|
-      entry =~ /^\w*\.?#{ignore}[$\(\s]/ or
-        entry =~ /\s#{ignore}\s/
+    ignore.delete(method_name)
+    ignore_bracket_methods, ignore_other_methods = ignore.partition {|i| i.start_with?('[') }
+
+    if ignore_other_methods.any?
+      ignore_union = Regexp.union(ignore_other_methods)
+      entries.reject! do |entry|
+        /\A(?:\w*\.)?#{ignore_union}(?:[(\s]|\z)/.match?(entry) ||
+          /\s#{ignore_union}\s/.match?(entry)
+      end
+    end
+    if ignore_bracket_methods.any?
+      entries.reject! do |entry|
+        # Ignore `receiver[arg] -> return_type` `[](arg)` `[]`
+        /\A\w*\[.*?\](?:[(\s]|\z)/.match?(entry)
+      end
     end
 
-    matching.empty? ? nil : matching.join("\n")
+    entries.empty? ? nil : entries.join("\n")
   end
 end
--- a/test/rdoc/parser/c_test.rb
+++ b/test/rdoc/parser/c_test.rb
@@ -2107,6 +2107,57 @@ str.slice(start, length) -> new_str or nil
     assert_match 'str === obj', equals3.call_seq
   end
 
+  def test_scan_method_copy_bracket_name
+    parser = util_parser <<~C
+      /*
+      *  call-seq:
+      *    obj[index] -> nil
+      *    obj.aref(index) -> nil
+      */
+      static VALUE
+      rb_obj_aref(VALUE obj, VALUE index) { return Qnil; }
+
+      Init_Blah(void)
+      {
+        rb_define_method(rb_cObject, "[]", rb_obj_aref, 1);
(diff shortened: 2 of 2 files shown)


lib/rdoc/code_object/any_method.rb
```diff
@@ -360,15 +360,25 @@ def deduplicate_call_seq(call_seq)
...
     ignore.delete(method_name)
-    ignore = Regexp.union(ignore)
 
-    matching = entries.reject do |entry|
-      entry =~ /^\w*\.?#{ignore}[$\(\s]/ or
-        entry =~ /\s#{ignore}\s/
+    if ignore.any?
+      ignore_union = Regexp.union(ignore)
+      entries.reject! do |entry|
+        entry =~ /^(\w*\.)?#{ignore_union}([\(\s]|$)/ or
```
tompng: Main change
`/^\w*\.?#{ignore_union}[$\(\s]/`
↓
`/^(\w*\.)?#{ignore_union}([\(\s]|$)/`

`^\w*\.?value?[$\(\s]` matches to `has_value?`, but we only want to ignore `value?` and `receiver.value?`
Fix is: `\w*\.?` → `(\w*\.)?`

The latter part `[$\(\s]`, `$` seems like a mistake of `$` (end-of-line), not `"$"` (dollar character)

tompng: Unfortunately, with this change, `receiver[arg]` won't be handled here.
So I added bracket_method remove logic in line 375-379.

lib/rdoc/code_object/any_method.rb
```diff
@@ -360,15 +360,25 @@ def deduplicate_call_seq(call_seq)
...
+          entry =~ /\s#{ignore_union}\s/
```
kou:
```suggestion
        /^(\w*\.)?#{ignore_union}([\(\s]|$)/.match?(entry) or
          \s#{ignore_union}\s/.match?(entry)
```

BTW:

* Can we use `\A`/`\z` not `^`/`$` here?
* Can we remove needless `\` escape from `[\(]`?
* How about using `(?:...)` not `(...)` not referred groups?

lib/rdoc/code_object/any_method.rb
```diff
@@ -360,15 +360,25 @@ def deduplicate_call_seq(call_seq)
...
+      end
+    end
+    if bracket_methods.any?
+      entries.reject! do |entry|
+        # Ignore `receiver[arg] -> return_type` `[](arg)` `[]`
+        entry.match?(/^\w*\[.*\]([\(\s]|$)/)
```
kou:
```suggestion
        /\A\w*\[.*?\](?:[(\s]|\z)/.match?(entry)
```

lib/rdoc/code_object/any_method.rb
kou: The `$\(\s]` pattern was added by https://github.com/ruby/rdoc/pull/897 .

> The latter part `[$\(\s]`, `$` seems like a mistake of `$` (end-of-line), not `"$"` (dollar character)

I think that the assumption is correct.

lib/rdoc/code_object/any_method.rb
```diff
@@ -360,15 +360,25 @@ def deduplicate_call_seq(call_seq)
       ignore << is_alias_for.name
       ignore.concat is_alias_for.aliases.map(&:name)
     end
-    ignore.map! { |n| n =~ /\A\[/ ? /\[.*\]/ : n}
+
+    bracket_methods = ignore.grep(/\A\[/)
+    ignore -= bracket_methods
```
kou: How about prefixing `ignore_` to `bracket_methods` for readability?

```suggestion
    ignore_bracket_methods, ignore_other_methods = ignore.partition {|i| i.start_with?("[")}
```

lib/rdoc/code_object/any_method.rb
tompng: done!

tompng:
These docs will be improved

Source and description will be available:
https://docs.ruby-lang.org/en/master/Array.html#method-i-find_index
https://docs.ruby-lang.org/en/master/Date.html#method-i-asctime
https://docs.ruby-lang.org/en/master/Date.html#method-i-mday
https://docs.ruby-lang.org/en/master/Hash.html#method-i-has_value-3F
https://docs.ruby-lang.org/en/master/Time.html#method-i-mday

Wrong call-seq (alias is wrongly used) will be fixed:
https://docs.ruby-lang.org/en/master/Symbol.html#method-i-next
https://docs.ruby-lang.org/en/master/JSON/Ext/Generator/State.html#method-i-escape_slash
https://docs.ruby-lang.org/en/master/JSON/Ext/Generator/State.html#method-i-script_safe-3F
https://docs.ruby-lang.org/en/master/JSON/Ext/Generator/State.html#method-i-strict-3F
https://docs.ruby-lang.org/en/master/JSON/Ext/Generator/State.html#method-i-escape_slash
https://docs.ruby-lang.org/en/master/JSON/Ext/Generator/State.html#method-i-to_hash

tompng:
> It may be better that we extract method names from "call_seq"

I agree. I think call-seq's name part and params part extraction needs refactor.
But I'd like to fix this change in regexp way for now.

`AnyMethod#name` already have a name extraction from call-seq, but the logic seems incomplete. It doesn't handle method like `foo!`, `foo?`.
```ruby
def name
    return @name if @name

    @name =
      @call_seq[/^.*?\.(\w+)/, 1] ||
      @call_seq[/^.*?(\w+)/, 1] ||
      @call_seq if @call_seq
  end
```
Maybe we can have a call-seq common format parser and use it in `deduplicate_call_seq`, `name`, `param_list` and `param_seq` but it will be a relatively large change.

PR #1495, opened 2025-12-14
Avoid linking file paths inside backticks

st0012:
Highlighting file paths like `lib/foo/bar.rb` should not be linked to the file automatically. If the writer's intention is to link it, `rdoc-ref` should be used.

### DIFF
--- a/lib/rdoc/markup/to_html_crossref.rb
+++ b/lib/rdoc/markup/to_html_crossref.rb
@@ -156,7 +156,7 @@ class RDoc::Markup::ToHtmlCrossref < RDoc::Markup::ToHtml
     ref = @cross_reference.resolve name, text if name
 
     case ref
-    when String then
+    when String
       if rdoc_ref && @options.warn_missing_rdoc_ref
         puts "#{@from_path}: `rdoc-ref:#{name}` can't be resolved for `#{text}`"
       end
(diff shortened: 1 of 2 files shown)


lib/rdoc/markup/to_html_crossref.rb
```diff
@@ -239,6 +239,9 @@ def convert_tt_crossref(flow_items, index)
     return unless RDoc::Markup::AttrChanger === closer
     return unless tt_tag?(closer.turn_off, true)
 
+    # Skip file paths - they should remain as literal code, not be linked
+    return if string.include?('/')
```
tompng: Another way to check if the cross-ref is a page link is `RDoc::TopLevel === ref`

How about adding an option like `allow_page` or something to `ToHtmlCrossRef#cross_reference` and `ToHtmlCrossRef#link` like this?

```ruby
def link(name, text, code = true, rdoc_ref: false, allow_page: true)
  ...
  ref = @cross_reference.resolve name, text if name
  return if !include_page && RDoc::TopLevel === ref
  ...
end
```

It can properly reject link to a page that doesn't have `/` such as `` `README.md` ``
(It seems to be called `page` in `RDoc::Store#page`)

lib/rdoc/markup/to_html_crossref.rb
```diff
@@ -156,16 +156,26 @@ def link(name, text, code = true, rdoc_ref: false)
...
 
-      if code and RDoc::CodeObject === ref and !(RDoc::TopLevel === ref)
-        text = "<code>#{CGI.escapeHTML text}</code>"
+      text = if code && RDoc::TopLevel === ref
+        # Allow explicit rdoc-ref links to files, but don't wrap in code tags
+        if rdoc_ref
+          text
+        else
+          # Don't auto-link file paths in backticks
+          return "<code>#{CGI.escapeHTML text}</code>"
```
tompng, line 173: If the README.md content is
```markdown
README `README`
```
Two `"README"` will both trigger `link("README", "README", true, false)`.
One from `handle_regexp_CROSSREF` → `cross_reference` → `link`
Another from `convert_tt_crossref` → `cross_reference` → `link`
So if we need to detect backticks, we need to pass a new argument from `convert_tt_crossref`.

st0012: I'm surprised that `README` without backticks would also enter here. But I guess it's the same mechanism linking `String` or `Array` automatically.
I kinda want to disable this in a breaking change release, perhaps in v8.0. WDYT?

tompng: I think it's a good idea to disable it.

tompng: For this pull request, we need to fix this.

```markdown
See README.md `README.md`
```
image
Only filenames inside backtick should use `<code>` but currently filenames outside backtick also converted to `<code>`.

PR #1502, opened 2025-12-15
Change the default theme to Aliki

st0012:


vinistock (approved):
Will we consider this a breaking change? Will doc generation break if someone has darkfish customizations?

tompng (approved):
:+1:

st0012:
> Will we consider this a breaking change?

Yes this will go out with v7.0.0 release

PR #1504, opened 2025-12-16
Rebuild Aliki's searching mechanism

### LINKED ISSUE #1497: Classes named like common methods are very difficult to search for
https://docs.ruby-lang.org/en/master/

If I want to look at docs for class `Hash` and search for it exactly, it's the 29th result (if I didn't misscount). That is probably because of https://github.com/ruby/rdoc/pull/1433, it priotizes exact method matches and there are many classes that defined custom `hash`. The same is true when searching for the `String` class, there are quite a few string methods that take precedence.

I don't know if the "solution" would be to prioritize exact classes even before exact methods. When searching for `Integer`, the `Integer` method comes before the integer class, so maybe? I don't think that method is what users are looking for.

Probably the whole search logic needs a rework tbh, it never felt "quite right". At the moment it is very unfortunate that such common classes are buried so far down in the search results.

st0012:
Rebuild Aliki's search mechanism with a custom search index and improved ranking algorithm.

Key changes:
- Aliki now generates its own search index (`js/search_data.js`) instead of using the shared JsonIndex generator (can be removed along with Darkfish in the future)
- New search ranking algorithm with better prioritization:
  - Lowercase queries prioritize methods; uppercase queries prioritize classes/modules/constants
  - Exact matches rank higher; shorter names preferred
  - Class methods rank above instance methods (I don't have strong opinion on this one)
- Search results now display type badges (class, module, const, method)
- Constants are now included in search results

Closes #1497 

Screenshot 2025-12-16 at 17 57 50

### DIFF
--- a/lib/rdoc/generator/template/aliki/js/search_ranker.js
+++ b/lib/rdoc/generator/template/aliki/js/search_ranker.js
@@ -7,7 +7,7 @@
  * 3. Match types:
  *    - Namespace queries (::) and method queries (# or .) match against full_name
  *    - Regular queries match against unqualified name
- *    - Prefix match (1000) > substring match (100) > fuzzy match (10)
+ *    - Prefix (10000) > substring (5000) > fuzzy (1000)
  * 4. First character determines type priority:
  *    - Starts with lowercase: methods first
  *    - Starts with uppercase: classes/modules/constants first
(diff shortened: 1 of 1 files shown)


lib/rdoc/generator/template/aliki/js/search_ranker.js
```diff
@@ -0,0 +1,223 @@
...
+  // "hash" query matching "Hash" class as exact match
+  if (isNamespaceQuery && fullNameLower === queryLower) {
+    return 1000000;
+  }
+
+  var matchScore = 0;
+
+  if (isNamespaceQuery) {
+    // For namespace queries like "Foo::B", match against full_name
+    if (!fullNameLower.startsWith(queryLower)) {
```
tompng: Query that contains `#` should also match against full_name.
For example, `Array#filter` should match to something but currently it doesn't.
(Search with `filter` in ruby/ruby's doc and press down arrow will fill `Array#filter` to the search form.)

lib/rdoc/generator/template/aliki/js/search_ranker.js
```diff
@@ -0,0 +1,223 @@
...
+      return null;
+    }
+    matchScore = 100;
+  } else {
+    // For regular queries, match against unqualified name (prefix match)
+    if (!nameLower.startsWith(queryLower)) {
```
tompng: I often search methods that includes query in IRB, like:
`Kernel.methods.grep /_defined\?/`

Can we include substring match? and prioritize lower than prefix match.

tompng:
I don't think the current search is bad, except for the search result order and constant not searched.
It has advanced search: `query = add_alias` match to `add_module_alias`.

How about first match against full_name with regexps generated by the old searcher, and then reorder the result with score?

st0012:
I've updated the implementation so now these are supported:
- `Array#fil` lists `Array#fill`, `Array#filter`, and `Array#filter!`
- `Array.try` lists `Array.try_convert`
- `_defined?` lists
    - Screenshot 2025-12-17 at 21 44 57
- `insmet` lists
    - Screenshot 2025-12-17 at 22 13 26

lib/rdoc/generator/template/aliki/js/search_ranker.js
```diff
@@ -0,0 +1,249 @@
...
+      return null;
+    }
+  } else {
+    // For regular queries, match against unqualified name
+    if (nameLower.startsWith(q.normalized)) {
+      matchScore = 1000;  // Prefix
+    } else if (nameLower.includes(q.normalized)) {
+      matchScore = 100;   // Substring
+    } else if (fuzzyMatch(nameLower, q.normalized)) {
+      matchScore = 10;    // Fuzzy
```
tompng: I think the difference of fuzzy match score and prefix/substring match score should be larger.

Fuzzy-match class method > Substring-match instance method (`alias_extension` > `do_aliases` for `q=aliases`)
because score diff(90) less than 500
https://6f5f33e5.rdoc-6cd.pages.dev?q=aliases

Fuzzy-match method > Prefix-match class name (`gemdirs` > `MissingFileError` for `q=mis`)
because score diff(900) less than 10000
https://6f5f33e5.rdoc-6cd.pages.dev?q=mis

PR #1508, opened 2025-12-18
Make sure path link in footers work correctly

### LINKED ISSUE #1498: Broken link in footer
Link text: `Documentation`
Link href: https://docs.ruby-lang.org/en/master/CGI/index.html

Should be? https://docs.ruby-lang.org/en/master/index.html

May be seen in `https://docs.ruby-lang.org/en/master/CGI/Escape.html` (and many others).

st0012:
Fixes #1498

### DIFF
--- a/lib/rdoc/generator/aliki.rb
+++ b/lib/rdoc/generator/aliki.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'uri'
+
 ##
 # Aliki theme for RDoc documentation
 #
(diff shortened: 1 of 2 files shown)


lib/rdoc/generator/aliki.rb
```diff
@@ -115,6 +117,21 @@ def write_search_index
     File.write search_index_path, "var search_data = #{JSON.generate(data)};"
   end
 
+  ##
+  # Resolves a URL for use in templates. Absolute URLs are returned unchanged.
+  # Relative URLs are prefixed with rel_prefix to ensure they resolve correctly from any page.
+
+  def resolve_url(rel_prefix, url)
+    uri = URI.parse(url)
+    if uri.absolute?
```
vinistock, line 126: Can a URI ever not be absolute? I thought parsing a URI for a relative path always failed.

st0012: Yeah you can create `URI` with a wide range of strings:

```
irb(main):005> URI.parse("/foo")
=> #<URI::Generic /foo>
irb(main):006> URI("/foo")
=> #<URI::Generic /foo>
irb(main):007> URI("foo")
=> #<URI::Generic foo>
irb(main):008> URI("foo").absolute?
=> false
irb(main):009> URI("/foo").absolute?
=> false
irb(main):010> URI("foobar?baz>").absolute?
=> false
```

vinistock: Ah, right. I think the limitation is with `file` URIs.

We use URIs for files so often for tooling stuff I forget there are other URIs 😅

PR #1509, opened 2025-12-18
[Aliki] Fix search result type styling on mobile

Earlopain:
Styling only applied to `#search-results`. I added a common class to both mobile and desktop to use instead. Also renamed `#search-results` to `#search-results-desktop` to avoid ambiguity with the new class.

Before:
image

After:
image


st0012 (approved):
Good catch. Thanks!

PR #1512, opened 2025-12-19
Fix search dropdown behaviour

st0012:
After this change:

| Scenario | Behavior |
|----------|----------|
| Type search, click result | Navigate without `?q=`, clean URL without dropdown |
| Type search, press Enter | Navigate without `?q=`, clean URL without dropdown |
| Land on URL with `?q=` param | Show dropdown with pre-filled query |
| Click outside dropdown | Hide dropdown, keep input value |
| Focus input with existing query | Show dropdown again |

### DIFF
--- a/lib/rdoc/version.rb
+++ b/lib/rdoc/version.rb
@@ -5,6 +5,6 @@ module RDoc
   ##
   # RDoc version you are using
 
-  VERSION = '7.0.1'
+  VERSION = '7.0.2'
 
 end


lib/rdoc/generator/template/aliki/js/aliki.js
```diff
@@ -97,15 +89,27 @@ function hookSearch() {
   const search = createSearchInstance(input, result);
   if (!search) return;
 
+  // Hide search results when clicking outside the search area
+  document.addEventListener('click', (e) => {
+    if (!e.target.closest('.navbar-search-desktop')) {
+      result.setAttribute('aria-expanded', 'false');
```
tompng: Up and down arrow keys still moves in the hidden search navigation.
I think `search.setNavigationActive(false)` `search.setNavigationActive(true)` are needed where changing `aria-expanded` attribute.

Adding `search.hide()` and `search.show()` that does both might be a good idea.

lib/rdoc/version.rb
```diff
@@ -5,6 +5,6 @@ module RDoc
   ##
   # RDoc version you are using
 
-  VERSION = '7.0.1'
+  VERSION = '7.0.2'
```
tompng, line 8: Maybe wrong commit pushed?
I'm fine cutting 7.0.2

lib/rdoc/generator/template/aliki/js/aliki.js
st0012: Ah good catch. Updated 👍

lib/rdoc/version.rb
st0012: No I just thought if we could merge the PR we can ship it immediately in 7.0.2 😛
This would still work tho. I'd love to cut a new release right after this is merged.

tompng: I see :+1:

PR #1515, opened 2025-12-20
Fix class/module alias document having wrong name

### LINKED ISSUE #1014: Incorrect module aliasing
File context.rb contains:
```ruby
module Mod1
end

module Mod2
  Mod3 = Mod1
end
```
After executing `rdoc context.rb`:
* the left "Class and Module Index" shows Mod3, Mod2, Mod3
* clicking on the first "Mod3" leads to "module Mod1"
* clicking on the second "Mod3" leads to "module Mod3"

What should have been generated:
* the left "Class and Module Index" shows Mod1, Mod2, Mod2::Mod3
* clicking on "Mod1" leads to "module Mod1"
* clicking on "Mod2::Mod3" leads to "module Mod1"

Cheers.

tompng:
Fixes #1014

Some aliased class/module document had wrong title.
Example: https://docs.ruby-lang.org/en/master/Net/HTTP.html
Title should be `class Net::HTTP`, Not `class Net::HTTPSession`.

```ruby
# Source
module Net
  class HTTP
    ...
  end
  HTTPSession = HTTP
  deprecate_constant :HTTPSession
end
```

RDoc first generates `Net/HTTP.html` with the correct content, and then overwrite the same file with aliased-name content.

Fixes:
- Stop aliased class/module document file overwrite original one.
- Fix aliasing-to-toplevel check (because test failed with the above change).

Documents of `ruby/ruby` that will change:
```
DidYouMean/Formatter.html
  Title change from DidYouMean::VerboseFormatter
Net/HTTP.html
  Title change from Net::HTTPSession
Net/HTTPFound.html
  Title change from Net::HTTPMovedTemporarily
Net/HTTPPayloadTooLarge.html
  Title change from Net::HTTPRequestEntityTooLarge
Net/HTTPRangeNotSatisfiable.html
  Title change from Net::HTTPRequestedRangeNotSatisfiable
Net/HTTPRequestURITooLong.html
  Title was Net::HTTPRequestURITooLarge, file deleted
OpenSSL/PKey/PKeyError.html
  Title change from OpenSSL::PKey::RSAError
Prism/Translation/Parser33.html
  Title change from Prism::Translation::ParserCurrent
```

### DIFF
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -807,11 +807,13 @@ class RDoc::ClassModule < RDoc::Context
       cm_alias = cm.dup
       cm_alias.name = const.name
 
-      # Don't move top-level aliases under Object, they look ugly there
-      unless RDoc::TopLevel === cm_alias.parent then
+      if full_name == 'Object'
+        # Don't move top-level aliases under Object, they look ugly there
+        cm_alias.parent = top_level
+      else
         cm_alias.parent = self
-        cm_alias.full_name = nil # force update for new parent
       end
+      cm_alias.full_name = nil # force update for new parent
 
       cm_alias.aliases.clear
       cm_alias.is_alias_for = cm
--- a/lib/rdoc/generator/darkfish.rb
+++ b/lib/rdoc/generator/darkfish.rb
@@ -356,7 +356,9 @@ class RDoc::Generator::Darkfish
 
     current = nil
 
-    @classes.each do |klass|
+    # Document files are generated only for non-alias classes/modules
+    @classes.reject(&:is_alias_for).each do |klass|
+
       current = klass
 
       generate_class klass, template_file
--- a/test/rdoc/generator/darkfish_test.rb
+++ b/test/rdoc/generator/darkfish_test.rb
@@ -297,11 +297,13 @@ class RDocGeneratorDarkfishTest < RDoc::TestCase
   def test_setup
     @g.setup
 
-    assert_equal [@klass_alias, @ignored, @klass, @object],
+    assert_equal %w[Ignored Klass Klass::A Object],
+                 [@ignored, @klass, @klass_alias, @object].map(&:full_name)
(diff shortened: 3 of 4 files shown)


test/rdoc/generator/darkfish_test.rb
```diff
@@ -297,11 +297,13 @@ def test_install_rdoc_static_file_missing
   def test_setup
     @g.setup
 
-    assert_equal [@klass_alias, @ignored, @klass, @object],
+    assert_equal %w[Ignored Klass Klass::A Object],
+                 [@ignored, @klass, @klass_alias, @object].map(&:full_name)
+    assert_equal [@ignored, @klass, @klass_alias, @object],
```
st0012, line 302: Can you add one of these tests to Aliki’s generator test as well?

lib/rdoc/code_object/class_module.rb
```diff
@@ -807,11 +807,13 @@ def update_aliases
...
-      # Don't move top-level aliases under Object, they look ugly there
-      unless RDoc::TopLevel === cm_alias.parent then
+      if full_name == 'Object'
+        # Don't move top-level aliases under Object, they look ugly there
+        cm_alias.parent = top_level
+      else
         cm_alias.parent = self
-        cm_alias.full_name = nil # force update for new parent
       end
+      cm_alias.full_name = nil # force update for new parent
```
st0012, line 816: Can we have a `ClassModule`/`CodeObject` level test to prevent regression too?

lib/rdoc/generator/darkfish.rb
```diff
@@ -322,6 +322,10 @@ def generate_class(klass, template_file = nil)
 
     debug_msg "  working on %s (%s)" % [klass.full_name, klass.path]
     out_file   = @outputdir + klass.path
+
+    # If klass is an alias, out_file is already generated by the original class/module
+    return if klass.is_alias_for
```
st0012: Can we move the guard to `generate_class_files`'s `@classes.each` block? I feel the intention would be clearer that way.

PR #1519, opened 2025-12-24
Standardize HardBreak markup

vinistock:
Continuing the work from #1389. This PR standardizes the `HardBreak` element to a consistent style and starts inheriting from `Element`.

### DIFF
--- a/lib/rdoc/markup/hard_break.rb
+++ b/lib/rdoc/markup/hard_break.rb
@@ -1,31 +1,34 @@
 # frozen_string_literal: true
-##
-# A hard-break in the middle of a paragraph.
 
-class RDoc::Markup::HardBreak
+module RDoc
+  class Markup
+    # A hard-break in the middle of a paragraph.
+    class HardBreak < Element
+      @instance = new
 
-  @instance = new
+      # RDoc::Markup::HardBreak is a singleton
+      #: () -> HardBreak
+      def self.new
+        @instance
+      end
 
-  ##
-  # RDoc::Markup::HardBreak is a singleton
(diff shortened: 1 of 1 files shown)


lib/rdoc/markup/hard_break.rb
```diff
@@ -1,31 +1,34 @@
...
+      def accept(visitor)
+        visitor.accept_hard_break(self)
+      end
+
+      #: (top) -> bool
+      def ==(other) # :nodoc:
+        self.class === other
+      end
+
+      # @override
```
st0012, line 27: I think this is [Sorbet-specific RBS annotation](https://sorbet.org/docs/rbs-support#method-annotations). Do you know if RBS will officially adopt it or a similar syntax?

st0012 (approved):

lib/rdoc/markup/hard_break.rb
vinistock: Yes, indeed. The RBS ecosystem doesn't have support for abstract/override yet. I assume support will be added at some point, but I'm not sure about the syntax.

For now, since we're not running a type checker on the codebase, these are more documentation than actual type annotations.

st0012: I see. I'm a bit worried that this could confuse other maintainers/contributors that are now familiar with Ruby's typing development. Do you mind mentioning we're using Sorbet flavored RBS purely for documentation in `CONTRIBUTING.md` and `AGENTS.md`?

vinistock: Yeah, I'll open a separate PR to update it.

PR #1521, opened 2025-12-24
Standardize Raw markup

vinistock:
Continuing the work from #1389. This PR standardizes the `Raw` element to a consistent style and starts inheriting from `Element`.

### DIFF
--- a/lib/rdoc/markup/raw.rb
+++ b/lib/rdoc/markup/raw.rb
@@ -1,69 +1,66 @@
 # frozen_string_literal: true
-##
-# A section of text that is added to the output document as-is
 
-class RDoc::Markup::Raw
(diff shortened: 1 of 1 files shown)


lib/rdoc/markup/raw.rb
```diff
@@ -1,69 +1,66 @@
...
+    # A section of text that is added to the output document as-is
+    class Raw
+      # The component parts of the list
+      #: Array[String]
+      attr_reader :parts
+
+      # Creates a new Raw containing +parts+
+      #: (*String) -> void
+      def initialize(*parts)
+        @parts = [].concat(parts)
```
st0012:
```suggestion
        @parts = parts
```

PR #1524, opened 2025-12-24
Standardize BlockQuote markup

vinistock:
Continuing the work from #1389. This PR standardizes the `BlockQuote` element to a consistent style and starts inheriting from `Element`.

I also added the missing parent in `Raw`, which I missed on the other PR.

### DIFF
--- a/lib/rdoc/markup/block_quote.rb
+++ b/lib/rdoc/markup/block_quote.rb
@@ -1,14 +1,14 @@
 # frozen_string_literal: true
-##
-# A quoted section which contains markup items.
 
-class RDoc::Markup::BlockQuote < RDoc::Markup::Raw
-
-  ##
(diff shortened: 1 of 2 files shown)


lib/rdoc/markup/block_quote.rb
```diff
@@ -1,14 +1,14 @@
...
-class RDoc::Markup::BlockQuote < RDoc::Markup::Raw
-
-  ##
-  # Calls #accept_block_quote on +visitor+
-
-  def accept(visitor)
-    visitor.accept_block_quote self
+module RDoc
+  class Markup
+    class BlockQuote < Raw
```
tompng, line 5: I think we need class comment before this line: `# A quoted section which contains markup items.`

PR #1526, opened 2025-12-24
Standardize Rule markup

vinistock:
Continuing the work from #1389. This PR standardizes the `Rule` element to a consistent style and starts inheriting from `Element`.


tompng:
Closing this because changing superclass of RDoc::Markup::Rule from Struct to another one might cause the same problem as #1535 (fixed by partially reverting superclass change in #1549)
```ruby
class RDoc::Markup::Rule < Struct.new :weight
```

I think other standardize pull requests are safe

PR #1532, opened 2025-12-26
Simplify newline handling of comment token in TokenStream#to_html

tompng:
The logic of newline handling to make html `<span>` tag pretty does not need to be restricted to comment token.
Remove comment-specific condition and make it simple.

### What this code is doing
Moves trailing `"\n"` outside of closing `</span>` tag to make generated HTML pretty.
```html
<pre>
<span># comment1
</span><span># comment2
<span><span>42</span>
</pre>
```
↓
```html
<pre>
<span># comment1</span>
<span># comment2</span>
<span>42</span>
</pre>
```

### Background
In most case, only `:on_comment` token (and weird percent literal string token) might end with `"\n"`  because `RDoc::Parser::Ruby` has token-squashing.
I'm writing a new syntax highlighter for RDoc and want to make this pretty-html generation applied to other tokens too.
Removing token-squashing will generate more tokens that ends with `"\n"`.

### DIFF
--- a/.github/workflows/cloudflare-preview.yml
+++ b/.github/workflows/cloudflare-preview.yml
@@ -19,7 +19,7 @@ jobs:
           ref: ${{ github.event.client_payload.pr_head_sha }}
 
       - name: Setup Ruby
-        uses: ruby/setup-ruby@d354de180d0c9e813cfddfcbdc079945d4be589b # v1.275.0
+        uses: ruby/setup-ruby@ae195bbe749a7cef685ac729197124a48305c1cb # v1.276.0
         with:
           ruby-version: '3.4'
           bundler-cache: true
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - name: Setup Ruby
-        uses: ruby/setup-ruby@d354de180d0c9e813cfddfcbdc079945d4be589b # v1.275.0
+        uses: ruby/setup-ruby@ae195bbe749a7cef685ac729197124a48305c1cb # v1.276.0
         with:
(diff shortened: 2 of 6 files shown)


lib/rdoc/token_stream.rb
```diff
@@ -65,7 +59,11 @@ def self.to_html(token_stream)
       text = CGI.escapeHTML text
 
       if style then
-        "<span class=\"#{style}\">#{text}</span>#{"\n" if comment_with_nl}"
+        if text.end_with?("\n")
+          newline = "\n"
+          text = text.byteslice(0...-1)
```
kou: Can we use `chomp` or `delete_suffix` here?

```suggestion
          text = text.delete_suffix("\n")
```

lib/rdoc/token_stream.rb
```diff
@@ -65,7 +59,11 @@ def self.to_html(token_stream)
...
+        end
+        "<span class=\"#{style}\">#{text}</span>#{newline}"
```
st0012: It took me a while to realise that we're utilizing `newline` being `nil` for not generating newlines, which only exists because we declare `newline` conditionally.
Can we add a `newline = nil` in the beginning of the method?

lib/rdoc/token_stream.rb
```diff
@@ -65,7 +59,9 @@ def self.to_html(token_stream)
       text = CGI.escapeHTML text
 
       if style then
-        "<span class=\"#{style}\">#{text}</span>#{"\n" if comment_with_nl}"
+        end_with_newline = text.end_with?("\n")
+        text = text.chomp if end_with_newline
+        "<span class=\"#{style}\">#{text}</span>#{"\n" if end_with_newline}"
       else
         text
       end
```
st0012, line 67: Nit: I think this is more expressive:

```rb
      if style
        if text.end_with?("\n")
          "<span class=\"#{style}\">#{text.chomp}</span>\n"
        else
          "<span class=\"#{style}\">#{text}</span>"
        end
      else
        text
      end
```

PR #1536, opened 2025-12-31
Prevent style rendering in code blocks

st0012:
### Before

Screenshot 2025-12-31 at 23 44 19

### After

Screenshot 2025-12-31 at 23 43 58


lib/rdoc/markup/attribute_manager.rb
```diff
@@ -245,6 +251,20 @@ def mask_protected_sequences
     @str.gsub!(/\\(\\[#{Regexp.escape @protectable.join}])/m, "\\1")
```
st0012, line 251: There's a lot of repeated regexp creation like this in this file. I think most of them can be extracted into ivars.

tompng (approved):

PR #1538, opened 2026-01-01
Improve code block language detection

st0012:
### Fix verbatim blocks with non-Ruby formats getting Ruby highlighting

Code blocks with explicit language identifiers (e.g., ```markdown) were sometimes incorrectly receiving Ruby syntax highlighting when their content happened to be parseable as Ruby.

Now Ruby highlighting is only applied when:
- Format is explicitly `ruby`, or
- No format specified and content is parseable as Ruby

Other formats now get their language class (e.g., `class="markdown"`) for potential frontend-based highlighting support.

### Support `rb` as an alternative markdown code block language


kou (approved):
+1

PR #1539, opened 2026-01-02
Fix cross-reference resolution for constants

### LINKED ISSUE #1011: [Bug] Cannot link to constants in Markdown
How to reproduce:

Create the following files:
```ruby
# test.rb
module TestModule
  TEST_CONSTANT = 1108
  def test_instance_method
  end
end
```
```markdown
<!-- test.md -->
TestModule
TestModule::TEST_CONSTANT
TestModule#test_instance_method
```
Run the command:
```shell
rdoc test.rb test.md
```
Then, use a browser to open `doc/test_md.html`.

Result:
image

Expected:
image

extern-c:
This PR resolves #1011 by ensuring that constant names ~~with `::` (e.g. `::CONSTANT_NAME`) can be used to find constants in contexts~~ can be resolved by `RDoc::CrossReference#resolve_local_symbol` (originally named `resolve_method`). Please note that the darkfish and aliki themes do not display links for constants in the constants section (just the names of the constants in a description list), so the changes in this PR just make it possible to cross reference constants. I will create a separate PR to propose changes to the themes.

### DIFF
--- a/lib/rdoc/code_object/context.rb
+++ b/lib/rdoc/code_object/context.rb
@@ -779,7 +779,9 @@ class RDoc::Context < RDoc::CodeObject
   # Finds a constant with +name+ in this context
 
   def find_constant_named(name)
-    @constants.find { |m| m.name == name || m.full_name == name }
+    @constants.find do |m|
+      m.name == name || m.full_name == name
+    end
   end
 
   ##


tompng:
What happens in the reproduction code in #1011 is:
1. `find_symbol('TestModule::TEST_CONSTANT')` is called.
2. `find_local_symbol('TestModule::TEST_CONSTANT')` is called because constant can't be found globally. (I think this is the bug)
3. Module and constant not found, fallback to method search
4. Name is treated as method name, decomposed into container part(`TestModule`) and method name part (`::TEST_CONSTANT`) (`::` means `.` in here. `Math::sin(1) == Math.sin(1)`)
5. Method search mechanism somehow calls `find_constant_named('::TEST_CONSTANT')`

So I think modifying `find_constant_named` to handle `::TEST_CONSTANT` is not the correct fix.
I think we should handle this case in step `1.`: `find_symbol('TestModule::TEST_CONSTANT')` should find constant correctly.

extern-c:
> What happens in the reproduction code in #1011 is:
> 
> 1. `find_symbol('TestModule::TEST_CONSTANT')` is called.
> 2. `find_local_symbol('TestModule::TEST_CONSTANT')` is called because constant can't be found globally. (I think this is the bug)
> 3. Module and constant not found, fallback to method search
> 4. Name is treated as method name, decomposed into container part(`TestModule`) and method name part (`::TEST_CONSTANT`) (`::` means `.` in here. `Math::sin(1) == Math.sin(1)`)
> 5. Method search mechanism somehow calls `find_constant_named('::TEST_CONSTANT')`
> 
> So I think modifying `find_constant_named` to handle `::TEST_CONSTANT` is not the correct fix. I think we should handle this case in step `1.`: `find_symbol('TestModule::TEST_CONSTANT')` should find constant correctly.

Hi @tompng,

Thank you for your feedback. To expand on what you said, `RDoc::CrossReference#resolve_method`  decomposes `TestModule::TEST_CONSTANT` into `container` + `symbol` and then tries to find `symbol` via `Context#find_local_symbol`:
```ruby
  def find_local_symbol(symbol)
    find_method_named(symbol) or
    find_constant_named(symbol) or
    find_attribute_named(symbol) or
    find_external_alias_named(symbol) or
    find_module_named(symbol) or
    @store.find_file_named(symbol)
  end
``` 
https://github.com/ruby/rdoc/blob/master/lib/rdoc/code_object/context.rb#L828.

I thought it would be okay to modify `find_constant_named` to handle `::` because `find_local_symbol` may call `find_constant_named` and  the decomposition logic in `resolve_method` _could work_ for constants as well. However, trying to find a constant through a method called `resolve_method` might sound confusing.

tompng:
I see. Changing `resolve_method` to handle constant seems OK.
It needs both of these:
- Rename to `resolve_method_or_constant`, `resolve_local_symbol` or something that express it's not only for methods
- Explicitly call `container.find_constant_named('TEST_CONSTANT')` in the renamed method. (Remove `::`)

We can clear the name confusion, don't need to make `find_constant` accept preceding `::` (because it was a side effect of method search).
Ideally, I think we need `Context#find_local_method_symbol`, similar to find_local_symbol but only search method including attribute, and call it from `resolve_method`.

extern-c:
> I see. Changing `resolve_method` to handle constant seems OK. It needs both of these:
> 
> * Rename to `resolve_method_or_constant`, `resolve_local_symbol` or something that express it's not only for methods

:+1:  Please see 2e0782112723607096636965a19b3e63bb41c317.

> * Explicitly call `container.find_constant_named('TEST_CONSTANT')` in the renamed method. (Remove `::`)
> 
> We can clear the name confusion, don't need to make `find_constant` accept preceding `::` (because it was a side effect of method search). Ideally, I think we need `Context#find_local_method_symbol`, similar to find_local_symbol but only search method including attribute, and call it from `resolve_method`.

I will look into the remaining items later this week. Thank you.

lib/rdoc/cross_reference.rb
```diff
@@ -161,18 +167,18 @@ def resolve_method(name)
...
+            ref = container.find_local_symbol "::#{symbol}"
+            ref = container.find_ancestor_local_symbol "::#{symbol}" unless ref
+            ref = container.find_local_symbol "##{symbol}" unless ref
+            ref = container.find_ancestor_local_symbol "##{symbol}" unless ref
           end
         else
-          ref = container.find_local_symbol method
-          ref = container.find_ancestor_local_symbol method unless ref
+          ref = container.find_local_symbol symbol
+          ref = container.find_ancestor_local_symbol symbol unless ref
```
tompng, line 184: How about checking for constant after this line?
```ruby
ref = container.find_constant_named(symbol[2..]) if !ref && type == '::'
```

Treating `::NAME` as local constant is really misleading because normally `::NAME` means top level constant in Ruby, and `::` is just a separator to be removed.

For background, current codebase of RDoc is hard to maintain. Removing complexity, fixing misleading code and refactoring is in a highest priority for me as a maintainer of RDoc.
So changing `find_constant_named` to ignore preceding `::` is not acceptable.

Adding an explicit constant-name regexp extraction is also a good idea.
```ruby
if /#{CLASS_REGEXP_STR}::([A-Z]\w*)\z/o =~ name then
  type = 'const'
  container = @context.find_symbol_module($1)
  symbol = $2
elsif ...
```

extern-c:
>We can clear the name confusion, don't need to make find_constant accept preceding :: (because it was a side effect of method search). 

The changes in 20b3cd18247333648d4cd2180fc5e368285bd496 address this. Thanks.

> - Explicitly call container.find_constant_named('TEST_CONSTANT') in the renamed method. (Remove ::)

`resolve_local_symbol` is still relying on `Context#find_local_symbol` (instead of calling `Context#find_method_named` and `Context#find_attribute_named` directly, or `Context#find_constant_named` directly) and `RDoc::ClassModule#find_ancestor_local_symbol` to find methods, attributes or constants. Do you still think this should be improved now as well?

>For background, current codebase of RDoc is hard to maintain. Removing complexity, fixing misleading code and refactoring is in a highest priority for me as a maintainer of RDoc.
So changing find_constant_named to ignore preceding :: is not acceptable.

Thanks for the background. That's understandable and I agree.

lib/rdoc/code_object/context.rb
```diff
@@ -779,9 +779,7 @@ def find_class_method_named(name)
   # Finds a constant with +name+ in this context
 
   def find_constant_named(name)
-    @constants.find do |m|
-      m.name == name || m.full_name == name
-    end
+    @constants.find { |m| m.name == name || m.full_name == name }
```
tompng: Just one more thing. Can you revert this change? because this is only a coding style change.

I think other part of this pull request looks good :+1:

extern-c: >Just one more thing. Can you revert this change? because this is only a coding style change.

Sure, no problem. Addressed in f2bfdddaca444bfddb6718d5cf7d798562865182. Thanks!

PR #1540, opened 2026-01-03
Support GitHub style markdown heading anchor and link reference

### LINKED ISSUE #1537: In `.md` files, `[link](#header)` should be supported
RDoc should be able to convert `#header` into a proper `rdoc-ref` style value so the generated html file has the correct link.

st0012:
I started the work to address #1537, but I think we should just go with the GH markdown style anchor/link generation in the future.

Anchors Generated:

| Context | Before | After |
|---------|--------|-------|
| Heading `== Hello` (standalone) | `id="label-Hello"` | `id="hello"` |
| Heading `== Hello` (in class Foo) | `id="class-Foo-label-Hello"` | `id="class-foo-hello"` |
| Heading `== Hello` (in method #bar) | `id="method-i-bar-label-Hello"` | `id="method-i-bar-hello"` |
| Class/Module anchor | `id="class-Foo::Bar"` | `id="class-foo-bar"` |

**Legacy anchors preserved as hidden `<span class="legacy-anchor">` elements. So links targeting old anchors should still work.**

Links Generated:

| Syntax | Before | After |
|--------|--------|-------|
| `rdoc-ref:@foo` | `href="#label-foo"` | `href="#foo"` |
| `rdoc-ref:Class@foo` | `href="#class-Class-label-foo"` | `href="#class-class-foo"` |
| `rdoc-ref:Class#method@foo` | `href="#method-i-method-label-foo"` | `href="#method-i-method-foo"` |
| `rdoc-ref:Class@Section Title` | `href="#Section Title"` | `href="#section-title"` |

These changes now enable Markdown links to work with GitHub-style anchors.

Markdown `[link](#foo)` passes through unchanged as `href="#foo"`.

- Before: Anchor was `id="label-Foo"` -> link broken
- After: Anchor is `id="foo"` -> link works

Supportive Changes:

- ToC sidebar links updated to use new anchor format

Closes #1537

### DIFF
--- a/ExampleMarkdown.md
+++ b/ExampleMarkdown.md
@@ -14,6 +14,18 @@ For the following styles, see ExampleRDoc.rdoc for style examples:
 
 These items all use the same styles as RDoc format files.
 
+## Anchor Links
+
+RDoc supports GitHub-style anchor links. You can link to any heading using its
+anchor, which is the heading text converted to lowercase with spaces replaced
+by hyphens and special characters removed.
+
+For example:
+
+* [Link to Footnotes](#footnotes)
+* [Link to Blockquotes](#blockquotes)
+* [Link to Anchor Links](#anchor-links)
+
 ## Footnotes
 
 Footnotes are rendered at the bottom of the documentation section[^1].  For
@@ -36,4 +48,3 @@ Here is how a blockquote looks.
 > > 75 years?
 
 This text is from [Riker Ipsum](http://rikeripsum.com)
-
--- a/ExampleRDoc.rdoc
+++ b/ExampleRDoc.rdoc
@@ -3,6 +3,18 @@
 This document contains example output to show RDoc styling.  This file was
 created from a RDoc Markup file.
 
+== Anchor Links
+
+RDoc generates GitHub-style anchors for headings.  The anchor is the heading
+text converted to lowercase with spaces replaced by hyphens and special
+characters removed.
+
+You can link to headings using Markdown-style syntax:
+
+- {Link to Headings}[#headings]
+- {Link to Paragraphs}[#paragraphs]
+- {Link to Verbatim sections}[#verbatim-sections]
+
 == Headings
 
 You should not use headings beyond level 3, it is a sign of poor organization
--- a/doc/rdoc/markup_reference.rb
+++ b/doc/rdoc/markup_reference.rb
@@ -957,22 +957,30 @@ require 'rdoc'
 #
 # [Heading]
 #
-#   - Link: <tt>RDoc::RD@LICENSE</tt> links to RDoc::RDoc::RD@LICENSE.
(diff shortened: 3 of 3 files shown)


lib/rdoc/markup/to_html.rb
```diff
@@ -312,6 +312,13 @@ def accept_heading(heading)
     level = [6, heading.level].min
 
     label = heading.label @code_object
+    legacy_label = heading.legacy_label @code_object
+
+    # Add legacy anchor before the heading for backward compatibility.
+    # This allows old links with label- prefix to still work.
+    if legacy_label && @options.output_decoration && !@options.pipe
```
kou: Do we need `legacy_label` check here? It seems that `heading.legacy_label` always returns a `String`.


```suggestion
    if @options.output_decoration && !@options.pipe
```

lib/rdoc/markup/to_html.rb
```diff
@@ -368,14 +375,18 @@ def convert_string(text)
...
+  # - File references (.rb, .rdoc, .md) are converted to .html paths
+  # - Anchor URLs (#foo) pass through unchanged for GitHub-style header linking
+  # - Footnote links get wrapped in <sup> tags
 
   def gen_url(url, text)
     scheme, url, id = parse_url url
 
-    if %w[http https link].include?(scheme) and
-       url =~ /\.(gif|png|jpg|jpeg|bmp)$/ then
+    if %w[http https link].include?(scheme) && url =~ /\.(gif|png|jpg|jpeg|bmp)$/
```
kou: This is not related to this PR change but `\z` is better than `$`:


```suggestion
    if %w[http https link].include?(scheme) && url =~ /\.(gif|png|jpg|jpeg|bmp)\z/
```

lib/rdoc/markup/to_html.rb
```diff
@@ -387,9 +398,11 @@ def gen_url(url, text)
 
       link = "<a#{id} href=\"#{url}\">#{text}</a>"
 
-      link = "<sup>#{link}</sup>" if /"foot/ =~ id
-
-      link
+      if /"foot/ =~ id
```
kou:
```suggestion
      if /"foot/.match?(id)
```

lib/rdoc/markup/to_html.rb
st0012: Yeah good catch 👍

Earlopain:
> Legacy anchors preserved as hidden <span class="legacy-anchor"> elements. So links targeting old anchors should still work.

I don't believe this is working as intended. I've seen PRs like https://github.com/ruby/ruby/pull/15963 (among others) so I guess something unexpected is happening?

st0012:
@Earlopain Thanks for reporting this. I'll take a look. It may be related to #1590 as well

PR #1541, opened 2026-01-04
Fix Markdown strikethrough (`~~text~~`) not rendering in HTML output

st0012:
The Markdown parser correctly parsed `~~text~~` syntax but the `AttributeManager` did not recognize the resulting formats for HTML conversion:

- Single words: `~~word~~` -> `~word~` (not registered as word pair)
- Multi-word: `~~words here~~` -> `<s>words here</s>` (tag not registered)

Add support for strikethrough formatting:

- Register `~` as word pair for `:STRIKE` attribute
- Register `<s>`, `<del>`, and `<strike>` HTML tags for `:STRIKE`
- Add `:STRIKE` -> `<del>` tag mapping in `ToHtml`

All strikethrough input formats now render as `<del>text</del>` in HTML.


kou (approved):
+1

PR #1544, opened 2026-01-04
Highlight bash commands

st0012:
The current bash command code blocks are highlighted inconsistently because some code blocks got detected as Ruby, so Ruby highlighting is applied. This has been addressed in #1538.

Screenshot 2026-01-04 at 18 31 29

But after the PR, we don't have any highlighting for bash, which I think we can provide rather simply.

With this PR, we can highlight code blocks tagged as `sh`, `shell`, `bash`, and `console` with a simple JS highlighter.

Screenshot 2026-01-04 at 18 30 02

### DIFF
--- a/lib/rdoc/generator/template/aliki/js/bash_highlighter.js
+++ b/lib/rdoc/generator/template/aliki/js/bash_highlighter.js
@@ -1,0 +1,167 @@
+/**
+ * Client-side shell syntax highlighter for RDoc
+ * Highlights: $ prompts, commands, options, strings, env vars, comments
+ */
+
+(function() {
+  'use strict';
+
+  function escapeHtml(text) {
+    return text
+      .replace(/&/g, '&amp;')
+      .replace(/</g, '&lt;')
+      .replace(/>/g, '&gt;')
+      .replace(/"/g, '&quot;')
+      .replace(/'/g, '&#39;');
+  }
+
+  function wrap(className, text) {
+    return '<span class="' + className + '">' + escapeHtml(text) + '</span>';
+  }
+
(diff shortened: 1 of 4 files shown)


lib/rdoc/generator/template/aliki/js/bash_highlighter.js
```diff
@@ -0,0 +1,167 @@
...
+            i = valEnd;
+          }
+          afterSpace = false;
+          continue;
+        }
+      }
+
+      // Option (must be after whitespace)
+      if (ch === '-' && afterSpace) {
+        var match = line.slice(i).match(/^--?[a-zA-Z0-9_-]+(=[^\s]*)?/);
```
tompng: Assuming this input
```sh
git commit --message="initial commit"
```
How about `(=[^\s]*)?` → `(=[^"\s]*)?`
`""` part will be handled in `// Double-quoted string` logic above.

lib/rdoc/generator/template/aliki/js/bash_highlighter.js
```diff
@@ -0,0 +1,167 @@
...
+          result += wrap('sh-option', match[0]);
+          i += match[0].length;
+          afterSpace = false;
+          continue;
+        }
+      }
+
+      // Command (first word: regular, ./path, ../path, ~/path, @scope/pkg)
+      if (!seenCommand && afterSpace) {
+        var isCmd = /[a-zA-Z0-9@~]/.test(ch) ||
```
tompng: How about adding `/` to accept `/bin/sh` as command?

PR #1546, opened 2026-01-04
Use class for highlight selector

st0012:
After #1538, code blocks with language declaration now have the class as the language name. So we can simply use the class for the highlight selector.


kou (approved):
+1

PR #1547, opened 2026-01-05
Expand GitHub style references in ChangeLog to URL

nobu:
Currently, ruby/ruby formats ChangeLog and expands such references to links.
But such links in ChangeLog that is a plain text does not make sense.
I think it is suitable to be done at conversion from ChangeLog to HTML

### DIFF
--- a/lib/rdoc/parser/changelog.rb
+++ b/lib/rdoc/parser/changelog.rb
@@ -286,7 +286,7 @@ class RDoc::Parser::ChangeLog < RDoc::Parser
             end
             # repo#PR, repo@HASH
             #   ruby/ruby#15791 -> [ruby/ruby#15791](https://github.com/ruby/ruby/pull/15791)
-            #   ruby/ruby@a8a989b6 -> [ruby/ruby@a8a989b6](https://github.com/ruby/net-http/commit/a8a989b6)
+            #   ruby/ruby@a8a989b6 -> [ruby/ruby@a8a989b6](https://github.com/ruby/ruby/commit/a8a989b6)
             # ref in branckets is not extended
             #   [ruby/net-imap#543][ruby/ruby#15791] -> [ruby/net-imap#543][ruby/ruby#15791]
             contents.gsub!(%r[(?<![-\w#/@]|\]\[)([-\w]+/[-\w]+)(?:@(\h{8,40})|\#(\d+))(?![-\w#/@]|\]\[)]) do


lib/rdoc/parser/changelog.rb
```diff
@@ -274,6 +274,18 @@ def create_entries(entries)
       def initialize(base, commit, author, email, date, contents)
         case contents
         when String
+          if base&.match(%r[\A([^:/]+:/+[^/]+/)[^/]+/[^/]+/])
+            repo, host = $&, $1
+            contents = contents.dup
+            contents.gsub!(/\b(?:(?i:fix(?:e[sd])?) +)\K#(\d+\b)|\bGH-(\d+)\b|\(\K\#(\d+)(?=\))/) do
```
kou: Could you add a comment with example substitutions for easy to understand? For example:

```ruby
# base: https://github.com/ruby/ruby/
#   Fix #15791 -> [Fix #15791](https://github.com/ruby/ruby/pull/15791)
#   ...
```

kou: Could you unify escape style to `\K#` or `\K\#`?

```suggestion
            contents.gsub!(/\b(?:(?i:fix(?:e[sd])?) +)\K\#(\d+\b)|\bGH-(\d+)\b|\(\K\#(\d+)(?=\))/) do
```

lib/rdoc/parser/changelog.rb
```diff
@@ -274,6 +274,18 @@ def create_entries(entries)
...
+              "[#$&](#{repo}pull/#{$1 || $2 || $3})"
```
kou: `.../issues/#{N}` is better than `.../pull/#{N}` because `.../issues/#{N}` works for both of issues and pull requests:

```suggestion
              "[#$&](#{repo}issues/#{$1 || $2 || $3})"
```

kou: Could you unify `#{...}` style?

```suggestion
              "[#{$&}](#{repo}pull/#{$1 || $2 || $3})"
```

nobu: I think `/pull/` also works as well as `/issues/`, that is the conversion is bidirectional.

kou: Oh... I didn't know it...

nobu: I chose `pull` just because ruby/ruby is not using GitHub Issues.

lib/rdoc/parser/changelog.rb
```diff
@@ -274,6 +274,22 @@ def create_entries(entries)
...
+            repo, host = $&, $1
+            contents = contents.dup
+            # base: https://github.com/ruby/ruby/
+            #   Fix #15791 -> Fix [#15791](https://github.com/ruby/ruby/pull/15791)
+            #   GH-15791 -> [GH-15791](https://github.com/ruby/ruby/pull/15791)
+            #   (#15791) -> ([#15791](https://github.com/ruby/ruby/pull/15791))
+            contents.gsub!(/\b(?:(?i:fix(?:e[sd])?) +)\K\#(\d+\b)|\bGH-(\d+)\b|\(\K\#(\d+)(?=\))/) do
+              "[#{$&}](#{repo}pull/#{$1 || $2 || $3})"
+            end
+            contents.gsub!(%r[(?<![-\w#/@]|\]\[)([-\w]+/[-\w]+)(?:@(\h{8,40})|\#(\d+))(?![-\w#/@]|\]\[)]) do
```
kou, line 292: Could you add an example comment here too?

lib/rdoc/parser/changelog.rb
```diff
@@ -274,6 +274,27 @@ def create_entries(entries)
...
+            # base: https://github.com/ruby/ruby/
+            #   Fix #15791 -> Fix [#15791](https://github.com/ruby/ruby/pull/15791)
+            #   GH-15791 -> [GH-15791](https://github.com/ruby/ruby/pull/15791)
+            #   (#15791) -> ([#15791](https://github.com/ruby/ruby/pull/15791))
+            contents.gsub!(/\b(?:(?i:fix(?:e[sd])?) +)\K\#(\d+\b)|\bGH-(\d+)\b|\(\K\#(\d+)(?=\))/) do
+              "[#{$&}](#{repo}pull/#{$1 || $2 || $3})"
+            end
+            # repo#PR, repo@HASH
+            #   ruby/ruby#15791 -> [ruby/ruby#15791](https://github.com/ruby/ruby/pull/15791)
+            #   ruby/ruby@a8a989b6 -> [ruby/ruby@a8a989b6](https://github.com/ruby/net-http/commit/a8a989b6)
```
kou:
```suggestion
            #   ruby/ruby@a8a989b6 -> [ruby/ruby@a8a989b6](https://github.com/ruby/ruby/commit/a8a989b6)
```

PR #1548, opened 2026-01-05
Align strikethrough support with GitHub Markdown

st0012:
1. GitHub doesn't support `~word~` syntax for strikethrough, so we shouldn't either.
2. It's hard to make RDoc markup support `~~word~~` without supporting `~word~` as well due to the way attribute manager works. So instead we rollback the support added in #1541. Users can still use `<del>text</del>` syntax for strikethrough.
3. Fix `<code><del>text</del></code>` not rendering as `<del>text</del>` in HTML output.
4. Update documentation to reflect the new support.

### DIFF
--- a/doc/markup_reference/rdoc.rdoc
+++ b/doc/markup_reference/rdoc.rdoc
@@ -664,7 +664,7 @@ For C code, the directive may appear in a stand-alone comment.
 
 Text markup is metatext that affects HTML rendering:
 
-- {Typeface}[rdoc-ref:rdoc@Typeface+Markup]: italic, bold, monofont.
+- {Typeface}[rdoc-ref:rdoc@Typeface+Markup]: italic, bold, monofont, strikethrough.
 - {Character conversions}[rdoc-ref:rdoc@Character+Conversions]: copyright, trademark, certain punctuation.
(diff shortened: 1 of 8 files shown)


tompng (approved):

doc/markup_reference/rdoc.rdoc
```diff
@@ -766,6 +767,23 @@ Rendered HTML:
...
+
+Example input:
+
+  <del>Strikethrough words</del> in a paragraph.
+
+  <del>Deleted passage containing _italics_ and *bold*.</del>
+
+Rendered HTML:
+
+>>>
```
skatkov, line 782: **question:** Is this is a leftover from merge conflict?

st0012: It's RDoc's way to declare [quote blocks](https://ruby.github.io/rdoc/doc/markup_reference/rdoc_rdoc.html#block-quotes).

nevans:
> 1. GitHub doesn't support `~word~` syntax for strikethrough, so we shouldn't either.

What?  I very rarely use ~word~ in rdoc, but I use it on Github all the time... I used it several times just this morning.  I've used it in this comment.

PR #1551, opened 2026-01-06
Allow using backticks to quote text in RDoc markup too

st0012:
Users can now use both ``` ` ``` and `+` to make simple code blocks in RDoc markup.

Using backticks to quote code has become a natural habit for many users, so it's common to see it being incorrectly used to contribute to Ruby documentation.

But instead of "correcting" users' behaviour, we should just support it. It'll also allow us to partially migrate some RDoc markup to Markdown.


st0012:
cc @nobu

PR #1552, opened 2026-01-07
[DOC] Remove _italic_ *bold* styling inside codeblock

tompng:
Styling in codeblock is already dropped as a bugfix (in https://github.com/ruby/rdoc/pull/1536) so that codeblock does not need to escape frequently used operators like `+`, `*`.

Some document in ruby/ruby also uses `_italic_` inside `<tt>` tags and we should also fix them.
For example, [ruby/ruby/doc/_regexp.rdoc](https://github.com/ruby/ruby/blob/25c72b0e8e206e5baec71d4ece7551b7da7da445/doc/_regexp.rdoc?plain=1#L417C1-L417C52) has `<tt>(?=_pat_)</tt>`

### DIFF
--- a/doc/markup_reference/rdoc.rdoc
+++ b/doc/markup_reference/rdoc.rdoc
@@ -549,7 +549,7 @@ and <tt>:enddoc:</tt> may appear in a stand-alone comment:
 
 ==== <tt>:markup:</tt>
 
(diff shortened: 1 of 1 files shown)


st0012 (approved):
Good catch 👍

Issue #1475, opened 2025-11-26
Display "generated with <commit sha>" in the footer


st0012:
I think it'd be beneficial for both the doc writer and consumer to be able to tell the version of the doc directly from the web page.

t-mangoe:
Hi! I'm interested in this issue (as my first contribution to rdoc). If possible, I am considering making a contribution.
Is it the Git hash that should be displayed in the footer?

st0012:
Yes. So to properly support this feature, RDoc needs to:
- Check if the project uses git
    - If not, don't display that section at all
- Use a safe way to fetch the current commit sha
    - We may also want to consider other info, like branch/tag name when applicable, and perhaps date too?

t-mangoe:
@st0012 
Thank you for your comment!
I'm considering using the [“git” gem](https://github.com/ruby-git/ruby-git) to fetch the current commit sha. In this way, we can check if the project uses git, and we may get other git information.
Adding a gem might be a concern...  What do you think?

t-mangoe:
I've implemented this to a certain extent, so would it be okay if I try submitting a PR?
Please review the code, and if there are any improvements to be made, I will make the corrections.

PR #1558, opened 2026-01-10
Display commit sha value in the footer.

t-mangoe:
The commit SHA value is displayed in the footer. The value is retrieved using the Git gem.
This pull request relates to issue #1475.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,7 +11,6 @@ gem 'rubocop', '>= 1.31.0'
 gem 'gettext'
 gem 'prism', '>= 0.30.0'
 gem 'webrick'
-gem 'git'
 
 platforms :ruby do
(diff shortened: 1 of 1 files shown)


lib/rdoc/rdoc.rb
```diff
@@ -494,6 +496,13 @@ def document(options)
     exit @stats.fully_documented? if @options.coverage_report
   end
 
+  def detect_git_commit
+    require "git"
+    Git.open(Dir.pwd).object("HEAD").sha
```
extern-c: You could consider using Ruby's  `Open3.capture3` to run Git commands directly and capture the output you need, instead of relying on the git gem. This can simplify things if you only need to run a single command.

t-mangoe: Thanks!! @extern-c 
I changed the code to use Open3.capture3.

PR #1559, opened 2026-01-10
Replace attribute_manager with a new rdoc-inline-format parser

tompng:
## RDoc::Markup::AttributeManager
RDoc parses block-level structure in `RDoc::Markup::Parser` and parses inline styling in `RDoc::Markup::AttributeManager`.
`RDoc::Markup::AttributeManager` is a string-replacing/macro based parser.
It's very complicated, and its mechanism are the root cause of many bugs. We need to remove and replace it.

It converts to a flow/stream of string and styling changing operation.
Unfortunately, inline styles are structured data, output format (HTML) is also a structured data. There's an architecture mismatch.

Tidylink `{label}[url]` should be a syntax rule, but it is currently handled in a
regexp-based macro called `regexp_handling`.
Of course parsing nested structure such as styled tidylink label frequently fails. It's uncontrollable.

## Solution

- Eliminate `RDoc::Markup::AttributeManager`
- Create a parser that generates structured data
- Traverse structured data to generate output instead of string replacing
- Use controllable regexp-handling macro: only apply to text nodes

### Tokens

- Word pairs `+word+` `*word*` `_word_` `` `word` `` and so on
- Standalone tags `<br>`, `<tt>code_text</tt>`, `<code>code_text</code>`
- Open and close tags `<i>` `</i>`, `<b>` `</b>`, `<em>` `</em>`, `<s>` `</s>`, `<del>` `</del>`
- Tidy link opening `{` and closing `}[url_part]`
- Simplified tidylink `word[url_part]`
- Text nodes

### Regexp handling macro

Matching with CROSSREF, RDOCREF, HYPERLINK regexp should be applied only to text nodes after parsing phase.
Parsed tree modication instead of text-node gsub is also another option to implement this.

### Error recovery

RDoc format doesn't have syntax error. We need to make the behavior as similar before.

- Closing tags and braces will invalidate unclosed tags, and unclosed tags are treated as plain text
  - `<a><b><c></a>` will be `<a>&lt;b&gt;&lt;c&gt;</a>`
  - `{<a><b>}[url]` will be a tidylink with label `&lt;a&gt;&lt;b&gt;`
- Unmatched closing tags and braces will be treated as plain text
  - `<a></b>}</a>` will be `<a>&lt;/b&gt;}</a>`
- Tidylink inside tidylink will invalidate outside tidylinks
  - `{{inner}[url]}[<b>]</b>` will be `"{" + inner_tidylink + "}[" + bold("]")`

### Simplified tidylink

RDoc was converting ``a*_`+<b>c[foo]`` to ``<a href="foo">a*_`+&lt;b&gt;c</a>``. This is terrible, it can't coexist with other syntaxes like `*word*` `_word_` `+word+`.
We should restrict characters and recommend `{label}[url]`.
In this pull request, only `Alphanumeric[url]` (should start with alphabet) is supported .

## Diff

Running rdoc in `ruby/ruby`, there were 48 HTML files that had changes (with commit hash:763f379).
There are too many diffs, but I think most of them are acceptable.

<details>
<summary>Some diff</summary>

### List of all files with diffs:

```
Array.html	ERB.html	Encoding/Converter.html	Hash.html
IO.html	Kernel.html	Net/HTTP.html	Net/HTTPHeader.html
OpenSSL/Cipher.html	OpenSSL/SSL/SSLSocket.html	OpenSSL/X509/Name.html	OpenSSL.html
PrettyPrint.html	Random/Formatter.html	Regexp.html	Socket.html
String.html	StringScanner.html	Struct.html	SyntaxSuggest/CodeLine.html
Time.html	TracePoint.html	__/ruby/LEGAL.html	__/ruby/doc/NEWS/NEWS-1_9_2.html
__/ruby/doc/NEWS/NEWS-2_2_0.html	__/ruby/doc/NEWS/NEWS-2_4_0.html	__/ruby/doc/NEWS/NEWS-3_0_0_md.html	__/ruby/doc/NEWS/NEWS-3_1_0_md.html
__/ruby/doc/NEWS/NEWS-3_2_0_md.html	__/ruby/doc/NEWS/NEWS-3_3_0_md.html	__/ruby/doc/NEWS/NEWS-3_4_0_md.html	__/ruby/doc/NEWS/NEWS-4_0_0_md.html
__/ruby/doc/contributing/building_ruby_md.html	__/ruby/doc/contributing/concurrency_guide_md.html	__/ruby/doc/contributing/documentation_guide_md.html	__/ruby/doc/contributing/dtrace_probes_rdoc.html
__/ruby/doc/contributing/reporting_issues_md.html	__/ruby/doc/distribution/windows_md.html	__/ruby/doc/extension_ja_rdoc.html	__/ruby/doc/extension_rdoc.html
__/ruby/doc/language/encodings_rdoc.html	__/ruby/doc/language/globals_md.html	__/ruby/doc/language/marshal_rdoc.html	__/ruby/doc/language/strftime_formatting_rdoc.html
__/ruby/doc/security/command_injection_rdoc.html	__/ruby/doc/security/security_rdoc.html	__/ruby/doc/syntax/keywords_rdoc.html	__/ruby/doc/syntax/pattern_matching_rdoc.html
```

### Array.html

| | expression | memo |
| --- | --- | --- |
| source | `A {%w or %W string-array Literal}[rdoc-ref:...]` | |
| master | `<p>A array literal:</p>` | bug |
| this PR | `<p>A %w or %W string-array Literal:</p>` | fixed |

### ERB.html

| | expression | memo |
| --- | --- | --- |
| source | `**Embedded Ruby**` | |
| master | `<strong>Embedded Ruby</strong>` | crossref inside em is disabled |
| this PR | `<strong>Embedded <a href="Ruby.html"><code>Ruby</code></a></strong>` | crossref enabled |

Same diff in `IO.html`, `Net/HTTPHeader.html`, `OpenSSL/Cipher.html` and `Socket.html`

### Encoding/Converter.html

| | expression | memo |
| --- | --- | --- |
| source | `"\uFFFD"` | |
| master | `“uFFFD”` | `\u` is unescaped |
| this PR | `“\uFFFD”` | backslash remains |

Same diff in `PrettyPrint.html`, `Random/Formatter.html`

### Hash.html

| | expression | memo |
| --- | --- | --- |
| source | `hash same \hash key` | |
| master | `same hash key` | `\h` is unescaped |
| this PR | `same \hash key` | backslash remains |

### Kernel.html

| | expression | memo |
| --- | --- | --- |
| source | ``{\`command`}[rdoc-ref:Kernel#`]`` | backquote escaped |
| master | `<a href="Kernel.html#method-i-60"><code>command</code></a>` | Escape ignored(bug) |
| this PR | ``<a href="Kernel.html#method-i-60">‘command`</a>`` | fixed |

| | expression | memo |
| --- | --- | --- |
| source | `Whether the entity's setgid bit is set.` | |
| master | `<td align="left">Whether the entity's setgid bit is set.</td>` | quote is `'` |
| this PR | `<td align="left">Whether the entity’s setgid bit is set.</td>` | quote is `’` |

### Net/HTTP.html

| | expression | memo |
| --- | --- | --- |
| source | `{Proxy Using ENV['http_proxy']}[rdoc-ref:...]` | Tidylink inside tidylink (invalid) |
| master | `<a href="...">Proxy Using <a href="&#39;http_proxy&#39;">ENV</a></a>` | Bug (`</a></a>`) |
| this PR | `{Proxy Using <a href="&#39;http_proxy&#39;">ENV</a>}[<a href="...">...</a>]` | |

### OpenSSL/SSL/SSLSocket.html

| | expression | memo |
| --- | --- | --- |
| source | `_remote\_host_ and _remote\_port_` | Wrong escaping |
| master | `<em>remote</em>host_ and <em>remote</em>port_` | messed |
| this PR | `_remote_host_ and _remote_port_` | unescaped |

### OpenSSL/X509/Name.html

| | expression | memo |
| --- | --- | --- |
| source | `or ++1+ if` | Should be considered closing `++` missing |
| master | `<code>+1</code>` | bug |
| this PR | `++1+` | fixed |

### OpenSSL.html

| | expression | memo |
| --- | --- | --- |
| source | `v2.0[http://www.rsa.com/rsalabs/node.asp?id=2127]` | |
| master | `<a href="http://www.rsa.com/rsalabs/node.asp?id=2127">v2.0</a>` | Simplified tidylink |
| this PR | `v2.0[<a href="http://www.rsa.com/rsalabs/node.asp?id=2127">www.rsa.com/rsalabs/node.asp?id=2127</a>]` | Simplified tidylink changed |

### Regexp.html

| | expression | memo |
| --- | --- | --- |
| source | `(see {%r: Regexp Literals}[...])` | |
| master | `(see Regexp Literals):` | bug |
| this PR | `(see %r: Regexp Literals):` | fixed |

| | expression | memo |
| --- | --- | --- |
| source | `{rdoc-ref:Regexp global variables}[rdoc-ref:Regexp@Global+Variables]` | redundant label |
| master | `<a href="Regexp.html#class-regexp-global-variables">Regexp global variables</a>` | |
| this PR | `<a href="Regexp.html#class-regexp-global-variables">rdoc-ref:Regexp global variables</a>` | |

### String.html

| | expression | memo |
| --- | --- | --- |
| source | `<b>+String+ +replacement+</b>` | |
| master | `<a href="String.html"><code>String</code></a> <code>replacement</code></strong>` | `</strong>` without opening `<strong>` (bug) |
| this PR | `<strong><a href="String.html"><code>String</code></a> <code>replacement</code></strong>` | fixed |

| | expression | memo |
| --- | --- | --- |
| source | `{string literal}[rdoc-ref:syntax/literals.rdoc@String+Literals] or a {here document literal}[rdoc-ref:syntax/literals.rdoc@Here+Document+Literals]` | |
| master | `<p>string literal or a string literal</p>` | bug in cross_reference cache |
| this PR | `<p>string literal or a here document literal</p>` | bug is somehow avoided |

Same diff in `Regexp.html`

### StringScanner.html

| | expression | memo |
| --- | --- | --- |
| source | `[stored string][1]` | |
| master | `[stored <a href="1">string]</a>` | Considered simplified tidylink |
| this PR | `[stored string][1]` | Considered plain text |

(and many more. give up listing)

</detail>

### DIFF
--- a/lib/rdoc/markup.rb
+++ b/lib/rdoc/markup.rb
@@ -110,8 +110,8 @@
 
 class RDoc::Markup
 
-  ##
-  # An AttributeManager which handles inline markup.
+  # Array of regexp handling pattern and its name. A regexp handling
+  # sequence is something like a WikiWord
 
   attr_reader :regexp_handlings
 
--- a/lib/rdoc/markup/inline_parser.rb
+++ b/lib/rdoc/markup/inline_parser.rb
@@ -4,7 +4,7 @@ require 'set'
 require 'strscan'
 
 # Parses inline markup in RDoc text.
-# THis parser handles em, bold, strike, tt, hard break, and tidylink.
+# This parser handles em, bold, strike, tt, hard break, and tidylink.
 # Block-level constructs are handled in RDoc::Markup::Parser.
 
 class RDoc::Markup::InlineParser


lib/rdoc/markup/formatter.rb
```diff
@@ -121,117 +91,172 @@ def convert(content)
...
+      pos = text.size
+      matched_name = matched_text = nil
+      @markup.regexp_handlings.each do |pattern, name|
+        m = text.match(pattern, start)
+        idx = m && m[1] ? 1 : 0
+        if m && m.begin(idx) < pos
+          pos = m.begin(idx)
+          matched_text = m[idx]
+          matched_name = name
+        end
```
st0012, line 110:
```suggestion
        m = text.match(pattern, start)
        next unless m
        idx = m[1] ? 1 : 0
        if m.begin(idx) < pos
          pos = m.begin(idx)
          matched_text = m[idx]
          matched_name = name
        end
```

lib/rdoc/markup/to_html.rb
```diff
@@ -113,11 +111,112 @@ def handle_RDOCLINK(url) # :nodoc:
...
+  end
+
+  # Special handling for tidy link labels.
+  # When a tidy link is <tt>{rdoc-image:path/to/image.jpg:alt text}[http://example.com]</tt>,
+  # label part is normally considered RDOCLINK <tt>rdoc-image:path/to/image.jpg:alt</tt> and a text <tt>" text"</tt>
+  # but RDoc's test code expects the whole label part to be treated as RDOCLINK only in tidy link label.
+  # TODO: reconsider this workaround.
+
+  def apply_tidylink_label_special_regexp_handling(label)
+
```
st0012, line 187:
```suggestion
```

lib/rdoc/markup/formatter.rb
```diff
@@ -121,117 +91,172 @@ def convert(content)
...
-        yield tag
+  # Traverses +nodes+ and calls the appropriate handler methods
+  # Nodes formats are described in RDoc::Markup::InlineParser#parse
+
+  def traverse_inline_nodes(nodes)
+    nodes.each do |node|
+      next handle_TEXT(node) if String === node
+      case node[:type]
+      when :TIDYLINK
+        @in_tidylink = true
```
st0012: Should this be moved into `handle_TIDYLINK`?

lib/rdoc/markup/formatter.rb
```diff
@@ -121,117 +91,172 @@ def convert(content)
...
+  def handle_TIDYLINK(label_part, url)
+    traverse_inline_nodes(label_part)
   end
 
-  def each_attr_tag(attr_mask, reverse = false)
-    return if attr_mask.zero?
+  # Parses inline +text+, traverse the resulting nodes, and calls the appropriate handler methods.
+
+  def handle_inline(text)
+    @in_tidylink = false
```
st0012: Should this be set in `initialize`?

lib/rdoc/markup/inline_parser.rb
```diff
@@ -0,0 +1,321 @@
...
+  end
+
+  # Pop the top node off the stack when node is closed by a closing delimiter or an error.
+
+  def stack_pop
+    delimiter = @current[:delimiter]
+    @delimiters[delimiter].pop
+    @delimiters.delete(delimiter) if @delimiters[delimiter].empty?
+    @stack.pop
+    @current = @stack.last
```
st0012: Since we're already using a stack here, maybe we can just have a `current` method that returns `@stack.last`? That'll reduce a state we need to handle.

lib/rdoc/markup/to_ansi.rb
```diff
@@ -19,10 +19,43 @@ def initialize(markup = nil)
...
+    else
+      emit_inline("\e[#{[0, after].join(';')}m")
+    end
+  end
+
+  def add_text(text)
+    attrs = @attributes.keys
+    if @current_attributes != attrs && @current_attributes != attrs.sort!
+      change_attribute(@current_attributes, attrs)
+      @current_attributes = attrs
```
st0012: Should this be part of `change_attribute`?

lib/rdoc/markup/to_ansi.rb
```diff
@@ -19,10 +19,43 @@ def initialize(markup = nil)
...
-    add_tag :TT,   "\e[7m", "\e[m"
-    add_tag :EM,   "\e[4m", "\e[m"
+  ANSI_STYLE_CODES = {
+    BOLD: 1,
+    TT: 7,
+    EM: 4,
+    STRIKE: 9
+  }
+
+  def change_attribute(before, after)
```
st0012: This should be plural:

```suggestion
  def change_attributes(before, after)
```

lib/rdoc/markup/formatter.rb
tompng: Nice catch!
I've move this to `ToHtml#handle_TIDYLINK` and renamed ivar to `@in_tidylink_label`.

`@in_tidylink_label` is true while processing tidylink label-part, but false while processing tidylink url-part.
Because `rdoc-ref:...` in tidylink label should be ignored (to avoid generating nested `<a>` tag) but `rdoc-ref:...` inside tidylink url should be handled.

lib/rdoc/markup/to_ansi.rb
tompng: Done. The method and ivar name was misleading.

`@attributes` is the current attributes to be applied to `text`
`@current_attributes` (renamed to `@applied_attributes`) is the styling applied/flushed to output as escape sequences.
`on` and `off`  changes attribute
`change_attributes` (renamed to `apply_attributes`) will flush the buffered attribute changes

lib/rdoc/markup/to_ansi.rb
```diff
@@ -19,10 +19,48 @@ def initialize(markup = nil)
...
+    end
+    return if before == after
+
+    if after.empty?
+      emit_inline("\e[m")
+    elsif (before - after).empty?
+      emit_inline("\e[#{(after - before).join(';')}m")
+    else
+      emit_inline("\e[#{[0, after].join(';')}m")
+    end
```
nobu, line 55: [Select Graphic Rendition parameters](https://en.wikipedia.org/wiki/ANSI_escape_code#Select_Graphic_Rendition_parameters) can disable each attribute separately.

tompng: Updated
```ruby
"<b><i>a</i></b><s>b</s>" => "\e[1;4ma\e[0;9mb\e[m"
"<b><i>a</i><s>b</s></b>" => "\e[1;4ma\e[24;9mb\e[m"
```

lib/rdoc/markup.rb
```diff
@@ -113,7 +113,7 @@ class RDoc::Markup
   ##
   # An AttributeManager which handles inline markup.
```
st0012: This needs to be updated too

lib/rdoc/markup/inline_parser.rb
```diff
@@ -0,0 +1,311 @@
+# frozen_string_literal: true
+
+require 'set'
+require 'strscan'
+
+# Parses inline markup in RDoc text.
+# THis parser handles em, bold, strike, tt, hard break, and tidylink.
```
st0012:
```suggestion
# This parser handles em, bold, strike, tt, hard break, and tidylink.
```

tompng:
> I want to deprecate simple tidylink like word[url] in the next major release (I think they're not used THAT often?) to simplify things a bit.

I agree. Something like `Array[Object]` are frequently handled as tidylink and it's not what we expect.
One idea is to only accept `\A(https?://||rdoc-ref:)/` as url part.

> And in another major release we ban tidylink support from markdown

Looks good. We can add an escape converting markdown to rdoc (`{a}[b]` → `\{a\}\[b\]`)

PR #1542, opened 2026-01-04
Rebuild RDoc markup documentation 

st0012:
1. README now has a better section about supported markup formats
2. RDoc markup related reference are now consolidated under `doc/markup_reference/rdoc.rdoc`
3. Markdown markup now has a more comprehensive reference in `doc/markup_reference/markdown.md`

(#1536, #1538, #1540, and #1541 were all discovered when I was preparing for this PR 😄)

PR #1560, opened 2026-01-12
Fix broken RDoc Markup Reference links in RDoc::Markup

extern-c:
Fix broken RDoc Markup Reference links in RDoc::Markup.

See https://github.com/ruby/rdoc/pull/1542/files.


st0012 (approved):
Thanks

PR #1568, opened 2026-01-19
Create doc checker system

st0012:
When using RDoc's features, there are some mistakes that can easily be detected by RDoc itself, such as missing `rdoc-ref` targets.

Currently RDoc only prints warnings to stdout. But to better notify users, it should exit the process with a non-zero status code.

I plan to use this system to implement more checks in the future.

### DIFF
--- a/lib/rdoc/checker.rb
+++ b/lib/rdoc/checker.rb
@@ -1,0 +1,89 @@
+# frozen_string_literal: true
+
+##
+# RDoc::Checker collects documentation quality check warnings during RDoc
+# execution. Warnings are reported at the end and cause a non-zero exit status.
+#
+# Example usage:
+#   RDoc::Checker.add("Missing reference", file: "lib/foo.rb", line: 42)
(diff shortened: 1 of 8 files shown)


lib/rdoc/checker.rb
```diff
@@ -0,0 +1,89 @@
...
+# RDoc::Checker collects documentation quality check warnings during RDoc
+# execution. Warnings are reported at the end and cause a non-zero exit status.
+#
+# Example usage:
+#   RDoc::Checker.add("Missing reference", file: "lib/foo.rb", line: 42)
+#
+# At the end of rdoc execution, if any warnings were added, they will be
+# printed and the process will exit with status 1.
+
+class RDoc::Checker
```
kou, line 13: If we don't instantiate this, we don't need to use `class`. `module` is better.

(And I think that we can improve name of this... This doesn't check anything. This just keeps reported warnings.)

PR #1570, opened 2026-01-19
Move html characeter converting mechanism to regexp-handling

tompng:
`Text#to_html_characters` was a postprocess that converts ascii quotes/marks to multibyte characters. Postprocessing HTML to do that is not a good idea. Convert plain text node with regexp-handling is better.

### DIFF
--- a/lib/rdoc/markup/to_html.rb
+++ b/lib/rdoc/markup/to_html.rb
@@ -42,6 +42,86 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
 
   # :section:
 
+  # Maps an encoding to a Hash of characters properly transcoded for that
+  # encoding.
+  #
+  # See also encode_fallback.
+
+  TO_HTML_CHARACTERS = Hash.new do |h, encoding|
+    h[encoding] = {
+      close_dquote: encode_fallback('”', encoding, '"'),
+      close_squote: encode_fallback('’', encoding, '\''),
+      copyright:    encode_fallback('©', encoding, '(c)'),
+      ellipsis:     encode_fallback('…', encoding, '...'),
+      dot_ellipsis: encode_fallback('.…', encoding, '....'),
+      em_dash:      encode_fallback('—', encoding, '---'),
+      en_dash:      encode_fallback('–', encoding, '--'),
+      open_dquote:  encode_fallback('“', encoding, '"'),
+      open_squote:  encode_fallback('‘', encoding, '\''),
+      trademark:    encode_fallback('®', encoding, '(r)'),
+    }
+  end
+
+  HTML_CHARACTER_ALIASES = {
+    '(c)' => :copyright,
+    '(C)' => :copyright,
+    '(r)' => :trademark,
+    '(R)' => :trademark,
+    '---' => :em_dash,
+    '--'  => :en_dash,
+    '....' => :dot_ellipsis,
+    '...' => :ellipsis,
+    '``' => :open_dquote,
+    "''" => :close_dquote,
+  }
+
+  # Transcodes +character+ to +encoding+ with a +fallback+ character.
+
+  def self.encode_fallback(character, encoding, fallback)
+    character.encode(
+      encoding,
+      fallback: { character => fallback },
+      undef: :replace,
+      replace: fallback
+    )
+  end
+
+  # Converts ascii quote pairs to multibyte quote characters
+  class QuoteConverter
+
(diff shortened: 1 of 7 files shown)


lib/rdoc/text.rb
```diff
@@ -29,34 +29,6 @@ module RDoc::Text
...
-  # Maps an encoding to a Hash of characters properly transcoded for that
-  # encoding.
-  #
-  # See also encode_fallback.
-
-  TO_HTML_CHARACTERS = Hash.new do |h, encoding|
-    h[encoding] = {
-      :close_dquote => encode_fallback('”', encoding, '"'),
-      :close_squote => encode_fallback('’', encoding, '\''),
-      :copyright    => encode_fallback('©', encoding, '(c)'),
```
tompng, line 42: I think html character is `&copy;`, not `©`, but it's another issue.

lib/rdoc/markup/to_html.rb
```diff
@@ -40,6 +40,82 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
...
   # :section:
 
+  # Maps an encoding to a Hash of characters properly transcoded for that
+  # encoding.
+  #
+  # See also encode_fallback.
+
+  TO_HTML_CHARACTERS = Hash.new do |h, encoding|
+    h[encoding] = {
+      :close_dquote => encode_fallback('”', encoding, '"'),
```
st0012: Can we use Ruby 1.9 hash syntax?

lib/rdoc/markup/to_html.rb
```diff
@@ -40,6 +40,82 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
...
+      :close_squote => encode_fallback('’', encoding, '\''),
+      :copyright    => encode_fallback('©', encoding, '(c)'),
+      :ellipsis     => encode_fallback('…', encoding, '...'),
+      :dot_ellipsis     => encode_fallback('.…', encoding, '....'),
```
st0012: nit: I think we can align this one too:

```suggestion
      :dot_ellipsis => encode_fallback('.…', encoding, '....'),
```

lib/rdoc/markup/to_html.rb
```diff
@@ -40,6 +40,82 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
...
+    '....' => :dot_ellipsis,
+    '...' => :ellipsis,
+    '``' => :open_dquote,
+    "''" => :close_dquote,
+  }
+
+  # Transcodes +character+ to +encoding+ with a +fallback+ character.
+
+  def self.encode_fallback(character, encoding, fallback)
+    character.encode(encoding, :fallback => { character => fallback },
```
st0012: nit: 1.9 hash syntax. Or we can do it via a cop in a separate PR too

lib/rdoc/markup/to_html.rb
```diff
@@ -40,6 +40,82 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
...
+        end
+      when '`'
+        # Opening quote of <tt>`quoted sentence'</tt>.
+        # This will conflict with code blocks <tt>`puts('hello')`</tt> in the future.
+        if !@insquotes && !after_word
+          type = :open_squote
+          @insquotes = true
+        end
+      end
+      TO_HTML_CHARACTERS[quote.encoding][type] if type
```
st0012, line 121: Could `TO_HTML_CHARACTERS[quote.encoding]` ever be `nil` and causes NoMethodError?

lib/rdoc/markup/to_html.rb
```diff
@@ -40,6 +40,82 @@ class RDoc::Markup::ToHtml < RDoc::Markup::Formatter
...
+      @in_squote = false
+    end
+
+    def convert(quote, after_word:)
+      case quote
+      when '"'
+        type = @in_dquote ? :close_dquote : :open_dquote
+        @in_dquote = !@in_dquote
+      when "'"
+        if @insquotes
```
st0012: Should these be `@in_squote`?

st0012 (approved):

lib/rdoc/markup/to_html.rb
tompng: `TO_HTML_CHARACTERS = Hash.new do |h, encoding| ... end` will generate a new hash if there is no key.

PR #1577, opened 2026-01-22
Fix a test that uses invalid syntax

Earlopain:
Writing it like this on the same line was never valid. The expected html also contains the `end` from the class definition, which I think should not happen.


tompng (approved):
Thank you!

PR #1580, opened 2026-01-24
Change PrismRuby not to depend on hack that stores module nesting information to context.parent

### LINKED ISSUE #1291: RDoc::Mixin needs module nesting information
This two `include M` are not the same.
```ruby
class A::B
  # A::B.include(M) with Module.nesting == [A::B]
  include M
end

class A
  class B
    # A::B.include(M) with Module.nesting == [A::B, A]
    include M
  end
end
```

RDoc::Mixin needs information of module nesting for correct module name lookup.

Currently, RDoc uses `CodeObject#parent` to represent module nesting but it changes while parsing.
Resolving module name of include/extend is done after all files are parsed. It is incorrect because it might be overwritten while parsing another file. We need something like `RDoc::Mixin#module_nesting`.
```ruby
class A::B
  # parent of A::B is TopLevel
  include M
end

class A
  class B
    # parent of A::B changes to A
  end
end
```

tompng:
Closes #1291

All constant/module/class paths are resolved by PrismRuby parser itself.
Resolves include/extend module name before adding it because `Context#parent` will not have an information for delayed full-path name resolve.

### DIFF
--- a/lib/rdoc/code_object/context.rb
+++ b/lib/rdoc/code_object/context.rb
@@ -785,7 +785,10 @@ class RDoc::Context < RDoc::CodeObject
   end
 
   ##
-  # Find a module at a higher scope
+  # Tries to find a module at a higher scope.
+  # But parent is not always a higher module nesting scope, so the result is not correct.
+  # Parent chain can only represent last-opened nesting, and may be broken in some cases.
+  # PrismRuby parser stopped representing module nesting with parent chain at all.
 
   def find_enclosing_module_named(name)
     parent && parent.find_module_named(name)
@@ -860,13 +863,21 @@ class RDoc::Context < RDoc::CodeObject
   end
 
   ##
-  # Find a module with +name+ using ruby's scoping rules
+  # Find a module with +name+ trying to using ruby's scoping rules.
+  # find_enclosing_module_named cannot use ruby's scoping so the result is not correct.
 
   def find_module_named(name)
-    res = @modules[name] || @classes[name]
+    res = get_module_named(name)
     return res if res
     return self if self.name == name
     find_enclosing_module_named name
+  end
+
+  # Get a module named +name+ in this context
+  # Don't look up for higher module nesting scopes. RDoc::Context doesn't have that information.
+
+  def get_module_named(name)
+    @modules[name] || @classes[name]
   end
 
   ##
--- a/lib/rdoc/parser/prism_ruby.rb
+++ b/lib/rdoc/parser/prism_ruby.rb
@@ -59,11 +59,6 @@ class RDoc::Parser::PrismRuby < RDoc::Parser
     @container = container
     @singleton = singleton
     @in_proc_block = false
-    unless singleton
-      # Need to update module parent chain to emulate Module.nesting.
-      # This mechanism is inaccurate and needs to be fixed.
-      container.parent = old_container
-    end
     @module_nesting.push([container, singleton])
     yield container
   ensure
@@ -482,12 +477,15 @@ class RDoc::Parser::PrismRuby < RDoc::Parser
     end
   end
 
+  # Adds includes/extends. Module name is resolved to full before adding.
+
   def add_includes_extends(names, rdoc_class, line_no) # :nodoc:
     return if @in_proc_block
     comment, directives = consecutive_comment(line_no)
     handle_code_object_directives(@container, directives) if directives
     names.each do |name|
-      ie = @container.add(rdoc_class, name, '')
+      resolved_name = resolve_constant_path(name)
+      ie = @container.add(rdoc_class, resolved_name || name, '')
       ie.store = @store
       ie.line = line_no
       ie.comment = comment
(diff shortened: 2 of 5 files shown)


lib/rdoc/parser/prism_ruby.rb
```diff
@@ -59,11 +59,6 @@ def with_container(container, singleton: false)
     @container = container
     @singleton = singleton
     @in_proc_block = false
-    unless singleton
-      # Need to update module parent chain to emulate Module.nesting.
-      # This mechanism is inaccurate and needs to be fixed.
-      container.parent = old_container
-    end
```
tompng, line 66: `RDoc::Parser::Ruby` updates `parent` only for NormalClass. This information is used by includes/extends name resolve and also in CrossReference resolve.
This is one of the reason that HTML generated by `RDoc::Parser::PrismRuby` and `RDoc::Parser::Ruby` has huge diffs.
And there is no way to reduce diff except re-implementing many bugs.

tompng: Changing this part to
```ruby
container.parent = old_container if !singleton && container.is_a?(RDoc::NormalClass)
```
will minimize crossref difference of generated HTML between two parsers.
(but this change doesn't make sense. It's just re-implementing a bug)

lib/rdoc/code_object/context.rb
```diff
@@ -860,7 +861,8 @@ def find_method_named(name)
   end
 
   ##
-  # Find a module with +name+ using ruby's scoping rules
+  # Find a module with +name+ trying to using ruby's scoping rules.
+  # find_enclosing_module_named cannot use ruby's scoping so the result is not correct.
 
   def find_module_named(name)
     res = @modules[name] || @classes[name]
```
st0012: Should this now call `get_module_named` instead?

lib/rdoc/parser/prism_ruby.rb
st0012: Can you this some examples of the diffs caused by this? It's not entirely clear to me.

lib/rdoc/code_object/context.rb
```diff
@@ -785,7 +785,8 @@ def find_constant_named(name)
   end
 
   ##
-  # Find a module at a higher scope
+  # Tries to find a module at a higher scope.
+  # But parent is not always a higher module nesting scope, so the result is not correct.
```
st0012, line 789: Can we add examples here?

Looks like it's for

```rb
module Foo
  module Bar; end
  module Baz; end # we're at this scope and we try to find Bar?
end
```

But in some cases this might not work either?

lib/rdoc/parser/prism_ruby.rb
tompng: This is really complicated. In `RDoc::Parser::Ruby` parser, parent chian is updated like this.
```ruby
class A; class B; class C; class D; class E; end; end; end; end; end
# Parent chain is:
# F → E → D → C → B → A → toplevel

class A::B::C
  # Method comment A::B B::C C::D D::E
  # (B::C is not linked because RDoc only search A::B::C and toplevel)
  def f; end
  module D::E
  end
end

# Parent chain is updated. This imitates module nesting
# E → C → toplevel (changed)
# B → A → toplevel, D → C (unchanged)
# `B::C` in method comment is not linked because it can't be found from C
```
Same situation for modules. It's strange that  parent chain is not updated. Maybe bug or workaround for something.
```ruby
module M; module N; module O; module P; module Q; end; end; end; end; end
# Parent chain is:
# R → Q → P → O → N → M → toplevel

module M::N::O
  # Method comment M::N N::O O::P P::Q
  # All linked because RDoc searches M::N::O, M::N, M and toplevel
  def f; end
end
```

Parent chain is emulating module nesting, used in crossref resolve, include/extend name resolve, superclass name resolve, etc.
Parent chain update is skipped for modules, so module parent seems to represent module owner. This inconsistency causes many bugs, but this also makes crossref work better in some cases.
PrismRuby parser updated both module and class to emulate module nesting, so crossref behavior got worse as a bugfix.

In this pull request, parent will no longer updated. Parent will exactly represent module/class owner, not module nesting.
As a result, `B::C` in the method comment will be linked too. It means crossref doesn't follow module nesting.
But I think it's OK, some comments already don't follow module nesting, and already linked.
```ruby
class RDoc::TopLevel < RDoc::Context
  ##
  # Returns the NormalClass "Object", creating it if not found.
  # (NormalClass can't be resolved in this module nesting but we often write this kind of comment)

  def object_class
  end
end
```

| | RDoc::Parser::Ruby | PrismRuby(master) | PrismRuby (this PR) |
| --- | --- | --- | --- |
| Linked classes | `A::B` `C::D` `D::E` | `A::B` `C::D` `D::E` | `A::B` `B::C` `C::D` `D::E` |
| Linked modules | `M::N` `N::O` `O::P` `P::Q` | `M::N` `O::P` `P::Q` | `M::N` `N::O` `O::P` `P::Q` |

lib/rdoc/code_object/context.rb
tompng: Comment added.

There are three cases I know.
1. PrismRuby stopped updating parent chain in this PR.
2. There's a bug updating parent chain in RDoc::Parser::Ruby, (not updated when opening module)
3. Module nesting will change with opening class, but parent chain can only represent one global module nesting.
```ruby
class A::B::C
  class D::E
    # module nesting 1
    include M
  end
end

class A::B
  class C::D::E
    # module nesting 2
    include N
  end
end
# Only the last opened module nesting will be used to resolve crossref and include/extend
```

PR #1587, opened 2026-01-30
Fix not scrolling in search results when `ArrowUp` or `ArrowDown` for `Aliki`

ccmywish:
If we use Up arrow key or Down arrow key in search results, the scroll bar doesn't scroll, instead, the scroll bar of the whole page scroll. 

This PR fixes it.


ccmywish:
@st0012 

This is not a big change. Could you please take a look and review it?

PR #1588, opened 2026-01-30
Add example of :method: directive getting ignored.

p8:
Also make the comment about hidden methods slightly clearer — these
methods don't appear in the code, but they appear in the generated
documentation.

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -131,8 +131,8 @@ require_relative 'ripper_state_lex'
 #
 # == Hidden methods and attributes
 #
-# You can provide documentation for methods that don't appear using
-# the :method:, :singleton-method: and :attr: directives:
+# You can provide documentation for methods that don't appear in the code
+# using the :method:, :singleton-method: and :attr: directives:
 #
 #   ##
 #   # :attr_writer: ghost_writer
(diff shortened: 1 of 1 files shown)


lib/rdoc/parser/ruby.rb
```diff
@@ -131,8 +131,8 @@
 #
 # == Hidden methods and attributes
 #
-# You can provide documentation for methods that don't appear using
-# the :method:, :singleton-method: and :attr: directives:
+# You can provide documentation for methods that don't appear in the code
+# using the :method:, :singleton-method: and :attr: directives:
```
tompng, line 135: This part looks good to me

lib/rdoc/parser/ruby.rb
```diff
@@ -148,8 +148,13 @@
...
 # Note that by default, the :method: directive will be ignored if there is a
-# standard rdocable item following it.
-
+# standard rdocable item following it:
+#
+#   ##
+#   # :method: ghost_method
+#   # ghost_method will be ignored.
+#
+#   def standard_rdocable_method() end
```
tompng, line 157: This part seems it's just describing a bug or unintentional behavior in detail.

In this example, HTML contains `":method: ghost_method"` as a plain text. Looks like a bug.
```ruby
##
# :method: ghost_method
# ghost_method will be ignored.
def standard_rdocable_method() end # rdocable item
```

In this example, `:method:` directive is ignored, nothing is documented. I also think this is a bug too.
```ruby
##
# :method: foobarbaz_foo
attr_reader("#{PREFIX}_foo") # RDoc::Parser::Ruby thinks this is an rdocable item
```

In both cases, result will changed in the experimental (and the next default) `RDoc::Parser::PrismRuby` parser.

PR #1595, opened 2026-02-04
Ignore visibility method, attr definition, module_function within block

tompng:
We need to ignore these within `Module.new do end` and any other block because it might be a metaprogramming block.

Ignoring `def` `include` `extend` in a block is already implemented. This pull request also make RDoc ignore visibility methods, attr definition and module_function in a block.

```ruby
class A
  def f; end
  X = 1
  anonymous_module = Module.new do
    # These visibility change and attribute defition should be applied to anonymous module, but applied to A
    module_function :f
    private :f
    private_constant :X
    attr_accessor :rw
  end
end
```

Found while generating document in rails/rails
https://github.com/rails/rails/blob/45ee3bf84f20af55b8619b5dcc22a5e22a4ac0e7/activerecord/lib/active_record/connection_adapters/abstract/schema_definitions.rb#L307
```ruby
module ColumnMethods
  class_methods do
    private
    def m; end # This is a private method of unknown class/module
  end

  def g; end # This method should be public, but was treated as private and not documented
end
```

### DIFF
--- a/lib/rdoc/parser/prism_ruby.rb
+++ b/lib/rdoc/parser/prism_ruby.rb
@@ -743,11 +743,14 @@ class RDoc::Parser::PrismRuby < RDoc::Parser
         when :extend
           _visit_call_extend(node)
         when :public
-          _visit_call_public_private_protected(node, :public) { super }
+          super
+          _visit_call_public_private_protected(node, :public)
         when :private
-          _visit_call_public_private_protected(node, :private) { super }
+          super
+          _visit_call_public_private_protected(node, :private)
         when :protected
-          _visit_call_public_private_protected(node, :protected) { super }
+          super
+          _visit_call_public_private_protected(node, :protected)
         when :private_constant
           _visit_call_private_constant(node)
         when :public_constant
@@ -757,11 +760,14 @@ class RDoc::Parser::PrismRuby < RDoc::Parser
         when :alias_method
           _visit_call_alias_method(node)
         when :module_function
-          _visit_call_module_function(node) { super }
+          super
+          _visit_call_module_function(node)
         when :public_class_method
-          _visit_call_public_private_class_method(node, :public) { super }
+          super
+          _visit_call_public_private_class_method(node, :public)
         when :private_class_method
-          _visit_call_public_private_class_method(node, :private) { super }
+          super
+          _visit_call_public_private_class_method(node, :private)
         else
           super
         end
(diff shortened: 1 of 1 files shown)


lib/rdoc/parser/prism_ruby.rb
```diff
@@ -1021,12 +1022,16 @@ def _visit_call_public_private_protected(call_node, visibility)
     end
 
     def _visit_call_alias_method(call_node)
+      return if @scanner.in_proc_block
```
st0012, line 1029: Should the same check be added to `visit_alias_method_node` too?

test/rdoc/parser/prism_ruby_test.rb
```diff
@@ -2032,6 +2032,93 @@ class B
...
+      class A
+        attr_reader :r
+        attr_writer :w
+        attr_accessor :rw
+        Module.new do
+          attr_reader :r2
+          attr_writer :w2
+          attr_accessor :rw2
+        end
+        alias_method :foo3, :foo
```
st0012: Probably not needed?

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -997,19 +997,20 @@ def _visit_call_require(call_node)
...
     def _visit_call_module_function(call_node)
       yield
-      return if @scanner.singleton
+      return if @scanner.in_proc_block || @scanner.singleton
       names = visibility_method_arguments(call_node, singleton: false)&.map(&:to_s)
       @scanner.change_method_to_module_function(names) if names
     end
 
     def _visit_call_public_private_class_method(call_node, visibility)
       yield
```
st0012: Not directly, related, but why does this method call `yield` here but `_visit_call_public_private_protected` calls it in a branch?

lib/rdoc/parser/prism_ruby.rb
tompng: Nice catch! Fixed

lib/rdoc/parser/prism_ruby.rb
tompng: No specific reason. I think it's better to make it align.
If we're not calling `yield` (block is `{ super }`) conditionally, calling `super` before calling this method is better.

I added this change:
```ruby
_visit_call_public_private_class_method(args) { super }

def _visit_call_public_private_class_method(call_node, visibility)
  yield
  ...
end
# ↓
super
_visit_call_public_private_class_method(args)

def _visit_call_public_private_class_method(call_node, visibility)
  ...
end
```

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -43,9 +43,10 @@ def initialize(top_level, content, options, stats)
   # example: `Module.new { include M }` `M.module_eval { include N }`
 
   def with_in_proc_block
+    in_proc_block = @in_proc_block
     @in_proc_block = true
     yield
-    @in_proc_block = false
+    @in_proc_block = in_proc_block
```
tompng, line 49:
```ruby
tap do
  include A # in_proc_block = true
  tap do end # Exiting this block was making `@in_proc_block = false
  include B # in_proc_block was false, fixed to true
end
```

PR #1600, opened 2026-02-07
Fix sections not displaying their text

### LINKED ISSUE #345: Section comments are not shown in HTML output
When commenting a section like

```
# :section: Some Section
# Some documentation for the section here
```

the section is created but the documentation is not shown in the HTML output.

You can actually see the problem in the RDoc documentation itself:
- https://github.com/rdoc/rdoc/blob/master/lib/rdoc/markup/to_html.rb#L148 - definition of "Visitor" section with some documentation
- http://docs.seattlerb.org/rdoc/RDoc/Markup/ToHtml.html#Visitor - Output with missing documentation

Thanks for this great tool!

Earlopain:
I'm currently trying to group methods in the prism docs by topic like "location related", " flag related", etc. so that methods not related to these conceps show up more prominently (not mixed in with potentially many other methods). But I noticed that sections do not currently show their comment.

There were a few things wrong that made this not work:
* Formatting expects `@comment` but there is only `@comments`
* The parse method used for marshaling overwrote the markdown parse method
* `@store` was not present, which the formatter requires

Comments can be added/removed, so there still needs to be the array to make it possible to properly modify the section.

Fixes #345, Closes #936

On https://docs.ruby-lang.org/en/master/Prism/Node.html#node-interface there is actually already a comment but it gets swallowed. With this change, it gets properly rendered:

image

### DIFF
--- a/lib/rdoc/code_object/context/section.rb
+++ b/lib/rdoc/code_object/context/section.rb
@@ -18,11 +18,6 @@ class RDoc::Context::Section
   MARSHAL_VERSION = 0 # :nodoc:
 
   ##
-  # Section comment
-
-  attr_reader :comment
-
-  ##
   # Section comments
 
   attr_reader :comments
@@ -38,11 +33,17 @@ class RDoc::Context::Section
   attr_reader :title
 
   ##
+  # The RDoc::Store for this object.
+
+  attr_reader :store
+
+  ##
   # Creates a new section with +title+ and +comment+
 
-  def initialize(parent, title, comment)
+  def initialize(parent, title, comment, store = nil)
     @parent = parent
     @title = title ? title.strip : title
+    @store = store
 
     @comments = []
 
@@ -151,7 +152,7 @@ class RDoc::Context::Section
     [
       MARSHAL_VERSION,
       @title,
-      parse,
+      to_document,
     ]
   end
 
@@ -169,7 +170,7 @@ class RDoc::Context::Section
   # Parses +comment_location+ into an RDoc::Markup::Document composed of
   # multiple RDoc::Markup::Documents with their file set.
 
-  def parse
+  def to_document
     RDoc::Markup::Document.new(*@comments.map(&:parse))
   end
 
@@ -180,6 +181,23 @@ class RDoc::Context::Section
 
   def plain_html
     @title || 'Top Section'
+  end
+
+  ##
+  # Section comment
+
+  def comment
+    return nil if @comments.empty?
+    RDoc::Comment.from_document(to_document)
+  end
+
+  def description
+    return '' if @comments.empty?
+    markup comment
+  end
+
+  def language
+    @comments.first&.language
   end
 
   ##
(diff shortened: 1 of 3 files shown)


lib/rdoc/code_object/context/section.rb
```diff
@@ -182,6 +190,12 @@ def plain_html
     @title || 'Top Section'
   end
 
+  def markup(comment)
+    # Don't try to format the default section. We don't have a store yet.
+    return "" if comment.nil? || comment.empty?
+    super(normalize_comment(comment.text))
```
tompng: This will double-normalize a comment if comment is already normalized.
```
# :section: Foo
# # aaa (should be "# aaa")
```
`RDoc::Text#markup` will accept `RDoc::Comment`, so we don't need to use `comment.text`.
Instead of overriding `markup`, how about overriding `description` like this?

```ruby
def description
   markup comment if @comments.any?
end
```

For reference, `RDoc::ClassModule` overrides `description` too
```ruby
class RDoc::ClassModule

  ##
  # Handy wrapper for marking up this class or module's comment

  def description
    markup @comment_location
  end

end
```

lib/rdoc/code_object/context/section.rb
```diff
@@ -190,6 +204,18 @@ def remove_comment(target_comment)
...
+    set_comment!
   end
 
+  private
+
+  def set_comment!
+    if @comments.any?
+      donor = @comments[0]
+      @language = donor.language
+      @comment = RDoc::Comment.new(@comments.map(&:text).join, donor.location)
```
tompng: If a section comment is separated, I think it should be separated with `<p>` tag.
```ruby
# Document1
class A
end

# Document2
class A
end
```
will generate
```html
<h1 id="class-a" class="anchor-link class">class A</h1>
<section class="description">
<p>Foo</p>
<p>Bar</p>
</section>
```
So
```ruby
class A
  # :section: Section1
  # hello

  # :section: Section1
  # world
end
```
is better to generate
```html
<p>hello</p><p>world</p>
```

One way to create a normalized separated comment is to use:
```ruby
RDoc::Comment.from_document(to_document)
```
This will create a concatenated normalized comment. `comment.text` is nil, and the parsed result `comment.instance_variable_get(:@document)` is present.

tompng: I think using both `@comments` and `@comment` is not so great.
Renaming this method to `comment`, just return a comment without storing to `@comment` will make the state of this class more simple, I think. To do this, we need to modify `def description`.

Earlopain: :+1:, I applied your suggestion and it's a bit cleaner now.

I removed the tests I added about `comment` in `test_add_comment`. Since `from_document` always sets the comment text to the empty string, they don't hold anymore. I'm not sure how it works but it seems to work out in the end.

lib/rdoc/code_object/context/section.rb
Earlopain: Works nicely, thanks

tompng (approved):
Looks good. Thank you :+1:

PR #1601, opened 2026-02-07
Add sections to the TOC

### LINKED ISSUE #1598: Sections not part of TOC
On https://docs.ruby-lang.org/en/master/Prism/Node.html, there is a section called `Node interface`. It's shown in the left navigation menu but is missing from the TOC. So you get "Public Class Methods" and "Public Instance Methods" twice without any indication of why (I first thought it was a bug)

Image

I feel like they should instead be indented like
* Node interface
* * Public Class Methods
* * Public Instance Methods

so that you can see the intended structure.

https://github.com/ruby/rdoc/issues/345 😢 but that's something else entirely

Earlopain:
Closes #1598

They are currently in the left sidebar but they don't make much sense there I believe. Now they are in the TOC with the other headers.

If a heading is part of a section, its indentation is incremented by one level.

Before:

image

After:

image


st0012:
Thanks for reporting the issue and the PR 🙏 But I think we may be able to fix it directly in the template: #1603

Earlopain:
Clearly the better solution. Thanks :+1:

PR #1603, opened 2026-02-08
Add unique IDs to section headings in Aliki template

### LINKED ISSUE #1598: Sections not part of TOC

st0012:
The Aliki TOC is generated by JavaScript that scans h1/h2/h3 elements with IDs. Section title h2 elements had no ID, so they were invisible to the TOC. Meanwhile, h3 IDs like "constants" and "public-instance-methods" were duplicated across sections.

Prefix all heading IDs with the section's aref to make them unique. Move the section ID from the `<section>` element to the `<h2>` so the TOC picks it up naturally.

Screenshot 2026-02-08 at 10 20 30

Fixes #1598

### DIFF
--- a/lib/rdoc/generator/template/aliki/class.rhtml
+++ b/lib/rdoc/generator/template/aliki/class.rhtml
@@ -40,10 +40,10 @@
 
   <%- klass.each_section do |section, constants, attributes| %>
   <span id="<%= section.legacy_aref %>" class="legacy-anchor"></span>
-  <section id="<%= section.aref %>" class="documentation-section anchor-link">
+  <section class="documentation-section anchor-link">
     <%- if section.title then %>
     <header class="documentation-section-title">
-      <h2>
+      <h2 id="<%= section.aref %>">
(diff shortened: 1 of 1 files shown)


Earlopain:
Do you want to keep sections in the left sidebar? It feels redundant with the TOC (which itself is also not present in the sidebar)

st0012:
ToC is currently hidden on mobile. So if we removed it from sidebar, the sections info would just be unavailable on mobile.
I wouldn't fight hard to keep it if users think the sections section don't provide much value, but IMO that should be a separate discussion and change.

tompng (approved):

Issue #1286, opened 2025-02-05
Alias for attribute causes aliasing method to appear in Attributes section


gemmaro:
When defining an alias for an attribute in a class, the aliasing method is incorrectly displayed in the attributes section of the generated documentation.

For example, consider the following code in `lib/foo.rb`:

```ruby
class Foo
  attr :bar
  alias baz bar
end
```

When running `rdoc`, the documentation for the `Foo` class shows:

![Image](https://github.com/user-attachments/assets/b74e722a-cec1-4aec-8b74-c0c696112b50)

This behavior is misleading, as `baz` is not a separate attribute but an alias method for `bar` method (and attribute). It would be beneficial to adjust the documentation generation to prevent alias methods from appearing in the attributes section.

flavorjones:
Validated this behavior on v6.11.0.

@gemmaro Can you help me understand what you expect to see? It sounds like you want to completely omit mentioning the alias, but I think there's value in mentioning it as an alias, perhaps similar to how rdoc renders non-attribute methods:

![Image](https://github.com/user-attachments/assets/8b4d3381-2f22-4c9f-86e4-50caa41f1839)

WDYT?

gemmaro:
@flavorjones  I agree that having links between aliasing methods and aliased attributes is beneficial.  Here's what I expect to see, based on the Darkfish theme (though it may vary in other themes):

- Aliasing method names should be listed under the instance methods section (e.g., "Public Instance Methods").
- Each aliasing method should have its own document comment.
- The description of the aliasing method should include a link to the aliased instance variable, labeled as "Alias for:".
- The aliased attribute should link back to the aliasing methods, indicated as "Also aliased as:".
- The left side menu should include the aliasing method under the "Methods" section.

For example, consider the following `lib/foo.rb`:

```ruby
class Foo
  # Some document comment about bar.
  attr :bar

  # Some document comment about baz.
  alias baz bar

  # Some document comment about qux.
  def qux; end

  # Some document comment about quux.
  alias quux qux
end
```

Currently, RDoc generates:

![actual Image](https://github.com/user-attachments/assets/72c98af4-5831-4f33-a152-a857bb16687a)

What I expect to see is:

![expected Image](https://github.com/user-attachments/assets/bcff9478-5b65-49b0-a1bd-f98ba122e1df)

<details>
<summary>manually edited diff for <code><var>op_dir</var>/Foo.html</code></summary>

```diff
[...]

PR #1605, opened 2026-02-08
Fix RDoc::Attr#add_alias handling of aliased attribute accessor

extern-c:
Ensure aliased attr_accessor attributes have the correct access type ('R' or 'W') instead of always being 'RW'.

For more details on the issue being addressed, see https://github.com/ruby/rdoc/issues/1286#issuecomment-3794521252.


tompng (approved):
Looks good :+1:

PR #1613, opened 2026-02-16
Respect markup comment in non-ruby files

### LINKED ISSUE #1597: RDoc parses `.c` file as ruby code if markup is set to markdown in the first 3 lines
This file is parsed as ruby.
```c
/* file.c */
/* :markup: markdown */
/* ↑ deleting the above line will make this file parsed as c */

#include <stdio.h>

#define def void
def foobar() {
}

/* method comment */
VALUE rb_a_foo(VALUE self) {
}

int main(){
  /* class comment */
  klass = rb_define_class("A", rb_cObject);
  rb_define_method(klass, "f", rb_a_foo, 0);
}
```

https://github.com/ruby/rdoc/blob/5b15418871726b75a8d2d34e12a12e2558a9a445/lib/rdoc/parser.rb#L233-L238

Earlopain:
Fixes #1597

Prism uses markdown for documentation. In the prism repo itself we can set `rdoc.markup = "markdown"` but that doesn't work in ruby/ruby. So I wanted to add the markup directive to the C files that are currently missing them. But it is not respected.

### DIFF
--- a/lib/rdoc/parser.rb
+++ b/lib/rdoc/parser.rb
@@ -170,7 +170,7 @@ class RDoc::Parser
     file_name = top_level.absolute_name
     return if binary? file_name
 
-    parser = use_markup content
+    parser = use_markup content, file_name
 
     unless parser then
       parse_name = file_name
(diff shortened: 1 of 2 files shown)


lib/rdoc/parser.rb
```diff
@@ -229,21 +229,22 @@ def self.remove_modeline(content)
   #
   # Any comment style may be used to hide the markup comment.
 
-  def self.use_markup(content)
+  def self.use_markup(content, file_path)
     markup = content.lines.first(3).grep(/markup:\s+(\w+)/) { $1 }.first
 
-    return unless markup
+    if markup
+      return RDoc::Parser::Ruby if markup == "tomdoc"
```
Earlopain, line 236: There doesn't seem to be a tomdoc parser. So I kept this for backwards compatibility.

lib/rdoc/parser.rb
```diff
@@ -229,21 +229,22 @@ def self.remove_modeline(content)
...
-    _, selected = RDoc::Parser.parsers.find do |_, parser|
-      /^#{markup}$/i =~ parser.name.sub(/.*:/, '')
+      markup = Regexp.escape markup
+      _, selected = RDoc::Parser.parsers.find do |_, parser|
+        /^#{markup}$/i =~ parser.name.sub(/.*:/, '')
+      end
+      selected
+    else
+      selected = RDoc::Parser.can_parse_by_name(file_path)
+      return if selected == RDoc::Parser::Simple
```
Earlopain, line 245: This one has an empty regexp and would match on any file. This breaks shebang detection, so I explicitly exclude it here.

Earlopain:
It doesn't look like I know what I'm doing. markup is used to control both the parser and markup format and I have no idea what to do about that.

PR #1620, opened 2026-02-21
Add server mode with live reload (`rdoc --server`)

st0012:
A better attempt of #1151

Implement `rdoc --server[=PORT]` for previewing documentation with automatic browser refresh on source file changes. The server parses all sources on startup, serves pages from memory via the Aliki generator, and watches for file modifications, additions, and deletions — re-parsing only what changed.

https://github.com/user-attachments/assets/06d5ae7b-cc54-4cc5-a8b0-1cc30177cadb

## Changes

- **New `RDoc::Server`** (`lib/rdoc/server.rb`) — minimal HTTP server using Ruby's built-in `TCPServer` (no WEBrick or external dependencies)
  - Thread-per-connection with `Connection: close`
  - Persistent Aliki generator instance rendering pages to strings
  - In-memory page cache with full invalidation on changes
  - Live reload via inline JS polling `/__status` endpoint every 1 second
  - Background file watcher polling source file mtimes every 1 second
  - Incremental re-parse: only changed files are re-parsed, old data removed first
  - Detection of new and deleted source files
- **`--server[=PORT]`** CLI option (default port 4000) and `rdoc:server` Rake task
- **`RDoc::Store#remove_file`** — removes a file's entries from the store hashes
- **`RDoc::Store#clear_file_contributions`** — surgically removes a file's methods, constants, comments, includes, extends, and aliases from its classes/modules, preserving classes that span multiple files. Supports `keep_position:` for server re-parse to preserve comment ordering.
- **`RDoc::RDoc#relative_path_for`** — extracted path normalization (against `options.root` and `options.page_dir`) shared by `parse_file` and the server
- **`Darkfish#refresh_store_data`** — extracted for reuse by the server after re-parsing
- **`RDoc::Servlet` → `RDoc::RI::Servlet`** — moved to clarify RI-specific usage

## Security & robustness

- Binds to `127.0.0.1` only (localhost)
- Path traversal protection in asset serving (`File.expand_path` containment check)
- Proper HTTP error responses: 400, 404, 405, 500
- 5-second `IO.select` read timeout on client sockets
- Mutex protects all store mutations, generator refresh, and cache invalidation atomically
- Individual `parse_file` errors rescued so one failure doesn't block remaining files
- Watcher thread uses `@running` flag with clean shutdown via `Thread#join`

## Known limitations

- **Full cache invalidation:** any file change clears all cached pages. Rendering is fast (~ms per page); parsing is the expensive part and is done incrementally.
- **Template/CSS changes:** require server restart (only source files are watched).

### DIFF
--- a/lib/rdoc/server.rb
+++ b/lib/rdoc/server.rb
@@ -363,6 +363,8 @@ class RDoc::Server
             $stderr.puts "Error parsing #{f}: #{e.message}"
           end
         end
+
+        @store.cleanup_stale_contributions
       end
 
       @store.complete(@options.visibility)
--- a/lib/rdoc/store.rb
+++ b/lib/rdoc/store.rb
@@ -280,6 +280,35 @@ class RDoc::Store
   end
 
   ##
+  # Removes stale empty placeholders left by +clear_file_contributions+ with
+  # <tt>keep_position: true</tt>.  After re-parsing, a file may no longer
+  # define a class it previously contributed to, leaving an empty entry in
+  # +comment_location+ and a stale +in_files+ reference.  Call this after
+  # all re-parsing is complete.
+
+  def cleanup_stale_contributions
+    all_classes_and_modules.each do |cm|
+      cm.comment_location.delete_if { |_, comments| comments.empty? }
+      cm.rebuild_comment_from_location
+
+      cm.in_files.select! { |tl| cm.comment_location.key?(tl) ||
+        cm.method_list.any? { |m| m.file == tl } ||
+        cm.attributes.any? { |a| a.file == tl } ||
+        cm.constants.any? { |c| c.file == tl } }
+
+      if cm.in_files.empty?
+        if cm.is_a?(RDoc::NormalModule)
+          @modules_hash.delete(cm.full_name)
+        else
+          @classes_hash.delete(cm.full_name)
+        end
+        cm.parent&.classes_hash&.delete(cm.name)
+        cm.parent&.modules_hash&.delete(cm.name)
+      end
+    end
+  end
+
+  ##
   # Make sure any references to C variable names are resolved to the corresponding class.
   #
 
--- a/test/rdoc/rdoc_store_test.rb
+++ b/test/rdoc/rdoc_store_test.rb
@@ -1255,4 +1255,53 @@ class RDocStoreTest < XrefTestCase
     assert_equal 'updated comment from a', klass.comment_location[file_a].first.to_s
   end
 
+  def test_cleanup_stale_comment_locations
+    file_a = @s.add_file 'a.rb'
+    file_b = @s.add_file 'b.rb'
+
+    klass = file_a.add_class RDoc::NormalClass, 'StaleClass'
+    klass.record_location file_a
+    klass.record_location file_b
+    file_a.add_to_classes_or_modules klass
+    file_b.add_to_classes_or_modules klass
+
+    klass.add_comment 'comment from a', file_a
+    klass.add_comment 'comment from b', file_b
+
+    # Simulate keep_position clearing followed by re-parse that no longer
+    # defines StaleClass in a.rb (no add_comment called for file_a)
+    @s.clear_file_contributions 'a.rb', keep_position: true
+
+    # Stale placeholder exists
+    assert_equal [], klass.comment_location[file_a]
+
+    # Cleanup should remove the stale entry
+    @s.cleanup_stale_contributions
+
+    assert_not_include klass.comment_location.keys, file_a
+    assert_not_include klass.in_files, file_a
+    assert_includes klass.comment_location.keys, file_b
+    assert_includes klass.in_files, file_b
+  end
+
+  def test_cleanup_stale_contributions_removes_empty_class
+    file_a = @s.add_file 'a.rb'
+
+    klass = file_a.add_class RDoc::NormalClass, 'GoneClass'
+    klass.record_location file_a
+    file_a.add_to_classes_or_modules klass
+
+    klass.add_comment 'comment from a', file_a
+
+    @s.clear_file_contributions 'a.rb', keep_position: true
+
+    # Stale placeholder exists, class still in store
+    assert_includes @s.classes_hash, 'GoneClass'
+
+    @s.cleanup_stale_contributions
+
+    # Class should be removed from store
+    assert_not_include @s.classes_hash, 'GoneClass'
+  end
+
 end


lib/rdoc/server.rb
```diff
@@ -0,0 +1,450 @@
...
+
+  ##
+  # Serves an HTML page, rendering from the generator or returning a cached
+  # version.
+
+  def serve_page(path)
+    name = path.sub(%r{\A/}, '')
+    name = 'index.html' if name.empty?
+
+    html = @mutex.synchronize { @page_cache[name] } || render_page(name)
```
kou: It seems that `@mutex.synchronize { @page_cache[name] }` is needless because `render_page` does it:

```suggestion
    html = render_page(name)
```

st0012: Ah good catch

lib/rdoc/server.rb
```diff
@@ -0,0 +1,450 @@
...
+  def build_search_index
+    @search_index_cache ||=
+      "var search_data = #{JSON.generate(index: @generator.build_search_index)};"
+  end
+
+  ##
+  # Injects the live-reload polling script before +</body>+.
+
+  def inject_live_reload(html)
+    html.sub('</body>', "#{LIVE_RELOAD_SCRIPT}</body>")
```
tompng: Just an idea, how about inserting `last_change_time` to this script?

There's a possibility to skip some change:
1. `last_change_time` updated to time1
2. Reload the page generated with time1
3. `last_change_time` updated again, value becomes time2
4. LIVE_RELOAD_SCRIPT loads initial `lastChange` value, get `time2` but the content is generated in `time1`

lib/rdoc/server.rb
```diff
@@ -0,0 +1,450 @@
...
+  # refreshes the generator, and invalidates caches.
+
+  def reparse_and_refresh(changed_files, removed_files)
+    @mutex.synchronize do
+      unless removed_files.empty?
+        $stderr.puts "Removed: #{removed_files.join(', ')}"
+        removed_files.each do |f|
+          @file_mtimes.delete(f)
+          relative = relative_path_for(f)
+          @store.remove_file(relative)
```
tompng, line 350: I think what we need to do for removed files and changed files has overlap. File changed to empty file and file deletion seems essentially the same.
- Remove: Remove contributions and remove file from store
- Update: Remove contributions and re-parse the file

How about calling `clear_file_contributions` here without removing all classes and modules?
In `clear_file_contributions`, we can remove class or module if it is only defined in this file.
```ruby
cm.in_files.delete(top_level)
if cm.in_files.empty?
  # Remove from store
  @store.send(:remove_classes_and_modules, [cm])
  # Remove from parent's classes_hash
  cm.parent.classes_hash.delete(cm.name) # or modules_hash.delete(cm.name)
 end
```
This way, the limitation for reopened classes and file deletion will be gone/improve.
File change that renames a class defined in it might also work as expected.

st0012: Good catch. Updated 👍

lib/rdoc/rdoc.rb
```diff
@@ -456,6 +465,19 @@ def document(options)
...
+      @store.load_cache
+
+      parse_files @options.files
+
+      @options.default_title = "RDoc Documentation"
+
+      @store.complete @options.visibility
+
+      start_server
+      return
```
kou: `exit` may be better because other code in this method uses `exit` not `return`.

lib/rdoc/server.rb
```diff
@@ -0,0 +1,393 @@
...
+    client.write(header)
+    client.write(body_bytes)
+    client.flush
+  end
+
+  ##
+  # Serves a static asset (CSS, JS) from the Aliki template directory.
+
+  def serve_asset(path)
+    rel_path = path.sub(%r{\A/}, '')
```
kou:
```suggestion
    rel_path = path.delete_prefix("/")
```

lib/rdoc/server.rb
```diff
@@ -0,0 +1,393 @@
...
+    content_type = CONTENT_TYPES[ext] || 'application/octet-stream'
+    [200, content_type, File.read(real_asset)]
+  end
+
+  ##
+  # Serves an HTML page, rendering from the generator or returning a cached
+  # version.
+
+  def serve_page(path)
+    name = path.sub(%r{\A/}, '')
```
kou:
```suggestion
    name = path.delete_prefix("/")
```

lib/rdoc/server.rb
```diff
@@ -0,0 +1,393 @@
...
+  # Returns a live-reload polling script with the given +last_change_time+
+  # embedded so the browser knows the exact timestamp of the content it
+  # received.  This avoids a race where a change that occurs between page
+  # generation and the first poll would be silently skipped.
+
+  def self.live_reload_script(last_change_time)
+    <<~JS
+      <script>
+      (function() {
+        var lastChange = #{last_change_time};
```
kou: In this case, `last_change_time` is always `Float`. So we can use `to_s` here but using `to_json` is better for embedding a Ruby object to JavaScript code.

```suggestion
        var lastChange = #{last_change_time.to_json};
```

lib/rdoc/ri/servlet.rb
```diff
@@ -24,14 +24,14 @@
...
+#   server.mount '/', RDoc::RI::Servlet
 #
 # If you want to mount the servlet some other place than the root, provide the
 # base path when mounting:
 #
-#   server.mount '/rdoc', RDoc::Servlet, '/rdoc'
+#   server.mount '/rdoc', RDoc::RI::Servlet, '/rdoc'
 
-class RDoc::Servlet < WEBrick::HTTPServlet::AbstractServlet
+class RDoc::RI::Servlet < WEBrick::HTTPServlet::AbstractServlet
```
kou, line 34: I don't have a strong opinion whether we re-implement our HTTP server or we use WEBrick but do we want to keep depending on WEBrick?

kou (approved):
+1

lib/rdoc/ri/servlet.rb
st0012: I think we can let it go and potentially merge the server implementation between RDoc and RI. But it'll be outside of the scope of the PR.

lib/rdoc/code_object/class_module.rb
```diff
@@ -125,7 +122,7 @@ def initialize(name, superclass = nil)
     @is_alias_for     = nil
     @name             = name
     @superclass       = superclass
-    @comment_location = [] # Array of [comment, location] pairs
+    @comment_location = {} # Hash of { location => comment }
```
tompng: Same location comment will be removed with this change.

Example input:
```ruby
# comment1
class A; end
# comment2
class A; end
```
Output:
```html
<section class="description">
<p>comment1</p> <!-- this line disappears -->
<p>comment2</p>
</section>
```

I don't think these kind of reopening with comment is normally used, but the original deduplicating logic is clearly only for C code.
```ruby
if location.parser == RDoc::Parser::C
  @comment_location.delete_if { |(_, l)| l == location }
end
```
Since reopening class with a comment is both included in the document in some case(different file), I think it's natural to accept the same thing written in a single file too.

lib/rdoc/store.rb
```diff
@@ -193,6 +193,92 @@ def add_file(absolute_name, relative_name: absolute_name, parser: nil)
...
+      # Remove includes, extends, and aliases from this file
+      cm.includes.reject! { |i| i.file == top_level }
+      cm.extends.reject! { |e| e.file == top_level }
+      cm.aliases.reject! { |a| a.file == top_level }
+      cm.external_aliases.reject! { |a| a.file == top_level }
+
+      # Clear or remove comment entries from this file
+      if cm.is_a?(RDoc::ClassModule)
+        if keep_position
+          cm.comment_location[top_level] = [] if cm.comment_location.key?(top_level)
```
tompng, line 251: A very edge case, `comment_location[top_level]` will be kept empty if reparsed file doesn't contain that ClassModule anymore, or a comment to the ClassModule is erased.
After this has happened, even every contribution files are removed, `if cm.in_files.empty?` won't be true.

I think one possible solution for this is:
```ruby
def reparse_and_refresh(...)
  unless removed_files.empty?
    ...
    # Don't need to perform `@modules_hash.delete(cm.full_name)` logic
  end

  unless changed_files.empty?
    ...
  end

  if removed_files.any? || changed_files.any?
    # Check all classes/modules for empty comment_location entries
    # and perform `@modules_hash.delete(cm.full_name)` logic
  end
  ...
end
```

lib/rdoc/code_object/class_module.rb
st0012: Addressed in #1634

lib/rdoc/store.rb
```diff
@@ -193,6 +193,121 @@ def add_file(absolute_name, relative_name: absolute_name, parser: nil)
...
+
+    top_level.classes_or_modules.each do |cm|
+      # Remove methods and attributes contributed by this file
+      cm.method_list.reject! { |m| m.file == top_level }
+      cm.attributes.reject! { |a| a.file == top_level }
+
+      # Rebuild methods_hash from remaining methods and attributes
+      cm.methods_hash.clear
+      cm.method_list.each { |m| cm.methods_hash[m.pretty_name] = m }
+      cm.attributes.each { |a| cm.methods_hash[a.pretty_name] = a }
```
tompng, line 235: When method is added/removed, method list in the left sidebar navigation is not updated.
Old entries seems to be cached in `@instance_methods` and `@class_method_list`. I think these instance variables also need to be cleared.

```diff
+ cm.instance_variable_set(:@instance_methods, nil)
+ cm.instance_variable_set(:@class_method_list, nil)
# or move rebuild-method logic to ClassModule's method
```

I'm curious if these two instance variable caches are really needed.
I haven't checked thoroughly though, it seems only used in sidebar and called only once per class/module.
Similar cached value `@instance_attributes` and `@class_attributes` seems unused.

st0012: Good catch. I'll clear them now here and check if we can drop them completely

PR #1621, opened 2026-02-21
Support constant assign parsed before class/module definition

tompng:
Support `Const = RHS` parsed before `class RHS;end` defined in another file.

This will fix one document-coverage check failure in https://github.com/ruby/ruby/pull/16194

### DIFF
--- a/lib/rdoc/code_object/constant.rb
+++ b/lib/rdoc/code_object/constant.rb
@@ -92,8 +92,16 @@ class RDoc::Constant < RDoc::CodeObject
       @is_alias_for = found if found
       @is_alias_for
     else
-      @is_alias_for
+      @is_alias_for ||= find_alias_for
     end
+  end
+
+  # Find alias constant for a value.
+  # This is used when the constant is parsed before the class or module defined in another file.
+  # Note that module nesting information is lost, so constant lookup is inaccurate.
+
+  def find_alias_for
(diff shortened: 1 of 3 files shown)


lib/rdoc/code_object/constant.rb
```diff
@@ -86,6 +86,8 @@ def full_name
   # The module or class this constant is an alias for
 
   def is_alias_for
+    @is_alias_for ||= find_alias_for
```
kou: May `find_alias_for` return a `String`? If `find_alias_for` must not return a `String`, how about doing this in `else`?

```diff
diff --git a/lib/rdoc/code_object/constant.rb b/lib/rdoc/code_object/constant.rb
index 8823b0bd..69c3133a 100644
--- a/lib/rdoc/code_object/constant.rb
+++ b/lib/rdoc/code_object/constant.rb
@@ -92,7 +92,7 @@ class RDoc::Constant < RDoc::CodeObject
       @is_alias_for = found if found
       @is_alias_for
     else
-      @is_alias_for
+      @is_alias_for ||= find_alias_for
     end
   end
 ```

kou (approved):
+1

lib/rdoc/code_object/constant.rb
tompng: That looks much better, thank you. Updated

PR #1622, opened 2026-02-22
Decouple Store internals from CodeObject classes

st0012:
Move resolution logic out of CodeObjects into Store:
- Add `Store#resolve_parent` to encapsulate lazy parent loading (previously `CodeObject#parent` called `load_class` for disk I/O)
- Add `Store#resolve_mixin` to encapsulate namespace-walking (previously 30 lines of scoping logic lived in `Mixin#module`)

Consolidate file normalization in base `CodeObject#store=`:
- Remove redundant `store=` overrides from `MethodAttr`, `AnyMethod`, `Mixin`, and `Constant` that all did identical `add_file` calls
- Fix pre-existing bug where `AnyMethod` called `add_file` twice

Replace direct hash reads with existing finder methods:
- Use `find_class_or_module`, `find_class_named`, `find_module_named` instead of `@store.classes_hash[]` / `@store.modules_hash[]`

Simplify `add_class_or_module` signature:
- Drop the `all_hash` parameter that passed Store's internal hash by reference; method now registers directly via the hash accessors

### DIFF
--- a/test/rdoc/parser/ruby_test.rb
+++ b/test/rdoc/parser/ruby_test.rb
@@ -2016,8 +2016,8 @@ end
 
     @parser.parse_method m, RDoc::Parser::Ruby::NORMAL, tk, @comment
 
-    assert_empty @store.modules_hash.keys
-    assert_equal %w[M], @store.classes_hash.keys
+    assert_equal %w[M], @store.modules_hash.keys
+    assert_empty @store.classes_hash.keys
   end
 
   def test_parse_method_false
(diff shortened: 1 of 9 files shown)


test/rdoc/parser/ruby_test.rb
```diff
@@ -2016,8 +2016,8 @@ def test_parse_method_constant
 
     @parser.parse_method m, RDoc::Parser::Ruby::NORMAL, tk, @comment
 
-    assert_empty @store.modules_hash.keys
-    assert_equal %w[M], @store.classes_hash.keys
```
st0012, line 2020: I think the test was wrong: if `M` is added as a `NormalModule`, it should appear in the modules hash, not classes hash.

PR #1623, opened 2026-02-22
Refactor formatter options

st0012:
1. The `markup` param is never actually used. So let's drop it
2. Instead of holding `Option` objects, formatters should just take the relevant option values. This will unlock further refactors to decouple `CodeObject` and `Store` later.

### DIFF
--- a/lib/rdoc/markup/to_tt_only.rb
+++ b/lib/rdoc/markup/to_tt_only.rb
@@ -16,13 +16,6 @@ class RDoc::Markup::ToTtOnly < RDoc::Markup::Formatter
   attr_reader :res
 
   ##
-  # Creates a new tt-only formatter.
-
-  def initialize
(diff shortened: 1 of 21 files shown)


lib/rdoc/markup/to_tt_only.rb
```diff
@@ -18,8 +18,8 @@ class RDoc::Markup::ToTtOnly < RDoc::Markup::Formatter
   ##
   # Creates a new tt-only formatter.
 
-  def initialize(markup = nil)
-    super nil, markup
+  def initialize
+    super
```
tompng: It only calls `super`, so maybe we don't need to override `def initialize`

st0012: Right, removed this and `ToJoinedParagraph#initialize`

PR #1624, opened 2026-02-23
Stop generating a separate page file for `main_page`

st0012:
When a page is used as `main_page` (e.g. `README.md`), its content is already displayed on index.html. Generating a separate page file (e.g. `README_md.html`) is redundant.

Skip the main_page in:
- darkfish generator (page file generation)
- json search index (avoids linking to a nonexistent page)
- aliki sidebar pages template

### DIFF
--- a/lib/rdoc/code_object/top_level.rb
+++ b/lib/rdoc/code_object/top_level.rb
@@ -221,9 +221,15 @@ class RDoc::TopLevel < RDoc::Context
   # Path to this file for use with HTML generator output.
 
   def path
+    base = if options.main_page == full_name
+             'index.html'
+           else
+             http_url
+           end
+
     prefix = options.file_path_prefix
-    return http_url unless prefix
-    File.join(prefix, http_url)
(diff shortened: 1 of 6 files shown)


tompng:
~Looks like separate page file is still generated. `README_md.html` is generated~ and linked from [main page](https://50c952f8.rdoc-6cd.pages.dev/#:~:text=Please%20follow%20the%20theme%E2%80%99s%20README%20for%20usage%20instructions.) by crossref resolve.
```html
<p>Please follow the theme’s <a href="README_md.html">README</a> for usage instructions.</p>
```

~However, I don't think this is a problem.~ If we don't generate the file, we need to also fix crossref part to avoid linking to a file which is not generated.

tompng:
Sorry, https://50c952f8.rdoc-6cd.pages.dev/README_md.html looks like a cached file, it's not generated.
Text `"README"` in a document is linked to a file which is not generated.

st0012:
@tompng Thanks, fixed

PR #1626, opened 2026-02-26
Fix markdown table parser consuming lines without pipes as table rows

st0012:
The TableRow rule in the PEG grammar allowed rows with zero pipe characters (`TableItem2*`). This caused lines like `<br>` immediately following a table to be parsed as single-cell table rows, producing spurious `<td><br></td>` in the rendered HTML.

Change `TableItem2*` to `TableItem2+` so that rows not starting with `|` must contain at least one `|` to be recognized as table rows.

### DIFF
--- a/test/rdoc/rdoc_markdown_test.rb
+++ b/test/rdoc/rdoc_markdown_test.rb
@@ -1334,6 +1334,19 @@ and an extra note.[^2]
     assert_equal expected, doc
   end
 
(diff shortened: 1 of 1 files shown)


lib/rdoc/markdown.kpeg
kou: Should we change this too?

st0012: Done in 60820a6b. Changed `TableAlign2*` to `TableAlign2+` for consistency.

PR #1550, opened 2026-01-06
Add comparison with GitHub Flavored Markdown spec

st0012:
I compiled the list with help from Claude. I tested some common elements manually but not all of them. That said, I think this would still be a good comparison reference to have.

PR #1627, opened 2026-02-28
Fix Markdown blockquote parsing

st0012:
## Summary

Two fixes to the `BlockQuoteRaw` rule in the PEG grammar:

- Separate blockquotes (separated by an unquoted blank line) were merged into one. Now only blank lines prefixed with `>` continue the blockquote, so an unquoted blank line ends it.
- Lazy continuation was consuming block-level elements (headings, list markers, code fences) that should break out of the blockquote. Added negative lookaheads for these elements.

Ref: https://github.com/ruby/rdoc/pull/1550#discussion_r2665504990

### DIFF
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -617,8 +617,8 @@ BlockQuote = BlockQuoteRaw:a
 
 BlockQuoteRaw =  @StartList:a
                  (( ">" " "? Line:l { a << l } )
-                  ( !">" !@BlankLine Line:c { a << c } )*
-                  ( @BlankLine:n { a << n } )*
+                  ( !">" !@BlankLine !(AtxStart @Spacechar) !Bullet !Enumerator !( &{ github? } Ticks3 ) Line:c { a << c } )*
+                  ( ">" @BlankLine:n { a << n } )*
                  )+
                  { inner_parse a.join }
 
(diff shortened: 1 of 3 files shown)


lib/rdoc/markdown.kpeg
```diff
@@ -617,8 +617,8 @@ BlockQuote = BlockQuoteRaw:a
 
 BlockQuoteRaw =  @StartList:a
                  (( ">" " "? Line:l { a << l } )
-                  ( !">" !@BlankLine Line:c { a << c } )*
-                  ( @BlankLine:n { a << n } )*
+                  ( !">" !@BlankLine !AtxStart !Bullet !Enumerator !Ticks3 Line:c { a << c } )*
```
kou: Does this work with the following cases?

```markdown
> x
#no-space-heading
```

> x
#no-space-heading

```markdown
> x
|header|
|--|
|content|
```

> x
|header|
|--|
|content|

st0012: 1. Kinda yes. It's displayed that way but `#` is stripped. It's a bug I'll address separately in #1676 
2. The table is included on the same time as `x` in the quote block. I tried to fix it but it's a much more complex fix so I think it deserves a separate PR.

kou: It makes sense.

We may need to add more `!XXX` eventually. (`XXX` will be came from https://github.com/ruby/rdoc/blob/799e551a94c5591861c62d606deb85b8ca722bcf/lib/rdoc/markdown.kpeg#L566-L581 .)

PR #1628, opened 2026-03-01
Enable break_on_newline extension by default for Markdown

st0012:
## Summary

- Enable the existing `break_on_newline` extension in `DEFAULT_EXTENSIONS` so the Markdown parser converts soft line breaks to `HardBreak` objects at parse time
- This produces visible `<br>` line breaks in HTML output, matching GFM rendering where `> Foo\nbar` shows "Foo" and "bar" on separate lines
- The conversion happens in the Markdown parser's `paragraph()` method, which is the proper place for Markdown-specific behavior — no changes to the generic `accept_paragraph` in `ToHtml`
- Also removes the CJK-aware newline-to-space gsub from `accept_paragraph` that is no longer needed (the RDoc markup parser already handles newline joining at parse time in `build_paragraph`)

Ref: https://github.com/ruby/rdoc/pull/1550#discussion_r2665504990

### DIFF
--- a/test/rdoc/markup/to_html_test.rb
+++ b/test/rdoc/markup/to_html_test.rb
@@ -236,7 +236,7 @@ class RDocMarkupToHtmlTest < RDoc::Markup::FormatterTestCase
   end
 
   def accept_paragraph_break
-    assert_equal "\n<p>hello<br> world</p>\n", @to.res.join
+    assert_equal "\n<p>hello<br>\nworld</p>\n", @to.res.join
   end
 
   def accept_paragraph_i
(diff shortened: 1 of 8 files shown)


test/rdoc/markup/to_html_test.rb
```diff
@@ -411,43 +411,13 @@ def test_accept_heading_dedup_resets_on_start_accepting
...
-    assert_equal "\n<p><code>hello</code> world</p>\n", @to.res.join
-
-    @to.start_accepting
-    @to.accept_paragraph para("hello\n", "+world+\n")
-    assert_equal "\n<p>hello <code>world</code></p>\n", @to.res.join
-
-    @to.start_accepting
-    @to.accept_paragraph para("+hello+\n", "+world+\n")
-    assert_equal "\n<p><code>hello</code> <code>world</code></p>\n", @to.res.join
+    assert_equal "\n<p><code>hello</code>\nworld\n</p>\n", @to.res.join
```
kou, line 420: Is this intentional?
Should `world` be surrounded with `<code>`?

```suggestion
    assert_equal "\n<p><code>hello</code>\n<code>world</code>\n</p>\n", @to.res.join
```

PR #1647, opened 2026-03-15
Fix server mode live reload for C files

st0012:
## Summary

- The C parser didn't call `add_to_classes_or_modules` when creating classes/modules, so `clear_file_contributions` couldn't clean up old entries during server mode re-parse — causing duplicate methods when editing C files with live reload.
- Add the call in both `find_class` and `handle_class_module`, matching what the Ruby parsers already do.


st0012:
Ah...I was trying to merge #1648 but clicked the merge button in the wrong tab 🤦‍♂️ 
Please still leave comments and I'll address them in separate PRs.

PR #1648, opened 2026-03-15
Print timing for page requests and re-parsing in server mode

st0012:
## Summary

- Silence the stats progress bar in server mode — it uses `\r` on a tty which collides with the server's own logging
- Print response time for page requests (not assets or status polls): `200 /RDoc.html (12.3ms)`
- Print re-parse duration with relative paths: `Re-parsed lib/rdoc.rb (17.5ms)`

```
Serving documentation at: http://localhost:4000
Press Ctrl+C to stop.

200 /RDoc.html (14.9ms)
Re-parsed lib/rdoc.rb (15.7ms)
200 /RDoc.html (8.5ms)
```

### DIFF
--- a/lib/rdoc/server.rb
+++ b/lib/rdoc/server.rb
@@ -200,6 +200,8 @@ class RDoc::Server
     client.write(header)
     client.write(body_bytes)
     client.flush
+  rescue Errno::EPIPE
+    # Client disconnected before we finished writing — harmless.
   end
 
   ##


lib/rdoc/server.rb
```diff
@@ -67,6 +67,8 @@ def initialize(rdoc, port)
     @store = rdoc.store
     @port = port
 
+    # Silence stats output — the server prints its own timing.
+    @rdoc.instance_variable_set(:@stats, RDoc::Stats.new(@store, 0, 0))
```
kou: Can we avoid using `instance_variable_set`?
For example, how about adding `RDoc::Stats#verbosity=` to change `@display` in `RDoc::Stats`?

lib/rdoc/server.rb
```diff
@@ -143,7 +151,14 @@ def handle_client(client)
       return write_response(client, 405, 'text/plain', 'Method Not Allowed')
     end
 
-    status, content_type, body = route(path)
+    if path.start_with?('/__') || path.match?(%r{\A/(?:css|js)/})
```
kou: In general, `Regexp#match?` is better than `String#match?` because Ruby is a pure object-oriented language. In this case, a regular expression is the main actor.

```suggestion
    if path.start_with?('/__') || %r{\A/(?:css|js)/}.match?(path)
```

kou (approved):
+1

PR #1649, opened 2026-03-16
Fix deadlock on Ctrl+C in server mode

st0012:
## Summary

- Pressing Ctrl+C in `rdoc --server` raised `ThreadError: deadlock; recursive locking` because the `INT` trap called `shutdown`, which closed the TCP server socket from the same thread blocked on `accept`
- Replace trap-based shutdown with `rescue Interrupt` around the accept loop, and move cleanup into an `ensure` block

### DIFF
--- a/lib/rdoc/server.rb
+++ b/lib/rdoc/server.rb
@@ -91,15 +91,10 @@ class RDoc::Server
     loop do
       client = @tcp_server.accept
       Thread.new(client) { |c| handle_client(c) }
-    rescue IOError
-      break
     end
-  end
-
-  ##
-  # Shuts down the server.
-
-  def shutdown
+  rescue Interrupt
+    # Ctrl+C
+  ensure
     @running = false
     @tcp_server&.close
     @watcher_thread&.join(2)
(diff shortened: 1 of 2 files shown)


lib/rdoc/server.rb
```diff
@@ -89,20 +89,26 @@ def start
     $stderr.puts "Press Ctrl+C to stop.\n\n"
 
     loop do
+      break unless @running
+
+      readable, = IO.select([@tcp_server], nil, nil, 1)
+      next unless readable
```
kou: It may be a overkill but we can remove the `1` timeout by adding a pipe that notifies whether shutdown is requested or not.

```ruby
IO.pipe do |shutdown_notify_input, shutdown_notify_output|
  running = true
  trap("INT") do
    running = false
    shutdown_notify_output.puts("done")
    shutdown_notify_output.flush
  end
  while running
    readables, = IO.select([$stdin, shutdown_notify_input])
    break if readables.include?(shutdown_notify_input)
    puts($stdin.gets)
  end
end
```

Or we can just use `rescue Interrupt` instead of `trap("INT")`:

```ruby
IO.pipe do |input, output|
  begin
    loop do
      puts($stdin.gets)
    end
  rescue Interrupt
    # cleanup
  end
end
```

st0012: Thanks, updated to the 2nd approach

Issue #924, opened 2022-09-05
Update licenses listed in gem


voxik:
On top of #921, I'd like to ask if the .gemspec license field can be updated. Currently, it says just `Ruby`:

https://github.com/ruby/rdoc/blob/d12272b4cf65d8a0004233a047065af093d2345e/rdoc.gemspec#L29

But I think that the field should list all licensed used in the package and therefore it should be:

`GPL-2.0 AND Ruby AND BSD-3-Clause AND CC-BY-2.5 and OFL-1.1-RFN`

Where the 
* `GPL-2.0 AND Ruby` comes from the `LICENSE.rdoc` file,
* `BSD-3-Clause` is specified in `lib/rdoc/generator/darkfish.rb`
* `CC-BY-2.5` should be the silk icon set license
* `OFL-1.1-RFN` is from `lib/rdoc/generator/template/darkfish/css/fonts.css`

The `LEGAL.rdoc` also mentions some test file which follows Perl licensing, however this code is not shipped in resulting packages, so it probably does not qualifies.

BTW the `GPL-2.0` is also questionable, because Ruby itself was relicensed quite some time ago, but the RDoc LICENSE file still states `GPL-2.0`, so that is probably fine as it is. Not sure.

voxik:
`GPL-2.0` is deprecated identifier. Therefore `GPL-2.0-only` should probably be used. But hard to tell.

Issue #401, opened 2016-04-12
license isn't same as ruby license


lksv:
It seems that this gem has the same license as ruby (https://github.com/rdoc/rdoc/blob/master/Rakefile#L44) but it is not true:

snippet of rdoc license:
> You can redistribute it and/or modify it under either the terms of the GPL version 2 (see the file GPL), or the conditions below:

ruby license:
> You can redistribute it and/or modify it under either the terms of the 2-clause BSDL (see the file BSDL), or the conditions below:

I think this is an mistake. Ruby changed the license in time (see https://github.com/ruby/ruby/commit/2cd6800fd8437b1f862f3f5c44db877159271d17#diff-7116ef0705885343c9e1b2171a06be0e)

My intention is not to push to change the license if you do not want.

zzak:
It's not really my call, but my guess is we just copied over the old Ruby license and when it changed, we never updated RDoc's.

/cc @drbrain

drbrain:
Ruby license means different things for different versions of ruby.

We'd need to get Dave Thomas and other contributors consent to change the license to the latest version of the Ruby license.

PR #1654, opened 2026-03-19
chore: Have license content match gemspec

thompson-tomo:
This updates the gemspec to explicitly mention the GPL license as per the license files which also make clear the license expression.

This ensures that tools such as Fossa are not generating false positives for using just GPL licenses in Apache licensed projects.

### DIFF
--- a/rdoc.gemspec
+++ b/rdoc.gemspec
@@ -26,7 +26,7 @@ RDoc produces HTML and command-line documentation for Ruby projects.
 RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentation from the command-line.
   DESCRIPTION
   s.homepage = "https://ruby.github.io/rdoc"
-  s.licenses = ["Ruby", "GPL-2.0-onlyGPL-2.0-only"]
+  s.licenses = ["Ruby", "GPL-2.0-only"]
 
   s.metadata["homepage_uri"] = s.homepage
   s.metadata["source_code_uri"] = "https://github.com/ruby/rdoc"


kou:
The (original) Ruby license is a dual-license of the Ruby's original license (that is described in LICENSE.rdoc) and GPL.

FYI: The current Ruby license is a dual-license of the Ruby's original license and BSD-2-Clause.

If we remove the GPL part, it means that we change our license. (Single-license not dual-license)

thompson-tomo:
I understand however the gemspec describes it as having a single licence `ruby` https://github.com/ruby/rdoc/blob/master/rdoc.gemspec which differs to the file hence creates a disconnect.

kou:
`Ruby` means dual-license not single-license.

thompson-tomo:
As per https://spdx.org/licenses/Ruby.html, the licence identifier ruby is a single licence type with the 6 conditions outlined which is different to the ruby project which is explained as:

> Ruby is disjunctively licensed project that allows the choice of this license and another.

Further based on https://guides.rubygems.org/specification-reference/#license= the gemspec is currently pointing to just those 6 conditions.

What I have now done is updated the gemspec to contain both licences and made the license file explicit in its license identifier.

LICENSE.rdoc
```diff
@@ -1,3 +1,5 @@
+# SPDX-License-Identifier: Ruby or GPL-2
```
kou: * `#` isn't a comment markup in RDoc markup: https://ruby.github.io/rdoc/doc/markup_reference/rdoc_rdoc.html
  * If we don't output it, we need to surround it with `:stopdoc:` and `:startdoc:` or `--` and `++`: https://ruby.github.io/rdoc/doc/markup_reference/rdoc_rdoc.html#directives-for-allowing-or-suppressing-documentation
* `GPL-2` isn't a valid ID. We should use `GPL-2.0-or-later`: https://spdx.org/licenses/GPL-2.0-or-later.html

rdoc.gemspec
```diff
@@ -26,7 +26,7 @@ RDoc produces HTML and command-line documentation for Ruby projects.
 RDoc includes the +rdoc+ and +ri+ tools for generating and displaying documentation from the command-line.
   DESCRIPTION
   s.homepage = "https://ruby.github.io/rdoc"
-  s.licenses = ["Ruby"]
+  s.licenses = ["Ruby", "GPL"]
```
kou: `GPL` isn't a valid ID. We should use `GPL-2.0-or-later`.

thompson-tomo:
Thanks for spoting the issue with the spdx-id.

I did not ignore the notes part, in fact it supports the change here and is why I included the key extract. This is because the note hightlights that the ruby license is not the same as the license used by the Ruby project. This is because the project allows a choice between the ruby license and either gpl/BSD.

rdoc.gemspec
thompson-tomo: Should be fixed.

LICENSE.rdoc
thompson-tomo: Id has been corrected. Is the comment syntax now correct? Note the need is to satisfy https://spdx.dev/learn/handling-license-info/

LICENSE.rdoc
```diff
@@ -1,3 +1,7 @@
+:stopdoc:
+# SPDX-License-Identifier: Ruby or GPL-2.0-or-later
```
kou: We don't need `#` here:

```suggestion
SPDX-License-Identifier: Ruby or GPL-2.0-or-later
```

kou:
Oh, sorry. I misunderstood that you cited other part. I don't think that it means that `Ruby` SPDX ID is single-license but I'm not good at English.

LICENSE.rdoc
thompson-tomo: Based on https://spdx.dev/learn/handling-license-info/ we do need it especially as the rdoc file is Bundled into the gem.

thompson-tomo:
No worries, key thing is the content under the text heading is classified as the license.

LICENSE.rdoc
kou: You're referring this part, right?

> It consists of the following parts on one line:
>
> * [if required] The ‘begin comment’ characters for the applicable programming language (/*, //, #, …)
> * The characters SPDX-License-Identifier: with whitespace following the colon
> * An SPDX license ID (e.g., Apache-2.0) or SPDX license expression (see below)
> * [if required] The ‘end comment’ characters for the applicable programming language (*/, …)

`#` isn't the "begin comment characters" in RDoc markup. Do we really need `#` here?

thompson-tomo: Hmm, let's go without it and see.

rhenium:
> I don't object that we also mention `GPL-2.0-or-later`

Do you mean `GPL-2.0-only`?

https://github.com/ruby/ruby/blob/2cd6800fd8437b1f862f3f5c44db877159271d17~/COPYING

rhenium:
Related issues: https://github.com/ruby/rdoc/issues/401 https://github.com/ruby/rdoc/issues/924

I think an `SPDX-License-Identifier` comment is typically used to declare the license of the file it is attached to, so attaching it to `LICENSE.rdoc` feels a bit unusual.

kou:
https://redmine.ruby-lang.org/issues/2032

Oh, I misunderstood that Ruby license uses `GPL-2.0-or-later` not `GPL-2.0-only`. We should use `GPL-2.0-only` not `GPL-2.0-or-later`.

thompson-tomo:
Agree it feels counter-intuitive to also specify the expression but this way we can be sure that the 2 licence types are being combined into an expression as expected when tooling is scanning the gem.

#924 is similar however alot broader whereas this is focused on licence and gemspec being sync.

I have updated the expression to now use GPL-2.0-only.

PR #1665, opened 2026-03-30
Display RBS type signatures in documentation

st0012:
## Summary

- Display RBS type signatures in HTML documentation and RI terminal output
- Source type information from inline `#:` annotations and `.rbs` files (including RBS stdlib)
- Link type names in signatures to their documentation pages

type-signatures

## Details

**New module: `RDoc::RbsHelper`**
- Loads RBS signatures from `sig/` directories and stdlib declarations via `RBS::EnvironmentLoader`
- Validates inline `#:` annotations using `RBS::Parser`, warns with file:line on invalid signatures
- Renders type signatures as HTML with type names linked to their documentation pages using `a.rbs-type`

**Parser integration (`PrismRuby`)**
- Extracts `#:` annotation lines from comments via `RBS_SIG_LINE` constant
- Attaches parsed type signatures to `MethodAttr#type_signature_lines`
- Inline annotations take priority over `.rbs` file signatures

**Store**
- `merge_rbs_signatures` fills in type signatures from loaded `.rbs` files where no inline annotation exists
- `type_name_lookup` builds a cached name-to-path map for type linking; ambiguous unqualified names are excluded to avoid wrong links

**Display**
- Aliki theme: method signatures render as styled `<pre>` blocks with dotted separator; attribute signatures render inline after the `[RW]` badge
- RI driver: type signatures display as verbatim blocks after argument lists
- JS updated to exclude type signature blocks from copy-button wrapping

**Serialization**
- `MARSHAL_VERSION` bumped to 4 for both `AnyMethod` and `Attr` (additive — appends `type_signature_lines` to marshal array)

**CI/Dependencies**
- Minimum Ruby bumped to 3.2 (required by `rbs` 4.0)
- Added `rbs >= 4.0.0` and `prism >= 1.6.0` dependencies
- Dropped JRuby from CI matrix (`rbs` has a C extension); TruffleRuby retained (verified working)

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -461,16 +461,15 @@ class RDoc::Parser::Ruby < RDoc::Parser
   def consecutive_comment(line_no)
     return unless @unprocessed_comments.first&.first == line_no
     _line_no, start_line, text = @unprocessed_comments.shift
-    type_signature_lines = extract_type_signature!(text, start_line)
-    result = parse_comment_text_to_directives(text, start_line)
-    return unless result
-    comment, directives = result
-    [comment, directives, type_signature_lines]
+    parse_comment_text_to_directives(text, start_line)
   end
 
-  # Parses comment text and returns a pair of RDoc::Comment and directives
+  # Parses comment text and returns +[RDoc::Comment, directives, type_signature_lines]+,
+  # or +nil+ if the comment is a section header (which has no associated code
+  # object).
 
   def parse_comment_text_to_directives(comment_text, start_line) # :nodoc:
+    type_signature_lines = extract_type_signature!(comment_text, start_line)
     comment_text, directives = @preprocess.parse_comment(comment_text, start_line, :ruby)
     comment = RDoc::Comment.new(comment_text, @top_level, :ruby)
     comment.normalized = true
@@ -484,7 +483,7 @@ class RDoc::Parser::Ruby < RDoc::Parser
       return
     end
     @preprocess.run_post_processes(comment, @container)
-    [comment, directives]
+    [comment, directives, type_signature_lines]
   end
 
   # Extracts the comment for this section from the normalized comment block.
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -828,7 +828,8 @@ class RDoc::Parser::Ruby < RDoc::Parser
 
   # Extracts RBS type signature lines (#: ...) from raw comment text.
   # Mutates the input text to remove the extracted lines.
-  # Returns the type signature string, or nil if none found.
+  # Returns an array of extracted type signature lines, or nil if none are
+  # found. The array may contain multiple lines for overloaded signatures.
 
   def extract_type_signature!(text, start_line)
     return nil unless text.include?('#:')
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'prism'
+require_relative '../rbs_helper'
 
 # Parse and collect document from Ruby source code.
 
@@ -126,6 +127,9 @@ require 'prism'
 class RDoc::Parser::Ruby < RDoc::Parser
 
   parse_files_matching(/\.rbw?$/)
+
+  # Matches an RBS inline type annotation line: #: followed by whitespace
+  RBS_SIG_LINE = /\A#:\s/ # :nodoc:
 
   attr_accessor :visibility
   attr_reader :container, :singleton, :in_proc_block
@@ -457,10 +461,14 @@ class RDoc::Parser::Ruby < RDoc::Parser
   def consecutive_comment(line_no)
     return unless @unprocessed_comments.first&.first == line_no
     _line_no, start_line, text = @unprocessed_comments.shift
-    parse_comment_text_to_directives(text, start_line)
+    type_signature_lines = extract_type_signature!(text, start_line)
+    result = parse_comment_text_to_directives(text, start_line)
+    return unless result
+    comment, directives = result
+    [comment, directives, type_signature_lines]
   end
 
-  # Parses comment text and retuns a pair of RDoc::Comment and directives
+  # Parses comment text and returns a pair of RDoc::Comment and directives
 
   def parse_comment_text_to_directives(comment_text, start_line) # :nodoc:
     comment_text, directives = @preprocess.parse_comment(comment_text, start_line, :ruby)
@@ -569,7 +577,7 @@ class RDoc::Parser::Ruby < RDoc::Parser
   # Handles `attr :a, :b`, `attr_reader :a, :b`, `attr_writer :a, :b` and `attr_accessor :a, :b`
 
   def add_attributes(names, rw, line_no)
-    comment, directives = consecutive_comment(line_no)
+    comment, directives, type_signature_lines = consecutive_comment(line_no)
     handle_code_object_directives(@container, directives) if directives
     return unless @container.document_children
 
@@ -577,6 +585,7 @@ class RDoc::Parser::Ruby < RDoc::Parser
       a = RDoc::Attr.new(nil, symbol.to_s, rw, comment, singleton: @singleton)
       a.store = @store
       a.line = line_no
+      a.type_signature_lines = type_signature_lines
       record_location(a)
       handle_modifier_directive(a, line_no)
       @container.add_attribute(a) if should_document?(a)
@@ -615,7 +624,7 @@ class RDoc::Parser::Ruby < RDoc::Parser
 
   def add_method(method_name, receiver_name:, receiver_fallback_type:, visibility:, singleton:, params:, calls_super:, block_params:, tokens:, start_line:, args_end_line:, end_line:)
     receiver = receiver_name ? find_or_create_module_path(receiver_name, receiver_fallback_type) : @container
-    comment, directives = consecutive_comment(start_line)
+    comment, directives, type_signature_lines = consecutive_comment(start_line)
     handle_code_object_directives(@container, directives) if directives
 
     internal_add_method(
@@ -630,11 +639,12 @@ class RDoc::Parser::Ruby < RDoc::Parser
       params: params,
       calls_super: calls_super,
       block_params: block_params,
-      tokens: tokens
+      tokens: tokens,
+      type_signature_lines: type_signature_lines
     )
   end
 
-  private def internal_add_method(method_name, container, comment:, dont_rename_initialize: false, directives:, modifier_comment_lines: nil, line_no:, visibility:, singleton:, params:, calls_super:, block_params:, tokens:) # :nodoc:
+  private def internal_add_method(method_name, container, comment:, dont_rename_initialize: false, directives:, modifier_comment_lines: nil, line_no:, visibility:, singleton:, params:, calls_super:, block_params:, tokens:, type_signature_lines: nil) # :nodoc:
     meth = RDoc::AnyMethod.new(nil, method_name, singleton: singleton)
     meth.comment = comment
     handle_code_object_directives(meth, directives) if directives
@@ -655,6 +665,7 @@ class RDoc::Parser::Ruby < RDoc::Parser
     meth.params ||= params || '()'
     meth.calls_super = calls_super
     meth.block_params ||= block_params if block_params
+    meth.type_signature_lines = type_signature_lines
     record_location(meth)
     meth.start_collecting_tokens(:ruby)
     tokens.each do |token|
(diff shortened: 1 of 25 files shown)


skatkov:
Do I understand correctly that the sorbet type signature will not work with this implementation?

st0012:
Yes, because it's not a official signature syntax for Ruby.

kou:
> Dropped JRuby/TruffleRuby from CI matrix (`rbs` has a C extension)

Does this mean that JRuby can't use RDoc with this?

st0012:
@kou Unfortunately, yes. It's essentially a decision between:
1. Having `rbs` as an official dependency with proper version restriction in gemspec, but JRuby won't be supported until it supports `rbs`
1. Having `rbs` as a soft dependency and bail when `rbs` version doesn't match.
    - Note that with this approach, if a project runs RDoc with `bundle exec`, `rbs` needs to be added in the bundle manually/by other gems too

Given that I want to make RBS support a key feature of future RDoc and directly encourage gems to write more RBS signatures, I think 1 is a more reasonable decision.

st0012:
@headius I assumed JRuby doesn't support `rbs` 4.0+ due to [this failure](https://github.com/ruby/rdoc/actions/runs/23756502189/job/69212536022#step:3:300). Is it true that it's not supported? Or I can actually just tweak the CI config to make it work.

kou:
@soutaro Can RBS add support for JRuby?

soutaro:
@kou I'm not sure. I remember @headius was working on FFI with RBS (https://github.com/ruby/rbs/pull/2572). I thought it was something related to JRuby, but I don't have exact context about this.

kou:
Thanks. Let's wait for a response from @headius .

st0012:
@kou At the same time, I'd appreciate review on the rest of implementation too 🙏

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -836,6 +846,35 @@ def add_module_or_class(module_name, start_line, end_line, is_class: false, supe
...
+
+  def extract_type_signature!(text, start_line)
+    return nil unless text.include?('#:')
+
+    lines = text.lines
+    sig_lines, doc_lines = lines.partition { |l| l.match?(RBS_SIG_LINE) }
+    return nil if sig_lines.empty?
+
+    text.replace(doc_lines.join)
+    type_sig = sig_lines.map { |l| l.sub(RBS_SIG_LINE, '').chomp }.join("\n")
```
kou: It seems that we can simplify this:

```suggestion
    type_sig = sig_lines.map { |l| l.sub(RBS_SIG_LINE, '') }.join
```

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -836,6 +846,35 @@ def add_module_or_class(module_name, start_line, end_line, is_class: false, supe
...
+    validate_type_signature(type_sig, start_line + doc_lines.size)
+    type_sig
+  end
+
+  def validate_type_signature(sig, line_no)
+    sig.split("\n").each_with_index do |line, i|
```
kou:
```suggestion
    sig.each_line(chomp: true).with_index do |line, i|
```

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -836,6 +846,35 @@ def add_module_or_class(module_name, start_line, end_line, is_class: false, supe
...
+      method_error = RDoc::RbsHelper.validate_method_type(line)
+      next unless method_error
+      type_error = RDoc::RbsHelper.validate_type(line)
+      next unless type_error
```
kou: If we don't use error messages in warning messages, `valid_XXX?` instead of `validate_XXX` may be better.

lib/rdoc/rbs_helper.rb
```diff
@@ -0,0 +1,173 @@
...
+        signatures = {}
+
+        env.class_decls.each do |type_name, entry|
+          class_name = type_name.to_s.delete_prefix('::')
+
+          entry.each_decl do |decl|
+            decl.members.each do |member|
+              case member
+              when RBS::AST::Members::MethodDefinition
+                key = member.singleton? ? "#{class_name}::#{member.name}" : "#{class_name}##{member.name}"
```
kou: `#{class_name}.#{member.name}` is better for singleton method.


```suggestion
                key = member.singleton? ? "#{class_name}.#{member.name}" : "#{class_name}##{member.name}"
```

lib/rdoc/rbs_helper.rb
```diff
@@ -0,0 +1,173 @@
...
+                sigs = member.overloads.map { |o| o.method_type.to_s }
+                signatures[key] = sigs.join("\n")
+              when RBS::AST::Members::AttrReader, RBS::AST::Members::AttrWriter, RBS::AST::Members::AttrAccessor
+                key = "#{class_name}.#{member.name}"
```
kou: Hmm. Can we use `#` instead of `.` as the separator because attribute accessors just methods?

test/rdoc/code_object/any_method_test.rb
```diff
@@ -242,6 +242,27 @@ def test_marshal_dump
     assert_equal section,        loaded.section
   end
 
+  def test_marshal_dump_with_type_signature
+    @store.path = Dir.tmpdir
+    top_level = @store.add_file 'file.rb'
+
+    m = RDoc::AnyMethod.new nil, 'method'
+    m.block_params   = 'some_block'
```
kou:
```suggestion
    m.block_params    = 'some_block'
```

test/rdoc/parser/prism_ruby_test.rb
```diff
@@ -2374,6 +2374,160 @@ def m2; end
...
+    util_parser <<~RUBY
+      class Foo
+        #: (Integer) -> void
+        def bar(x); end
+      end
+    RUBY
+
+    klass = @store.find_class_named 'Foo'
+    bar = klass.method_list.first
+    assert_equal '(Integer) -> void', bar.type_signature
```
kou, line 2459: Should we assert `bar.comment.text` here? (`nil`? An empty string?)

test/rdoc/parser/prism_ruby_test.rb
```diff
@@ -2374,6 +2374,160 @@ def m2; end
...
+        #
+        #: (String) -> void
+        def bar(x); end
+      end
+    RUBY
+
+    klass = @store.find_class_named 'Foo'
+    bar = klass.method_list.first
+    assert_equal '(String) -> void', bar.type_signature
+    assert_includes bar.comment.text, 'Documentation here'
```
kou: Can we use `assert_equal` not `assert_includes` to check type signature doesn't exist in `bar.comment`?

```suggestion
    assert_equal "Documentation here\n", bar.comment.text
```

lib/rdoc/parser/prism_ruby.rb
kou: I'm not familiar with RBS syntax but RBS doesn't support splitting one signature to multiple lines, right?

Earlopain:
What's the issue with truffleruby? It can run native extensions just fine and tests pass.

st0012:
@Earlopain In the beginning it was crashing but now it seems fine. I'll add it back, thanks for raising this!

lib/rdoc/parser/prism_ruby.rb
st0012: No it doesn't. I was thinking Sorbet's specific support for it, but native RBS doesn't support it.

lib/rdoc/rdoc.rb
```diff
@@ -540,6 +544,20 @@ def generate
...
+
+  def load_rbs_signatures
+    sig_dirs = []
+    sig_dir = File.join(@options.root.to_s, 'sig')
+    sig_dirs << sig_dir if File.directory?(sig_dir)
+    signatures = RDoc::RbsHelper.load_signatures(*sig_dirs)
+    @store.merge_rbs_signatures(signatures) unless signatures.empty?
+  rescue RBS::ParsingError, Errno::ENOENT, LoadError => e
+    @options.warn "Failed to load RBS type signatures: #{e.message}"
+  end
```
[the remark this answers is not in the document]
st0012: This is a non-issue because if rbs is unavailable even with gemspec declaration, it's outside of RDoc's control.

lib/rdoc/rbs_helper.rb
```diff
@@ -0,0 +1,171 @@
...
+            decl.members.each do |member|
+              case member
+              when RBS::AST::Members::MethodDefinition
+                key = member.singleton? ? "#{class_name}.#{member.name}" : "#{class_name}##{member.name}"
+                sigs = member.overloads.map { |o| o.method_type.to_s }
+                signatures[key] ||= sigs.join("\n")
+              when RBS::AST::Members::AttrReader, RBS::AST::Members::AttrWriter, RBS::AST::Members::AttrAccessor
+                key = "#{class_name}##{member.name}"
+                signatures[key] ||= member.type.to_s
+              end
```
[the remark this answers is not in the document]
st0012: Keeping the first method's signature matches how RDoc handles duplicate method definitions.

lib/rdoc/parser/prism_ruby.rb
```diff
@@ -461,10 +465,14 @@ def skip_comments_until(line_no_until)
   def consecutive_comment(line_no)
     return unless @unprocessed_comments.first&.first == line_no
     _line_no, start_line, text = @unprocessed_comments.shift
-    parse_comment_text_to_directives(text, start_line)
+    type_signature = extract_type_signature!(text, start_line)
+    result = parse_comment_text_to_directives(text, start_line)
+    return unless result
+    comment, directives = result
+    [comment, directives, type_signature]
```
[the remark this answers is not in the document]
st0012: This is by design. Unless we have use cases to preserve `#:` comments, we don't want this complexity.

lib/rdoc/rdoc.rb
```diff
@@ -429,7 +430,7 @@ def parse_files(files)
 
   def remove_unparseable(files)
     files.reject do |file, *|
-      file =~ /\.(?:class|eps|erb|scpt\.txt|svg|ttf|yml)$/i or
+      file =~ /\.(?:class|eps|erb|rbs|scpt\.txt|svg|ttf|yml)$/i or
```
kou: It's not a change in this PR but `\z` is better than `$` for matching the end of text:


```suggestion
      file =~ /\.(?:class|eps|erb|rbs|scpt\.txt|svg|ttf|yml)\z/i or
```

lib/rdoc/rdoc.rb
```diff
@@ -540,6 +552,73 @@ def generate
...
+      last_modified.nil? || mtime.to_i > last_modified.to_i
+    end
+  end
+
+  ##
+  # Records RBS signature file mtimes so normal generation freshness checks
+  # and the live server watcher can see signature-only edits.
+
+  def record_rbs_signature_mtimes
+    @last_modified.reject! { |file, _| File.extname(file) == '.rbs' }
```
kou:
```suggestion
    @last_modified.reject! { |file, _| rbs_signature_file?(file) }
```

lib/rdoc/rdoc.rb
```diff
@@ -540,6 +552,73 @@ def generate
...
+  end
+
+  def rbs_signature_file?(file) # :nodoc:
+    File.extname(file) == '.rbs'
+  end
+
+  def rbs_signature_mtimes # :nodoc:
+    rbs_signature_files.each_with_object({}) do |file, mtimes|
+      mtime = File.mtime(file) rescue nil
+      mtimes[file] = mtime if mtime
```
kou:
```suggestion
      begin
        mtimes[file] = File.mtime(file)
      rescue SystemCallError
      end
```

lib/rdoc/server.rb
```diff
@@ -310,23 +308,41 @@ def start_watcher(source_files)
     end
   end
 
+  def file_mtimes_for(files)
+    files.each_with_object({}) do |f, h|
+      h[f] = File.mtime(f) rescue nil
+    end
```
kou, line 314: Similar code exists in `lib/rdoc/rdoc.rb` too.

lib/rdoc/server.rb
```diff
@@ -385,6 +410,17 @@ def reparse_and_refresh(changed_files, removed_files)
 
       @store.complete(@options.visibility)
 
+      if rbs_changed || !changed_files.empty?
+        duration_ms = measure do
+          @rdoc.load_rbs_signatures
+          @rdoc.record_rbs_signature_mtimes
+          @rdoc.rbs_signature_files.each do |file|
+            @file_mtimes[file] = File.mtime(file) rescue nil
+          end
```
kou, line 417: This is also similar code.

kou (approved):
+1

lib/rdoc/store.rb
```diff
@@ -443,6 +530,7 @@ def clean_cache_collection(collection) # :nodoc:
   # See also RDoc::Context#remove_from_documentation?
 
   def complete(min_visibility)
+    @type_name_lookup = nil
```
tompng: Is the purpose of this stale lookup table invalidation only for server mode?
If so, I think it can be more explicit it is directly invalidated from `Server#reparse_and_refresh` like:
```ruby
@store.invalidate_type_name_lookup unless changed_files.empty? && removed_files.empty?
# I think it's also OK to always invalidate even if only rbs has changed
```

lib/rdoc/server.rb
```diff
@@ -385,6 +410,17 @@ def reparse_and_refresh(changed_files, removed_files)
 
       @store.complete(@options.visibility)
 
+      if rbs_changed || !changed_files.empty?
+        duration_ms = measure do
+          @rdoc.load_rbs_signatures
```
tompng, line 413: `load_rbs_signatures` which modifies `@store` is always called after `@store.complete`.
Can the order reversed? Load rbs, then call `complete`?

lib/rdoc/parser/ruby.rb
```diff
@@ -458,10 +462,14 @@ def skip_comments_until(line_no_until)
   def consecutive_comment(line_no)
     return unless @unprocessed_comments.first&.first == line_no
     _line_no, start_line, text = @unprocessed_comments.shift
-    parse_comment_text_to_directives(text, start_line)
+    type_signature_lines = extract_type_signature!(text, start_line)
+    result = parse_comment_text_to_directives(text, start_line)
```
tompng: I think ideally signatures should be parsed in `parse_comment_text_to_directives`
because `parse_comment_text_to_directives` stores line number of each directives, useful when implementing warning of invalid directive usage. Line number may be wrong.
But currently we're not using directive's line number yet, so we can consider it later.

lib/rdoc/server.rb
```diff
@@ -385,6 +410,17 @@ def reparse_and_refresh(changed_files, removed_files)
 
       @store.complete(@options.visibility)
 
+      if rbs_changed || !changed_files.empty?
```
tompng: In my understanding, `!changed_files.empty?` is needed because:
- Method may be added if `.rb` files are updated
- Added method may have signature in sig directory

Just an idea, nits, perhaps theres a blocker that I missed,
Is it possible to not assign loaded rbs signatures to `method#type_signature_lines`?
Make type_signature_lines a signature written in source code comment, and reference loaded rbs signatures when generating HTML files.

headius:
> @headius I assumed JRuby doesn't support `rbs` 4.0+ due to [this failure](https://github.com/ruby/rdoc/actions/runs/23756502189/job/69212536022#step:3:300). Is it true that it's not supported? Or I can actually just tweak the CI config to make it work.

I have JRuby support via FFI in progress but there is still quite a bit of work to do. Very much want to get that finished so JRuby users can take advantage of all of the RBS stuff you all are doing.

tompng (approved):
🚀

PR #1676, opened 2026-04-09
Preserve `#` prefix for unresolved cross-references

st0012:
## Summary

When `#name` doesn't resolve to a method, the cross-reference handler was stripping the `#` and returning just the name. Now the original text including `#` is restored when the lookup fails.

This fixes rendering of text like `#no-space-heading` in Markdown paragraphs, where the `#` was silently dropped in the final HTML.

### Refactoring

- `cross_reference` no longer mutates its `name` parameter; uses a separate `display` variable for `#`-stripped text
- `link` returns `nil` for unresolved references instead of returning bare text, letting the caller decide what to display
- Label handling is hoisted out of the `case` branches so it's shared between resolved refs and bare label references (`@foo`)
- Move new test methods before `private` helper section to follow file convention

Found while investigating blockquote lazy continuation in #1627 — `#text` (not a valid heading) inside a blockquote was rendered without the `#`.

### DIFF
--- a/test/rdoc/markup/to_html_crossref_test.rb
+++ b/test/rdoc/markup/to_html_crossref_test.rb
@@ -407,7 +407,7 @@ class RDocMarkupToHtmlCrossrefTest < XrefTestCase
   end
 
   def test_link
-    assert_equal 'n', @to.link('n', 'n')
+    assert_nil @to.link('n', 'n')
 
     assert_equal '<a href="C1.html#method-c-m"><code>m</code></a>', @to.link('m', 'm')
   end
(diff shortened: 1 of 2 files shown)


test/rdoc/markup/to_html_crossref_test.rb
```diff
@@ -435,6 +435,18 @@ def hyper(reference)
...
+
+    # #no should not lose its '#' when it doesn't resolve to a method
+    assert_equal "#no", REGEXP_HANDLING('#no')
+  end
+
+  def test_cross_reference_preserves_explicit_text_for_unresolved
+    # When explicit text is provided, it should be preserved on unresolved refs
+    assert_equal "Foo", @to.cross_reference("Missing", "Foo")
+  end
+
```
kou, line 437: How about moving them between `test_link_class_method_full` and `para`?
It seems that this file uses `def test_XXX` -> `def HELPER_METHODS` order.

kou (approved):
+1

test/rdoc/markup/to_html_crossref_test.rb
st0012: Sorry I missed that. Updated

PR #1686, opened 2026-04-27
Fix CSS to prevent method name overflow

ybiquitous:
This prevents method names from overflowing their container by adding `overflow-wrap: anywhere` to the `.method-name` CSS.

The `anywhere` property value is now widely supported.

Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-wrap#anywhere

Demo:

https://github.com/user-attachments/assets/336747ba-beb9-4fe0-94f2-b0db02960058


st0012 (approved):
Thank you!

PR #1690, opened 2026-05-03
Drop Ripper-based parser

st0012:


test/rdoc/rdoc_store_test.rb
```diff
@@ -172,7 +168,6 @@ def test_all_classes_and_modules
 
     # C8::S1 does not exist. It should not be in the list.
     # class C8; class << something; class S1; end; end; end
```
tompng: This comment (and the same one on line 228) is a comment for the workaround in line 175.
I think we can remove this comment along with the workaround code below.

tompng (approved):

PR #1694, opened 2026-05-05
Remove `ruby-title` class

Earlopain:
It was added for sdoc in https://github.com/ruby/rdoc/commit/acc499bad1984aeb4589334b4b434cb59df992ae
But since https://github.com/rails/sdoc/commit/41a790bad5fb95d10103342dee7a9c8b89880a2b it doesn't use it anymore, and rdoc has no use for it as well.


st0012 (approved):
Thanks!

PR #1696, opened 2026-05-06
Don't create tokens by hand

Earlopain:
There is some awkward code that dances around the fact that the tokens for a method actually contain a 3 extra tokens that don't exist in the source code.

Now `RipperStateLex` is only referenced to actually parse, rest is kept internal

### DIFF
--- a/lib/rdoc/generator/markup.rb
+++ b/lib/rdoc/generator/markup.rb
@@ -86,30 +86,32 @@ end
 class RDoc::MethodAttr
 
   ##
-  # Prepend +src+ with line numbers.  Relies on the first line of a source
-  # code listing having:
-  #
-  #   # File xxxxx, line dddd
-  #
-  # If it has this comment then line numbers are added to +src+ and the <tt>,
-  # line dddd</tt> portion of the comment is removed.
+  # Prepend +src+ with line numbers.
 
   def add_line_numbers(src)
-    return unless src.sub!(/\A(.*)(, line (\d+))/, '\1')
-    first = $3.to_i - 1
-    last  = first + src.count("\n")
-    size = last.to_s.length
+    return if src.empty? || !line
+    start_line = line
+    end_line  = start_line + src.count("\n")
+    number_digits = end_line.to_s.length
 
-    line = first
+    current_line = start_line
     src.gsub!(/^/) do
-      res = if line == first then
-              " " * (size + 1)
-            else
-              "<span class=\"line-num\">%2$*1$d</span> " % [size, line]
-            end
+      res = "<span class=\"line-num\">#{current_line.to_s.rjust(number_digits)}</span> "
 
-      line += 1
+      current_line += 1
       res
+    end
+  end
+
+  ##
+  # Prepend +src+ with a comment that declares its location in the source.
+
+  def add_location_comment(src)
+    path = CGI.escapeHTML(file.relative_name)
+    if options.line_numbers && !src.empty?
+      src.prepend("<span class=\"ruby-comment\"># File #{path}</span>\n")
+    else
+      src.prepend("<span class=\"ruby-comment\"># File #{path}, line #{line}</span>\n")
     end
   end
 
@@ -119,24 +121,25 @@ class RDoc::MethodAttr
   # Prepends line numbers if +options.line_numbers+ is true.
 
   def markup_code
-    return '' unless @token_stream
+    return '' if !@token_stream
 
     src = RDoc::TokenStream.to_html @token_stream
 
-    # dedent the source
-    indent = src.length
-    lines = src.lines.to_a
-    lines.shift if src =~ /\A.*#\ *File/i # remove '# File' comment
-    lines.each do |line|
-      if line =~ /^ *(?=\S)/
-        n = $~.end(0)
-        indent = n if n < indent
-        break if n == 0
-      end
-    end
-    src.gsub!(/^#{' ' * indent}/, '') if indent > 0
+    # add initial whitespace so that the indent gets calculated correctly
+    src.prepend(' ' * @token_stream.first[:char_no]) if source_language == 'ruby' && @token_stream.first
 
-    add_line_numbers(src) if options.line_numbers
+    # dedent the source
+    common_indent = src.length
+    src.scan(/^ *(?=\S)/) do |whitespace|
+      common_indent = whitespace.length if whitespace.length < common_indent
+      break if common_indent == 0
+    end
+    src.gsub!(/^#{' ' * common_indent}/, '') if common_indent > 0
+
+    if source_language == 'ruby'
+      add_line_numbers(src) if options.line_numbers
+      add_location_comment(src)
+    end
 
     src
   end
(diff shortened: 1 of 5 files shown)


test/rdoc/code_object/any_method_test.rb
```diff
@@ -149,42 +149,44 @@ def test_call_seq_returns_nil_if_alias_is_missing_from_call_seq
   end
 
   def test_markup_code
-    tokens = [
-      { :line_no => 0, :char_no => 0, :kind => :on_const, :text => 'CONSTANT' },
```
Earlopain, line 153: This testcase didn't make much sense since only methods have their source code shown. I simply removed it.

lib/rdoc/parser/ruby.rb
```diff
@@ -314,7 +314,7 @@ def parse_comment_tomdoc(container, comment, line_no, start_line)
 
     meth.start_collecting_tokens(:ruby)
     node = @line_nodes[line_no]
-    tokens = node ? visible_tokens_from_location(node.location) : [file_line_comment_token(start_line)]
+    tokens = node ? visible_tokens_from_location(node.location) : []
     tokens.each { |token| meth.token_stream << token }
```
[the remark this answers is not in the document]
Earlopain: Seems correct. Previously they were not empty because it contained the comment

lib/rdoc/parser/ruby.rb
```diff
@@ -385,7 +385,7 @@ def handle_meta_method_comment(comment, directives, node)
         tokens = visible_tokens_from_location(node.location)
         line_no = node.location.start_line
       else
-        tokens = [file_line_comment_token(line_no)]
+        tokens = []
```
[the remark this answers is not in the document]
Earlopain: See above

lib/rdoc/parser/ruby.rb
```diff
@@ -314,7 +314,7 @@ def parse_comment_tomdoc(container, comment, line_no, start_line)
 
     meth.start_collecting_tokens(:ruby)
     node = @line_nodes[line_no]
-    tokens = node ? visible_tokens_from_location(node.location) : [file_line_comment_token(start_line)]
+    tokens = node ? visible_tokens_from_location(node.location) : []
```
tompng, line 317: If a method doesn't have tokens, it used to have a line number information but it's gone.
```ruby
##
# :method: ghost_method

##
# :method:
# :call-seq: ghost_method2() -> Integer
```

Image

I think the line number is important to know why and where the ghost-method is defined.

How about adding a new attribute that represents source location and use it instead of `token_stream.first[:line_no]`

Earlopain: Makes sense, I'll try to fix that. It already exists as `line` on `CodeObject` and the parser sets it as far as I can tell so it shouldn't be too involved.

lib/rdoc/generator/markup.rb
```diff
@@ -86,57 +86,59 @@ class RDoc::CodeObject
...
+  ##
+  # Prepend +src+ with a comment that declares its location in the source.
+
+  def add_location_comment(src)
+    path = CGI.escapeHTML(file.relative_name)
+    if options.line_numbers
+      src.prepend("<span class=\"ruby-comment\"># File #{path}</span>\n")
+    else
+      src.prepend("<span class=\"ruby-comment\"># File #{path}, line #{line}</span>\n")
+    end
```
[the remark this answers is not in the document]
Earlopain: Is this necessary?

lib/rdoc/parser/ruby.rb
Earlopain: Should be done. I moved the tests to the ruby parser, since that is where the line number is actually being set. That said, I only ran rdoc against ruby/prism to check for differences (there are none).

Is there an easy way to run it on ruby/ruby with a local checkout? I couldn't figure it out.

lib/rdoc/generator/markup.rb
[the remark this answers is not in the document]
st0012: No I think it's not. Feel free to ignore it

lib/rdoc/generator/markup.rb
```diff
@@ -86,57 +86,59 @@ class RDoc::CodeObject
...
-  # If it has this comment then line numbers are added to +src+ and the <tt>,
-  # line dddd</tt> portion of the comment is removed.
+  # Prepend +src+ with line numbers.
 
   def add_line_numbers(src)
-    return unless src.sub!(/\A(.*)(, line (\d+))/, '\1')
-    first = $3.to_i - 1
-    last  = first + src.count("\n")
-    size = last.to_s.length
+    start_line = line or return
```
tompng: I think we can add one more early return when src is empty.

```console
bundle exec rdoc path/to/ruby/ruby/lib/delegate.rb --line-numbers --op outputdir
```
Master branch
Image

This branch
Image

If src is empty, line number wasn't and won't be displayed. So perhaps we can change `def add_location_comment` such as:
```ruby
if options.line_numbers && !src.empty? # Add empty check here
  src.prepend("<span class=\"ruby-comment\"># File #{path}</span>\n")
else
  src.prepend("<span class=\"ruby-comment\"># File #{path}, line #{line}</span>\n")
end
```

lib/rdoc/parser/ruby.rb
tompng: I usually use `bundle exec rdoc path/to/ruby/ruby --op ruby_doc` for quick check (though it may not be as same as document deployed to https://docs.ruby-lang.org/en/master/)

### Comparison without `--line-numbers` option
Only `Delegator.html` and `Singleton.html` had a small diff that we can ignore :+1:
```html
# Before
<pre class="ruby"><span class="ruby-comment"># File lib/delegate.rb, line 67</span></pre>

# After
<pre class="ruby"><span class="ruby-comment"># File lib/delegate.rb, line 67</span>
</pre>
```

### Comparison with `--line-numbers` option
Removing spaces before `# File file.rb` seems a great improvement :+1:
Most files doesn't have a diff after normalizing: compare after removing that space with gsub

`Delegator.html` and `Singleton.html` had a diff https://github.com/ruby/rdoc/pull/1696#discussion_r3213392085
```html
# Before
<pre class="ruby">   <span class="ruby-comment"># File ../ruby/lib/delegate.rb</span></pre>

# After
<pre class="ruby"><span class="ruby-comment"># File ../ruby/lib/delegate.rb</span>
<span class="line-num">67</span> </pre>
```

Earlopain: Nice, thanks for the tip!

lib/rdoc/generator/markup.rb
Earlopain: Yup, that works. Added a test for that

lib/rdoc/parser/ruby.rb
```diff
@@ -314,7 +314,7 @@ def parse_comment_tomdoc(container, comment, line_no, start_line)
 
     meth.start_collecting_tokens(:ruby)
     node = @line_nodes[line_no]
```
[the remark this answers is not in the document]
Earlopain: There is no testcase where the line numbers are be different and I can't tell if this suggestion is relevant. No idea what tomdoc is about.

tompng (approved):
Thank you :+1:

PR #1698, opened 2026-05-06
Strip markdown HTML comments within paragraph

tompng:
Standalone HTML comments is converted to `RDoc::Markup::Raw`. However, HTML comments in a paragraph were not handled correctly.
Since RDoc paragraph can't contain Raw, inline HTML comments should be stripped. HTML comments in a paragraph `"\n<!-- -->\n"` will generate double newline `"\n\n"`, so ToHtml also needs to handle this double newline.

Fix this markdown comment rendering
```ruby
##
# :markup: markdown
# <!-- comment1 -->
# hello
# <!-- comment2 -->
# world
```

Before
```html
<!-- comment1 -->
<p>hello &lt;!– comment2 –&gt; world</p>
```
After
```html
<!-- comment1 -->
<p>hello world</p>
```

### DIFF
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -975,6 +975,7 @@ Inline  = Str
         | InlineNote
         | Code
         | RawHtml
+        | StrippedComment
         | Entity
         | EscapedChar
         | Symbol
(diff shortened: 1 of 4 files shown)


kou (approved):
+1

st0012 (approved):

PR #1699, opened 2026-05-06
Remove parser module indirection used in testing

Earlopain:
The ripper parser is gone, so this is not needed anymore

Makes it so that I can easily run individual tests via ruby-lsp It's their limitation really but seems ok to do regardless


kou (approved):
+1

PR #1700, opened 2026-05-07
Drop dead rdoc/markdown handling from rubygems plugin

st0012:
`lib/rubygems_plugin.rb` carries a `begin / rescue LoadError / else` block around `require_relative 'rdoc/markdown'`, gating hook registration on the require succeeding. The accompanying comment explains the guard exists because `markdown.rb` used to be kpeg-generated during `bundle install` and could be absent mid-bootstrap.

That window no longer exists. `lib/rdoc/markdown.rb` is generated and committed, and `RDoc::RubyGemsHook` doesn't reference `RDoc::Markdown` itself — any markdown source actually encountered during doc generation is loaded lazily by the parser dispatcher. So both the guard and the require it gated are dead code.

This drops both, and as a side effect skips parsing ~16k LOC of kpeg-generated parser on every `gem` CLI invocation and every `bundle install` resolution path.

### DIFF
--- a/lib/rubygems_plugin.rb
+++ b/lib/rubygems_plugin.rb
@@ -9,7 +9,6 @@
 # - RDoc is a old version that doesn't have rubygems_plugin.rb.
 
 require_relative 'rdoc/rubygems_hook'
-require_relative 'rdoc/markdown'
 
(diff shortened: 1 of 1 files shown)


st0012:
In the next PR, I will move require into the hooks so we can partially mitigate #1609 and also speed things up

PR #1703, opened 2026-05-09
Remove obsolete -V short flag from --visibility option

st0012:
The `--visibility` option declared `-V` as a short flag, but `--verbose` declares the same `-V` later in the option parser. OptionParser silently lets the later registration win, so `-V` has actually been an alias for `--verbose` rather than `--visibility` — making the visibility short flag obsolete and misleading anyone who reads the source expecting it to work.

This drops `-V` from the `--visibility=VISIBILITY` definition in `lib/rdoc/options.rb`. `--verbose -V` continues to work exactly as before; `--visibility` retains its long form. There is no behavior change for users who were relying on whatever `-V` actually did at runtime.

Adds `test_parse_verbose` to `test/rdoc/rdoc_options_test.rb` so the `-V` → `--verbose` wiring is locked in and a future re-introduction of the duplicate short flag would be caught by the test suite.

### DIFF
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -895,7 +895,7 @@ Usage: #{opt.program_name} [options] [names...]
 
       opt.separator nil
 
-      opt.on("--visibility=VISIBILITY", "-V", RDoc::VISIBILITIES + [:nodoc],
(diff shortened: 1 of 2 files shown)


kou (approved):
+1

PR #1704, opened 2026-05-10
Syntax highlighting using Prism.parse_lex

tompng:
Completely drops ripper dependency.

### Drops `ruby-node` css class used to colorize dstring and backref
backref(`$&`) is now colored as identifier, same as global variable.
dstring(`"#{expr}"`, `` `#{expr}` ``) is now colored as string. Embedded expression parts are colored separately.

### Stops token squashing:
`"string"` was a single string token. Changed to `"` + `string` + `"`
`"string#{:symbol}"` was a single dstring token. Changed to string tokens `"` `string` `#{` `}` `"`. `:` and `symbol is colorized as symbol color.
So there should be lots of diff.

Before
before
After
after

### Fix some unkown bugs
```ruby
# Documenting this code crashed
def foo
  <<~A; :\
  A
  b
end
```


lib/rdoc/markup/to_html.rb
```diff
@@ -321,34 +322,33 @@ def accept_paragraph(paragraph)
     @res << "</p>\n"
   end
 
+  # Generate syntax highlighted html for ruby-like text.
+
+  def parsable_text_to_html(text)
+    tokens = RDoc::Parser::RubyColorizer.colorize(text)
+    result = RDoc::TokenStream.to_html tokens
+    result = result + "\n" unless "\n" == result[-1]
```
kou: This part isn't written by this PR but `end_with?` may be better:

```suggestion
    result = result + "\n" unless result.end_with?("\n")
```

lib/rdoc/parser/ruby.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'prism'
-require_relative 'ripper_state_lex'
+require_relative './ruby_colorizer'
```
kou: `./` is redundant:

```suggestion
require_relative 'ruby_colorizer'
```

kou (approved):
+1

st0012 (approved):

PR #1705, opened 2026-05-11
Remove `MetaMethod` and `GhostMethod`

Earlopain:
RDoc has no special handling for these. `MetaMethod` is unreferenced, and `GhostMethod` was only used for tomdoc.


st0012 (approved):
Thanks for the cleanup!

PR #1710, opened 2026-05-13
Remove `RDoc::MethodAttr#text`

Earlopain:
It is unused and in majority of the cases just set to nil or the empty string

### DIFF
--- a/lib/rdoc/code_object/alias.rb
+++ b/lib/rdoc/code_object/alias.rb
@@ -26,18 +26,12 @@ class RDoc::Alias < RDoc::CodeObject
   attr_reader :singleton
 
   ##
-  # Source file token stream
-
-  attr_reader :text
-
-  ##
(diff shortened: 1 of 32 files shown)


tompng (approved):
Thank you :+1:

st0012:
I think this will fail rbs' rdoc plugin at lib/rdoc_plugin/parser.rb. I think we may need to revert this until we find a way to keep both sides in sync.
Example build: https://github.com/ruby/ruby/actions/runs/25844623560/job/75937026188?pr=16940

Earlopain:
Right, I thought this was internal only. How about accepting `*args` and switching on the length? Then rbs can change whenever it wants and continue to function.

Or just revert if you don't think that is worth the effort. Let me know and I can do a PR

st0012:
I’m in support of cleaning up dead attributes, so if there’s a way to avoid crash without reverting then it’s worth a try

PR #1712, opened 2026-05-14
Resolve markdown crash for note references in labels

### LINKED ISSUE #654: RDoc::Markdown.parse crashes with invalid note reference
Hi, I've been testing rdoc with a Ruby fuzzing tool ([kisaten](https://github.com/zelivans/kisaten)) and found a few bugs. So far I only caught bugs in `markdown.rb`, this is the first one. I am testing against the `master` branch but the bugs should be reproducible in older releases. In my code I used the the following lines:

```ruby
formatter = RDoc::Markup::ToHtml.new(RDoc::Options.new, nil)
html = RDoc::Markdown.parse(data).accept(formatter)
```
## Raw crash
```
/shared/rdoc/lib/rdoc/markdown.rb:739:in `note_for': undefined method `<<' for nil:NilClass (NoMethodError)
	from /shared/rdoc/lib/rdoc/markdown.rb:15154:in `_NoteReference'
	from /shared/rdoc/lib/rdoc/markdown.rb:481:in `apply'
	from /shared/rdoc/lib/rdoc/markdown.rb:9518:in `_Inline'
	from /shared/rdoc/lib/rdoc/markdown.rb:481:in `apply'
	from /shared/rdoc/lib/rdoc/markdown.rb:11888:in `_Label'
	from /shared/rdoc/lib/rdoc/markdown.rb:481:in `apply'
	from /shared/rdoc/lib/rdoc/markdown.rb:11750:in `_Reference'
	from /shared/rdoc/lib/rdoc/markdown.rb:481:in `apply'
	from /shared/rdoc/lib/rdoc/markdown.rb:12275:in `_References'
	from /shared/rdoc/lib/rdoc/markdown.rb:481:in `apply'
	from /shared/rdoc/lib/rdoc/markdown.rb:388:in `parse'
	from /shared/rdoc/lib/rdoc/markdown.rb:777:in `parse'
	from /shared/rdoc/lib/rdoc/markdown.rb:621:in `parse'
	from tst.rb:10:in `<main>'
```

## Example input
I tried to minimize the crashing fuzzer file to the minimum needed to reproduce the bug. If needed, I'll upload the original files as well. It should crash with the following in `data`:
```
[[^0]
```

## Explanation
It seems that `note_ref` is getting called with nil as a parameter from the `_NoteReference` method. It doesn't expect nil and and tries to call `#<<` on it resulting in an unhandled `NoMethodError`.

skatkov:
Fixes #654.

Note references can be parsed during the reference-gathering pass, before footnote ordering is initialized. Previously this could crash with a NoMethodError for malformed reference labels.

`RDoc::Markdown.parse("[[^0]\n")`

Before the fix, it raised:
```ruby
NoMethodError: undefined method `<<' for nil
.../lib/rdoc/markdown.rb:...:in `note_for'
```
Or:
`RDoc::Markdown.parse("[foo[^1]bar]\n\n[^1]: footnote\n")`

### DIFF
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -403,7 +403,7 @@
   # the note order list for proper display at the end of the document.
 
   def note_for ref
-    raise ParseError, "invalid note reference: #{ref}" unless @note_order
+    return unless @note_order
 
     @note_order << ref
 
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -403,6 +403,8 @@
   # the note order list for proper display at the end of the document.
 
   def note_for ref
+    raise ParseError, "invalid note reference: #{ref}" unless @note_order
+
     @note_order << ref
 
     label = @note_order.length
--- a/lib/rdoc/markdown.rb
+++ b/lib/rdoc/markdown.rb
@@ -788,7 +788,7 @@ class RDoc::Markdown
   # the note order list for proper display at the end of the document.
 
   def note_for ref
-    raise ParseError, "invalid note reference: #{ref}" unless @note_order
+    return unless @note_order
 
     @note_order << ref
 
--- a/lib/rdoc/markdown.rb
+++ b/lib/rdoc/markdown.rb
@@ -788,6 +788,8 @@ class RDoc::Markdown
   # the note order list for proper display at the end of the document.
 
   def note_for ref
+    raise ParseError, "invalid note reference: #{ref}" unless @note_order
+
     @note_order << ref
 
     label = @note_order.length
--- a/test/rdoc/rdoc_markdown_test.rb
+++ b/test/rdoc/rdoc_markdown_test.rb
@@ -1005,11 +1005,24 @@ Some text. ^[With a footnote]
   def test_parse_note_invalid_reference
     @parser.notes = true
 
-    error = assert_raise RDoc::Markdown::ParseError do
-      parse "[[^0]\n"
-    end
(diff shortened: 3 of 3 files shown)


lib/rdoc/markdown.rb
```diff
@@ -788,6 +788,8 @@ def note label
   # the note order list for proper display at the end of the document.
 
   def note_for ref
+    raise ParseError, "invalid note reference: #{ref}" unless @note_order
```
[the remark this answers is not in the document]
skatkov: Seems like a valid problem. But different?

I'm proposing a solution in a separate PR:
https://github.com/ruby/rdoc/pull/1713

lib/rdoc/markdown.kpeg
```diff
@@ -403,6 +403,8 @@
   # the note order list for proper display at the end of the document.
 
   def note_for ref
+    raise ParseError, "invalid note reference: #{ref}" unless @note_order
```
tompng: I think the markdown below is valid, but currently fails. Valid markdown shouldn't raise ParseError.
```markdown
[foo[^1]bar]

[^1]: footnote
```

In `def parse markdown`, there's a source code comment:
```ruby
# using note_order on the first pass would be a bug
@note_order      = []
```
The reported error occurs in the first pass (`peg_parse 'References'`), so referencing `@note_order` is already a bug. I think just returning nil is better in this case.

```ruby
# No-op when called in the `References` pass
return unless @note_order
```

skatkov: Great feedback!

I assume, this is fixed now.

tompng (approved):
Looks good :+1:

PR #1713, opened 2026-05-14
Ignore inline notes during Markdown reference parsing

skatkov:
Inline notes inside reference labels can be parsed during the reference-gathering pass, before footnote ordering is initialized. 

`RDoc::Markdown.parse("[foo ^[note]]: /url\n")`

This will crash a Markdown parser with a `NoMethodError`.

This change treats inline-note creation as a no-op until note ordering exists.

### DIFF
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -439,6 +439,8 @@
   def parse markdown
     @references          = {}
     @unlinked_references = {}
+    @footnotes           = nil
+    @note_order          = nil
 
     markdown += "\n\n"
 
@@ -1236,7 +1238,9 @@ InlineNote = &{ notes? }
              @StartList:a
              ( !"]" Inline:l { a << l } )+
              "]"
-             { ref = [:inline, @note_order.length]
+             { raise ParseError, 'invalid inline note' unless @note_order
+
+               ref = [:inline, @note_order.length]
                @footnotes[ref] = paragraph a
 
                note_for ref
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -439,8 +439,6 @@
   def parse markdown
     @references          = {}
     @unlinked_references = {}
-    @footnotes           = nil
-    @note_order          = nil
 
     markdown += "\n\n"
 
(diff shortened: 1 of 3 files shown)


lib/rdoc/markdown.kpeg
```diff
@@ -437,6 +437,8 @@
   def parse markdown
     @references          = {}
     @unlinked_references = {}
+    @footnotes           = nil
+    @note_order          = nil
```
tompng: These nil-resets are not needed. Parsers won't be reused.

```ruby
# Usage: RDoc::Markdown.parse(text)

class RDoc::Markdown
  def self.parse markdown
    parser = new

    parser.parse markdown
  end
end
```

lib/rdoc/markdown.kpeg
```diff
@@ -1234,7 +1236,9 @@ InlineNote = &{ notes? }
              @StartList:a
              ( !"]" Inline:l { a << l } )+
              "]"
-             { ref = [:inline, @note_order.length]
+             { raise ParseError, 'invalid inline note' unless @note_order
```
tompng: Same as #1712, no-op on first pass is better.

lib/rdoc/markdown.kpeg
skatkov: Good tip! Thanks for raising.

lib/rdoc/markdown.kpeg
skatkov: @tompng changed. 

Now it's a no-op; also, adjusted the PR description and title.

tompng:
Looks good. Could you resolve the conflicts?

skatkov:
> Looks good. Could you resolve the conflicts?

All conflicts resolved.

lib/rdoc/markdown.kpeg
```diff
@@ -1236,10 +1236,12 @@ InlineNote = &{ notes? }
...
              "]"
-             { ref = [:inline, @note_order.length]
-               @footnotes[ref] = paragraph a
+             { if @note_order
+                 ref = [:inline, @note_order.length]
+                 @footnotes[ref] = paragraph a
 
-               note_for ref
+                 note_for ref
+               end
```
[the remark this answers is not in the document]
skatkov: That's the intention. It should be a no-op

tompng (approved):
Thank you :+1:

PR #1715, opened 2026-05-14
Change "check syntax" message when it isn't a ruby file

rwstauner:
I saw this message and wanted to track down where it was coming from.

> Before reporting this, could you check that the file you're documenting
> has proper syntax:
>
>   /nix/store/67c99rck1l5z07968ffnxi9ksg3jfvn9-ruby-4.0.4/bin/ruby -c object.c

It probably isn't helpful to suggest "ruby -c" on a C file.

Should we suggest something else?


st0012 (approved):
Thanks!

PR #1728, opened 2026-06-14
Parse RBS files as RDoc input

st0012:
## Problem

RDoc already supports parser classes that register handled file names through [`parse_files_matching`](https://ruby.github.io/rdoc/RDoc/Parser.html#method-c-parse_files_matching), but `.rbs` documentation parsing currently comes from the `rbs` gem's `rdoc/discover` hook, which defines `RDoc::Parser::RBS` and delegates to `RBS::RDocPlugin::Parser` at runtime ([discovery hook](https://github.com/ruby/rbs/blob/master/lib/rdoc/discover.rb), [plugin parser](https://github.com/ruby/rbs/blob/master/lib/rdoc_plugin/parser.rb)). That leaves RBS documentation generation coupled to a separately released plugin even though RDoc already uses RBS for type-signature display ([ruby/rdoc#1665](https://github.com/ruby/rdoc/pull/1665)), and internal RDoc API changes can break the plugin independently of RDoc releases ([ruby/rdoc#1711](https://github.com/ruby/rdoc/pull/1711)).

## Solution

- Add `RDoc::Parser::RBS` inside RDoc, adapted from the existing RBS plugin parser, so `.rbs` files are registered as RDoc input and common RBS declarations/members become RDoc code objects ([RBS plugin parser](https://github.com/ruby/rbs/blob/master/lib/rdoc_plugin/parser.rb), [PR diff](https://github.com/ruby/rdoc/pull/1728/files)).
- Extend existing Ruby-source documentation when matching RBS declarations are parsed, while still generating documentation for classes, modules, methods, attributes, and constants that only appear in RBS ([PR diff](https://github.com/ruby/rdoc/pull/1728/files)).
- Keep RDoc's existing `sig/**/*.rbs` type-signature loading path and skip the released `rbs` discovery hook when RDoc's parser is available, so RubyGems discovery does not replace the in-tree parser ([ruby/rdoc#1665](https://github.com/ruby/rdoc/pull/1665), [discovery hook](https://github.com/ruby/rbs/blob/master/lib/rdoc/discover.rb), [PR diff](https://github.com/ruby/rdoc/pull/1728/files)).


soutaro (approved):
Thanks Stan! 🎉 
I haven't reviewed all of the implementation line by line (I cannot!), but looks great to replace existing RDoc plugin in rbs-gem!

lib/rdoc/parser/rbs.rb
```diff
@@ -0,0 +1,275 @@
...
+    object
+  end
+
+  def rdoc_comment_for(decl, context)
+    rbs_comment = decl.comment if decl.respond_to?(:comment)
+    return unless rbs_comment
+
+    # TODO: Run RBS comments through RDoc's directive preprocessor so
+    # directives like :nodoc: affect the documented object.
+    comment = RDoc::Comment.new rbs_comment.string, context
```
tompng: The second argument should be `@top_level`, an instance of TopLevel. `Comment.new(content, @top_level)` is used in RDoc's code base.
So this method only needs one argument `def rdoc_comment_for(decl)`

lib/rdoc/parser/rbs.rb
```diff
@@ -0,0 +1,275 @@
...
+    comment.format = 'markdown'
+    comment
+  end
+
+  def local_module_name(type_name, namespace)
+    name = type_name.to_s
+    return name if name.start_with?('::')
+
+    namespace_names = namespace ? namespace.to_s.split('::') : []
```
tompng: How about:
```ruby
def local_module_name(type_name, context)
  ...
  namespace_names = TopLevel === context ? context.full_name.split('::') : []
  ...
end
```
Passing `context`, removing `namespace` will simplify `parse_decl` (will be `def parse_decl(decl, context)`)

tompng:
Ideally, superclass/include/extend names should be resolved with RBS's module nesting. This is an existing issue in RBS's rdoc plugin implementation. I think we can improve it in the future.

lib/rdoc/parser/rbs.rb
st0012: Updated 👍

tompng (approved):
Looks good 🚀

PR #1732, opened 2026-06-16
Fix alias detection for multiple `rb_define_singleton_method`

### LINKED ISSUE #1722: `Dir.getwd` shows `Dir.pwd` as its name
https://docs.ruby-lang.org/en/master/Dir.html#method-c-getwd

`Dir.getwd` is an alias for `Dir.pwd`. First I thought it was missing, but it seems to actually assume the name of the alias instead. https://docs.ruby-lang.org/en/master/Dir.html#method-c-pwd

Earlopain:
They were not detected because `singleton_classes` doesn't contain the variable. Fix this simply by moving the check for `singleton_method` further up.

Closes #1722

<details><summary>List of affected classes</summary>
<p>

```
Complex
Date
DateTime
Dir
ENV
File
Pathname
Process
Regexp
Socket
Thread
UNIXSocket
Fiddle::Handle
RubyVM::InstructionSequence
```

</p>
</details> 

grafik
grafik

### DIFF
--- a/lib/rdoc/parser/c.rb
+++ b/lib/rdoc/parser/c.rb
@@ -239,7 +239,7 @@ class RDoc::Parser::C < RDoc::Parser
       if comment.to_s.empty? and existing_method = class_obj.method_list.find { |m| m.name == old_name}
         comment = existing_method.comment
       end
-      add_alias(var_name, class_obj, old_name, new_name, comment)
+      add_alias(var_name, class_obj, old_name, new_name, comment, singleton: @singleton_classes.key?(var_name))
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/rdoc/parser/c.rb
```diff
@@ -996,7 +996,7 @@ def handle_method(type, var_name, meth_name, function, param_count,
     class_obj = find_class var_name, class_name
 
     if existing_method = class_obj.method_list.find { |m| m.c_function == function }
```
tompng: I think it is better to also check singleton here.
```suggestion
    if existing_method = class_obj.method_list.find { |m| m.c_function == function && m.singleton == singleton }
```

Earlopain: Sure, added a test for that. It passed before already though it did call `add_alias` which is unexpected.

tompng: Thanks :+1:

tompng (approved):

PR #1733, opened 2026-06-16
Fix colorization of imaginary/rational numbers

Earlopain:
In prism there is no singualr `RATIONAL`/`IMAGINARY` token. Instead they are more fine-grained.


st0012 (approved):
Thanks!

PR #1734, opened 2026-06-16
Add RDoc logo to documentation

st0012:
On official website:

Screenshot 2026-06-16 at 17 54 31
Screenshot 2026-06-16 at 17 54 28

On GitHub:

Screenshot 2026-06-16 at 17 55 29

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -29,6 +29,7 @@ task rdoc: :generate
 RDoc::Task.new do |doc|
   # RDoc task defaults to /html and overrides the op_dir option in .rdoc_options
   doc.rdoc_dir = "_site" # for GitHub Pages
+  doc.options << "--copy-files" << "rdoc-logo.svg"
 end
 
 task "coverage" do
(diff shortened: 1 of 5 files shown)


rdoc-logo.svg
st0012, line 1: Ideally this should be under `doc/`. But moving it there would cause file path mismatch in generated doc (copied file is always placed on `_site/`, so source will still be `rdoc-logo.svg`) and GH (will use `doc/rdoc-logo.svg`). So I'll leave that to another discussion/PR.

Rakefile
```diff
@@ -29,6 +29,7 @@ task rdoc: :generate
 RDoc::Task.new do |doc|
   # RDoc task defaults to /html and overrides the op_dir option in .rdoc_options
   doc.rdoc_dir = "_site" # for GitHub Pages
+  doc.options << "--copy-files" << "rdoc-logo.svg"
```
kou, line 32: Should we use `.rdoc_options` instead?

kou (approved):
+1

Rakefile
st0012: The `.rdoc_options` equivalent to this option is `static_path`. But in there paths don't get expanded "before" entering `_site`. So it means that it'd try to copy the svg from `_site`, which ofc doesn't exist there.
Using `../rdoc-logo.svg` also fails because it gets rejected directly. I think this is a bug that needs to be fixed separately.

PR #1736, opened 2026-06-17
Fix incorrect method visibility when merging with existing store cache

### LINKED ISSUE #346: RDoc ri store update corrupts visibility information
When creating an RDoc ri store and then later updating it, the visibility information is corrupted, ie. public or protected methods are shown as private methods.

Steps to reproduce (should be possible to reproduce with any Ruby files but for convenience):

```
$ git clone git@github.com:gettalong/cmdparse.git
$ cd cmdparse
$ rdoc -v
4.2.0
$ rdoc --op rdoc-op --ri --quiet lib                                              
$ grep public rdoc-op/CmdParse/CommandParser/parse-i.ri                          
Binary file rdoc-op/CmdParse/CommandParser/parse-i.ri matches
$ rdoc --op rdoc-op --ri --quiet lib
$ grep public rdoc-op/CmdParse/CommandParser/parse-i.ri 
$ grep private rdoc-op/CmdParse/CommandParser/parse-i.ri 
Binary file rdoc-op/CmdParse/CommandParser/parse-i.ri matches
```

4lllex:
Running `Store#save` with already [existing](https://github.com/ruby/rdoc/blob/bb8b2d4587d64df8cd70924fb52021c0941e728a/lib/rdoc/store.rb#L1014) cache incorrectly sets method visibility to `:private` on all methods. `@visibility` gets set during marshal_load:

https://github.com/ruby/rdoc/blob/bb8b2d4587d64df8cd70924fb52021c0941e728a/lib/rdoc/code_object/class_module.rb#L440-L448

and is reused during `merge` -> `add_method`:

https://github.com/ruby/rdoc/blob/bb8b2d4587d64df8cd70924fb52021c0941e728a/lib/rdoc/code_object/class_module.rb#L539-L542

https://github.com/ruby/rdoc/blob/bb8b2d4587d64df8cd70924fb52021c0941e728a/lib/rdoc/code_object/context.rb#L494

Example:

```rb
# rdoc v7.2.0
Dir.mktmpdir do |op_dir|
  store = RDoc::Store.new(RDoc::Options.new.tap { it.op_dir = op_dir })
  top = store.add_file("file.rb")
  meth = RDoc::AnyMethod.new(nil, "method") rescue RDoc::AnyMethod.new("method")
  meth.record_location(top)
  klass = top.add_class(RDoc::NormalClass, "Object").tap { it.add_method(meth) }
  store.save # new cache
  p meth.visibility # => :public
  store.save # merge
  p meth.visibility # => :private
end
```

Fixes #346

### DIFF
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -538,6 +538,7 @@ class RDoc::ClassModule < RDoc::Context
 
     merge_collections method_list, cm.method_list, other_files do |add, meth|
       if add then
+        @visibility = meth.visibility
         add_method meth
       else
         @method_list.delete meth
--- a/test/rdoc/rdoc_store_test.rb
+++ b/test/rdoc/rdoc_store_test.rb
@@ -951,6 +951,14 @@ class RDocStoreTest < XrefTestCase
     assert_empty result.constants
   end
 
+  def test_save_class_merge_method_visibility
+    @s.save_class @klass
+    assert_equal :public, @meth.visibility
+
+    @s.save_class @klass # save again to trigger ClassModule#merge
+    assert_equal :public, @meth.visibility
+  end
+
   def test_save_class_methods
     @s.save_class @klass
 


lib/rdoc/code_object/class_module.rb
```diff
@@ -538,6 +538,7 @@ def merge(class_module)
 
     merge_collections method_list, cm.method_list, other_files do |add, meth|
       if add then
+        @visibility = meth.visibility
```
tompng, line 541: Nice catch on the root cause - `merge` is definitely reusing a stale `@visibility` left over from `marshal_load`.

I'd suggest fixing this one level down instead. Visibility is lexical state, and `Parser::Ruby` already tracks it separately in its own `@visibility` - it never touches `Context#visibility`. So `add_method` overwriting `method.visibility` with the container's `@visibility` is a leftover from the old design, and every caller passing in an already-complete method has to work around it:

- `marshal_load` sets `@visibility` just so `add_method` picks it up
- `merge` (this bug) inherits what `marshal_load` left behind
- `prepare_to_embed` assigns `self.visibility = code_object.visibility` back onto
  the container, with a comment explaining the workaround
- `Parser::Ruby` orders its calls carefully:
  `a.visibility = visibility # should set after adding to container`

Dropping the assignment in `add_method` (and the matching one in `add_attribute`) fixes this at the root and lets all four workarounds go away.
No new argument needed - `AnyMethod` already defaults to `:public` and every caller sets visibility itself, so `marshal_load` just does `method.visibility = visibility` directly.

tompng:
Thank you for the PR. This is fixed by #1800.
Test is added as co-authored commit.

PR #1745, opened 2026-06-25
Bump RDoc to 8.0.0

st0012:
Prepare RDoc 8.0.0 by bumping the version constant from 7.2.0 to 8.0.0.

Darkfish stays available in 8.0; the deprecation docs now target removal in 9.0 instead.

Changes since 7.2.0: https://github.com/ruby/rdoc/compare/v7.2.0..master


kou (approved):
+1

PR #1753, opened 2026-07-22
Regression: Parse methods inside ActiveSupport::Concern included blocks

skatkov:
## Issue
When generating documentation for rails, i'm seeing a following error:

```bash
ActiveRecord/ConnectionHandling.html:
`rdoc-ref:Core.configurations` can't be resolved for ``
```
RDoc does not add ActiveRecord::Core.configurations to its object store. Resulting documentation renders label as plain text instead of linking to the method documentation.

## Context
Rails defines some documented methods inside concern inclusion blocks:

```ruby
module ActiveRecord
  module Core
    extend ActiveSupport::Concern

    included do
      # Returns a fully resolved ActiveRecord::DatabaseConfigurations object.
      def self.configurations
        @@configurations
      end
    end
  end
end
```

Previously this block could be parsed, support for ActiveSupport::Concern#included was originally added in #819. During the migration to the Prism parser and subsequent removal of the Ripper parser, this special handling was not carried over.

## Solution

Handle special-case receiverless `included` calls that have a block.

Their block body is visited directly, without marking it as a generic proc block. Calls without a block continue through the normal visitor path.

Ordinary metaprogramming blocks remain unchanged and method definitions inside them are still ignored.

## Follow-up

@zzak voiced a valid question:

https://github.com/ruby/rdoc/pull/819#issuecomment-871782267

This deserves a follow-up PR's. 

- [ ] `prepended` is analogous and could be a small follow-up using `when :included, :prepended`.
- [ ] `class_methods` needs separate handling.

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -893,11 +893,7 @@ class RDoc::Parser::Ruby < RDoc::Parser
         when :extend
           _visit_call_extend(node)
         when :included
-          if node.block
-            node.block.body&.accept(self)
-          else
-            super
-          end
+          node.block ? node.block.body&.accept(self) : super
         when :public
           super
           _visit_call_public_private_protected(node, :public)
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -892,6 +892,12 @@ class RDoc::Parser::Ruby < RDoc::Parser
           _visit_call_include(node)
         when :extend
           _visit_call_extend(node)
+        when :included
+          if node.block
+            node.block.body&.accept(self)
+          else
+            super
+          end
         when :public
           super
           _visit_call_public_private_protected(node, :public)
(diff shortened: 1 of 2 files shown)


tompng:
Thanks for the PR, but I don't think we should take this approach.

First, on the premise: this behavior didn't come from #819. The pre-Prism parser documented `def` inside *any* block (`Struct.new do ... end`, RSpec helper blocks, ...), which produced many wrong documents. The Prism-based parser intentionally dropped that — see the comment in `visit_block_node`, and the existing test this PR modifies, which asserts that `def` inside a block is ignored.

Second, `included` is not Ruby semantics — it's ActiveSupport::Concern's. In plain Ruby, `class A; included do end; end` raises ArgumentError (`Module#included` is the core inclusion hook, arity 1), and other libraries can define `included` with different block semantics. The parser deliberately interprets only Ruby's own semantics (`attr_accessor`, `include`, visibility, ...), and interpreting a method name by one library's spec doesn't fit there. If RDoc ever supports this, it should be an explicitly scoped extension (e.g. activated by detecting `extend ActiveSupport::Concern`, covering `class_methods do` as well) — but that's a separate discussion.

As a workaround, `# :singleton-method:` / `# :method:` directives work inside blocks today and can cover the ActiveRecord::Core case.

PR #1755, opened 2026-07-23
Speed up the Markdown parser

tompng:
The kpeg-generated Markdown parser addresses the input string by character index, so on non-ASCII input every position lookup scans the string from the start and parsing becomes **quadratic** in the input size — real changelogs (contributor names etc.) always contain some non-ASCII, so this is the default path for large `.md` files. There were also two smaller costs paid by every document.

Three changes, one per commit:

- **Address positions by byte offset.** Generated rule bodies only save and restore `pos` without inspecting it, and all generated regexps are `\G`-anchored, so overriding the five position helpers (`set_string`, `scan`, `match_string`, `get_byte`, `get_text`) with byte-offset versions backed by StringScanner is enough. `get_byte` (the grammar's `.`) still consumes one character and returns its codepoint, so the only observable change is the position representation. The override lives in a hand-written file (`lib/rdoc/markdown/byte_runtime.rb`) prepended from the grammar, so regeneration is unaffected.
- **Guard `HtmlBlock` with a `&"<"` lookahead.** Previously every block position tried ~50 memoized per-tag HTML rules before falling through to `Para`.
- **Skip the Notes pass when the document contains no `"[^"`.** `parse` always ran up to three full passes; a note definition must contain `[^`, which is rare (11 of 594 sampled markdown files). The References pass is left unconditional: `[` appears in almost every real document.

Benchmarks (M-series, Ruby 4.0):

| input | before | after |
|---|---|---|
| `bundle exec rdoc bundler/CHANGELOG.md` (5,526 lines) | 14.5s | 1.5s |
| 300 Japanese paragraphs | 2,911ms | 40ms |
| 1,000 one-word ASCII paragraphs | 160ms | 77ms |

### DIFF
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -452,8 +452,12 @@
     if notes? then
       @footnotes       = {}
 
-      setup_parser markdown, @debug
-      peg_parse 'Notes'
+      # A note definition must contain "[^", so documents without one
+      # have nothing for the collecting pass to find
+      if markdown.include? '[^' then
+        setup_parser markdown, @debug
+        peg_parse 'Notes'
+      end
 
       # using note_order on the first pass would be a bug
       @note_order      = []
--- a/lib/rdoc/markdown.kpeg
+++ b/lib/rdoc/markdown.kpeg
@@ -192,6 +192,10 @@
   require_relative 'markdown/entities'
 
   require_relative 'markdown/literals'
+  require_relative 'markdown/byte_runtime'
+
+  prepend ByteRuntime
+  Literals.prepend ByteRuntime
 
   ##
   # Supported extensions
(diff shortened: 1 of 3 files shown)


skatkov:
Is there a source code for these benchmarks? Curious to see.

tompng:
Benchmark script
```ruby
require 'rdoc'

CASES = {
  '300 Japanese paragraphs' =>
    "これはテストです。日本語の**強調**と`コード`を含む段落。\n\n" * 300,
  '1,000 one-word ASCII paragraphs' =>
    "word\n\n" * 1000,
}

CASES.each do |name, code|
  RDoc::Markdown.parse(code) # warmup
  times = 5.times.map do
    GC.start
    t = Process.clock_gettime(Process::CLOCK_MONOTONIC)
    RDoc::Markdown.parse(code)
    Process.clock_gettime(Process::CLOCK_MONOTONIC) - t
  end
  puts format('%-34s %8.1f ms (min of 5)', name, times.min * 1000)
end

st0012 (approved):

Issue #1591, opened 2026-01-31
Lexical scope document control directives


tompng:
Currently, document control directives such as `:stopdoc:`, `:startdoc:`, `:enddoc:`, `:nodoc:` are buggy.
Documentation and implementation are both wrong. Nobody knows the specification of these directives.
We need to make it clear what the expected specification is, what we are making and fixing.

My proposal is to introduce a lexical scope document control that follows class/module nesting.

<details>
<summary>Q: Isn't it already lexical scope?</summary>
Not actually.

- RDoc stores document_self and document_children to a code object. It's not controlled in parsing phase.
- I think RDoc is trying to emulate lexical scope with wrong mechanism with many bugs.
- Basically enclosed in file scope, but it sometimes leaks (#398).

So we need to implement a real and consistent lexical scope.
</details>

<details>
<summary>Q: Shouldn't it be a macro/preprocessor like `#ifdef` `#endif` instead of lexical scope?</summary>

If these directives are normally written without indent, then it makes sense, but it's not.
Preferred indented usage indicates that it is related to class/module's nesting scope.
```ruby
# Preprocessor-style startdoc/stopdoc. But we don't write like this.
class A
  class B
# :startdoc:
    class C
      X = 1
# :stopdoc:
      Y = 2
    end
  end
end
```

</details>

## Make `:stopdoc:` and `:startdoc:` lexical scope.

Don't modify container code object's state, just skip documenting in stopdoc region.
When reached an `end` that close a class/module, it also closes `:startdoc:` or `:stopdoc:` in the class/module scope.
This is similar to current behavior but clearly defines what happen if there is no corresponding `:startdoc:`.

```ruby
class A
  # :stopdoc:
  class B
    # (stopdoc applied)
    # :startdoc:
    # (startdoc applied)
  end # (close startdoc at L5)
  # (stopdoc applied)
end # (close stopdoc at L2)
```

<details>
<summary>Existing bugs</summary>

```ruby
class A
  # :stopdoc:
  class B
    def nodoc; end
  end
  # :startdoc:
end

class A
  class B
    # not documented
    def shoulddoc; end
  end
end
```
</details>

## Change `:enddoc:` to stop documenting in the current lexical scope

It will be almost the same as `:stopdoc:` but can't do startdoc.
Note that `:enddoc:` is rarely used (in `ruby/*`), mostly used at toplevel (in `rails/rails`).

```ruby
class A
  class B
    # :enddoc:
    class C
      # (document disabled here too)
[...]

Issue #55, opened 2011-07-21
:stopdoc: directive produces Object reference


thinkerbot:
The :stopdoc: directive produces a reference to Object as a class/module when used like this:

``` ruby
# :stopdoc:
class Example
  def method_name
  end
end
```

I generated the RDoc with:

```
rdoc example.rb
```

Incidentally, no reference to Object is generated unless the method is defined.  This is ok:

``` ruby
# :stopdoc:
class Example
end
```

Using RDoc 3.8.

PR #1763, opened 2026-08-01
Make `:stopdoc:`/`:startdoc:`/`:enddoc:` lexically scoped in the Ruby parser

tompng:
Part of #1591. This PR only changes the region directives (`:stopdoc:`/`:startdoc:`/`:enddoc:`). `:nodoc:` semantics are unchanged and will be addressed separately.

## Problem

These directives are currently implemented as mutations of shared `CodeObject` state (`document_self`/`document_children`/`done_documenting`). Since classes are open and shared across files in the store, a `:stopdoc:` without a matching `:startdoc:` leaks into later reopenings of the class within the same file. The actual behavior is a product of mutation order, and neither the documentation nor the implementation defines what these directives mean.

## New semantics

The parser now tracks the document control state (`:startdoc:`/`:stopdoc:`/`:enddoc:`) as a lexical region:

- The state is inherited when entering a class/module scope and automatically restored at its `end`. A `:stopdoc:` can no longer leak out of the scope it was written in.
- `:startdoc:` inside a nested scope re-enables documentation for that scope only.
- `:enddoc:` disables documentation for the rest of the current scope and cannot be cancelled within it, even by `:startdoc:`. It no longer marks the container object as `done_documenting` globally.
- Classes/modules defined inside a suppressed region are still created as namespaces (marked as ignored) and their bodies are visited, so an inner `:startdoc:` works. They become documentable when they receive documentable contents outside the region, possibly from another file. Namespaces that never receive contents are removed from the store on `Store#complete` (`Context#remove_from_documentation?` now also covers ignored contexts, which is what `CodeObject#ignore` was documented to do in GH #55).
- Document control scopes are the syntactic units that change the definition context (`class`/`module` bodies, `class << self`, and definition-creating blocks). `if`/`while`/`begin` do not create scopes, consistent with Ruby's own scoping rules.

## Bug fixes

On master, these directives mutate the state of whatever object they happen
to be applied to, so their effect escapes the scope they were written in.
Some examples of what this PR fixes:

**A `:stopdoc:` inside `class << self` no longer leaks out of the singleton
scope:**

```ruby
class Net::HTTPResponse
  class << self
    # :stopdoc:
    def internal; end
  end

  # master: NOT documented / this PR: documented
  attr_reader :http_version
end
```

This is exactly what happened to [the Ruby 4.0 documentation](https://docs.ruby-lang.org/en/4.0/Net/HTTPResponse.html)

**A `:stopdoc:` no longer leaks into later reopenings of the class.** Note
that on master the two ways of reopening `A::B` below do not even agree with
each other: `f` is documented but `g` is not, because the leaked state only
affects the nested-class form. The behavior was an artifact of the
implementation, not a specification:

```ruby
class A
  # :stopdoc:
  def internal; end
end

class A::B
  # master: documented / this PR: documented
  def f; end
end

class A
  class B
    # master: NOT documented / this PR: documented
    def g; end
  end
end
```

## Compatibility notes

- `:nodoc:` is untouched. However, `:startdoc:` additionally calls `start_doc` on the current container, because `module Net #:nodoc:` followed by `:stopdoc:`/`:startdoc:` regions is a common pattern that expects `:startdoc:` to make the container documentable again. This is a transitional measure until `:nodoc:` gets lexical semantics, at which point the pattern itself becomes unnecessary. Containers created inside a suppressed region are excluded, so a bare `:startdoc:` cannot revive an empty hidden class.
- A comment linked to a `class << X` line is now processed in the enclosing scope. Nothing consumes such a comment as documentation, and processing it inside the singleton scope would make a `:startdoc:` written there expire at the singleton's `end` (this pattern appears in net/http).
- Region directives in modifier position (e.g. `class Foo # :stopdoc:` at the end of a line) are no longer applied. There are no occurrences of this form in ruby/ruby, rails, or the bundled gems.

## Validation

Compared generated HTML for ruby/ruby `lib` and six Rails frameworks between master and this branch. The sets of generated files are identical, and every remaining difference is the directives working as their authors intended:

- Module aliases written inside `:stopdoc:` regions (`Ripper = Prism::Translation::Ripper` in prism's shim.rb, `HashWithIndifferentAccess = ...` in activesupport) are no longer documented. This also removes a duplicated "Ripper" entry from the navigation and stops bare "Ripper" mentions (which refer to the stdlib Ripper) from being cross-referenced to `Prism::Translation::Ripper`. Note: the old accidental link on `HashWithIndifferentAccess` in Hash.html was arguably useful; writing the full name in the comment restores it.
- `Net::HTTPResponse` regains its `include Net::HTTPHeader`. The old implementation lost it because a `:stopdoc:` inside `class << self` leaked out of the singleton scope; net/http.rb contains a "this is to fix bug in RDoc" workaround comment for this exact class of leak.
- `ActionDispatch::Journey` (`# :nodoc:`) no longer appears in the navigation. Its page is still generated because it has documented children.
- Namespaces fully contained in suppressed regions (e.g. `Net::HTTPBadResponse` region in net/http.rb, `Timeout::Request`) no longer produce empty pages.

## Follow-ups (out of scope here)

- Lexical/shallow semantics for `:nodoc:` (the hard part of #1591); the `start_doc` compatibility shim and the `received_nodoc` guard in promotion can be removed then.
- Warnings for degenerate usages (redundant `:stopdoc:`, directives after `:enddoc:`, standalone `:nodoc:` in the middle of a class body).
- A visibility trace output (`file:line: stopdoc applied to X`) as proposed in https://github.com/ruby/rdoc/issues/1591#issuecomment-3923094155; the parser now has the lexical information to make this cheap.
- The C parser still uses the old CodeObject-state mechanism and is intentionally untouched.

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -915,8 +915,10 @@ class RDoc::Parser::Ruby < RDoc::Parser
     handle_modifier_directive(mod, end_line)
     unless document_suppressed?
       # In a :stopdoc:/:enddoc: region, the container is still created as a
-      # namespace (the body is visited so that an inner :startdoc: works)
-      # but is not recorded to this file nor documented
+      # namespace but is not recorded to this file nor documented.
+      # The body is also visited: an inner :startdoc: re-enables documentation
+      # in a :stopdoc: region (not in an :enddoc: region), and nested
+      # namespaces need to be created for later promotion from other files
       mark_container_documentable(owner) if owner.is_a?(RDoc::ClassModule)
       if mod.ignored?
         mark_container_documentable(mod)
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -353,6 +353,7 @@ class RDoc::Parser::Ruby < RDoc::Parser
   # Signature section in the comment
 
   def parse_comment_tomdoc(container, comment, line_no, start_line)
+    return if document_suppressed?
     return unless signature = RDoc::TomDoc.signature(comment)
 
     name, = signature.split %r%[ \(]%, 2
@@ -561,6 +562,9 @@ class RDoc::Parser::Ruby < RDoc::Parser
     new_methods = []
     @container.methods_matching(names, singleton) do |m|
       if m.parent != @container
+        # A copy of an ancestor's method must not be documented
+        # in a :stopdoc:/:enddoc: region
+        next if document_suppressed?
         m = m.dup
         record_location(m)
         new_methods << m
@@ -583,6 +587,10 @@ class RDoc::Parser::Ruby < RDoc::Parser
 
   def change_method_to_module_function(names)
     @container.set_visibility_for(names, :private, false)
+    # In a :stopdoc:/:enddoc: region, the visibility of instance methods still
+    # changes but the singleton method copies must not be documented
+    return if document_suppressed?
+
     new_methods = []
     @container.methods_matching(names) do |m|
       s_m = m.dup
@@ -845,10 +853,11 @@ class RDoc::Parser::Ruby < RDoc::Parser
     constant.store = @store
     constant.line = start_line
     constant.is_alias_for_path = alias_path
-    mark_container_documentable(owner) if owner.is_a?(RDoc::ClassModule)
-    record_location(constant)
     handle_modifier_directive(constant, start_line)
     handle_modifier_directive(constant, end_line)
+    # A constant marked :nodoc: must not make an ignored owner documentable
+    mark_container_documentable(owner) if constant.document_self && owner.is_a?(RDoc::ClassModule)
+    record_location(constant)
     owner.add_constant(constant)
     return unless alias_path
     mod =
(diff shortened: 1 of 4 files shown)


lib/rdoc/parser/ruby.rb
```diff
@@ -152,6 +152,50 @@ def initialize(top_level, content, options, stats)
...
+  # Applies document control directives (:startdoc:, :stopdoc: and :enddoc:)
+  # to the current lexical scope. The state is restored when the enclosing
+  # class/module scope is closed.
+
+  def apply_document_control_directive(directives)
+    directives.each do |directive, _|
+      case directive
+      when 'startdoc', 'stopdoc'
+        # :enddoc: cannot be cancelled within the scope, even by :startdoc:
+        next if @doc_state == :enddoc
```
st0012: Should we emit a warning in this case?

lib/rdoc/parser/ruby.rb
```diff
@@ -755,14 +834,18 @@ def find_or_create_lexical_constant_owner_name(constant_path)
...
+    return if document_suppressed?
+
     owner, name = find_or_create_lexical_constant_owner_name(constant_name)
     return unless owner
 
     constant = RDoc::Constant.new(name, rhs_name, comment)
     constant.store = @store
     constant.line = start_line
     constant.is_alias_for_path = alias_path
+    mark_container_documentable(owner) if owner.is_a?(RDoc::ClassModule)
```
st0012: If `Foo` is the owner and has no documentation, and we do `class Foo::Bar # :nodoc:`, does it still make sense to make it documentable? Would it be too complicated to only mark owner as documentable when `constant` is documented?

tompng: Nice catch. Fixed with 91eef33a96f175b79c91ff4e8d97df3f2113f7ec

st0012 (approved):

PR #1766, opened 2026-08-07
Remove the unmatched constant alias queue

tompng:
`unmatched_constant_alias` is not used anymore.

Value was set in `add_module_alias(from, from_name, to, file)` when `from` is nil, but all path that calls `add_module_alias` doesn't pass `nil` to it.


st0012 (approved):
Good cleanup 👍

PR #1769, opened 2026-08-09
Make markup Raw, List, and ListItem inherit from Element

st0012:
These classes are indeed markup components and already implement the Element contract, so they should inherit from the right abstract class.


tompng (approved):
:+1:

PR #1771, opened 2026-08-10
Support newer comment token

kddnewton:
In the next major version of Prism, the comment token is going to be split up when it is a non embdoc comment and when it terminates an expression. This diff makes the tests support both.

In theory this should have no user-facing consequences because it does not matter if you colorize the newline or not.

This will let me merge https://github.com/ruby/prism/pull/4205


tompng (approved):
Thank you

PR #1772, opened 2026-08-11
Do not parse non-Ruby files as Ruby when :markup: is markdown or tomdoc

### LINKED ISSUE #1597: RDoc parses `.c` file as ruby code if markup is set to markdown in the first 3 lines

sjh9714:
## Summary

Fixes #1597.

A file whose first three lines contain `:markup: markdown` (or `tomdoc`) was always handed to the Ruby parser: `RDoc::Parser.for` consults `use_markup` before any file-name-based selection, and `use_markup` never sees the file name. The issue's `file.c` was therefore documented as Ruby source. The buggy line carried `# TODO Ruby should be returned only when the filename is correct`.

This change passes the file name into `use_markup` and returns the Ruby parser for these two markups only when the file name actually selects the Ruby parser. Other files fall through to `for`'s existing shebang/extension logic, so `file.c` is parsed by the C parser — while the `:markup:` directive still sets the comment format, since that is handled independently by `RDoc::Markup::PreProcess` (its directive regexp already matches `/* */` comments).

Prior art: #1613 attempted this and was self-closed over the parser-vs-format conflation; gating only the parser choice on the file name and leaving format handling to PreProcess avoids that problem.

Behavior notes:

- `.rb`/`.rbw` files and extensionless scripts with a Ruby shebang keep the Ruby parser (the shebang branch in `for` is unchanged).
- A `README.md` with a markup comment was previously also parsed as Ruby; it now falls through to the Markdown parser.
- `use_markup(content)` without a file name (existing public API) behaves as before.

## Changes

- `RDoc::Parser.use_markup` accepts an optional `file_name` and consults `can_parse_by_name` for the `markdown`/`tomdoc` markups; `RDoc::Parser.for` passes the file name through.
- Regression tests: a `.c` file with `/* :markup: markdown */` selects `RDoc::Parser::C` end to end (fails on master with `RDoc::Parser::Ruby`), plus `use_markup` unit coverage for `.rb` vs `.c` with both markups.

## Testing

- `bundle exec ruby -Ilib -Itest test/rdoc/parser/parser_test.rb`
- `bundle exec rake` (2520 unit tests + 20 RubyGems integration tests, 100% pass)
- `bundle exec rubocop lib/rdoc/parser.rb test/rdoc/parser/parser_test.rb`

### DIFF
--- a/lib/rdoc/parser.rb
+++ b/lib/rdoc/parser.rb
@@ -170,7 +170,7 @@ class RDoc::Parser
     file_name = top_level.absolute_name
     return if binary? file_name
 
-    parser = use_markup content, file_name
+    parser = use_markup content
 
     unless parser then
       parse_name = file_name
@@ -230,22 +230,18 @@ class RDoc::Parser
   # Any comment style may be used to hide the markup comment.
   #
   # The +tomdoc+ and +markdown+ markups name comment formats rather than
-  # parsers, so they select the Ruby parser only when the given +file_name+
-  # is a Ruby file (or when no +file_name+ is given).
+  # parsers, so no parser is selected for them and RDoc::Parser.for picks one
+  # from the file name instead.
 
-  def self.use_markup(content, file_name = nil)
+  def self.use_markup(content)
     markup = content.lines.first(3).grep(/markup:\s+(\w+)/) { $1 }.first
 
     return unless markup
 
-    # tomdoc and markdown are comment formats, not parsers, so they imply
-    # the Ruby parser only when the file name does not select another parser
-    if %w[tomdoc markdown].include? markup then
-      return RDoc::Parser::Ruby if file_name.nil? or
-        can_parse_by_name(file_name) == RDoc::Parser::Ruby
-
-      return
-    end
+    # tomdoc and markdown name a comment format, not a parser.  Skipping them
+    # keeps the search below from matching RDoc::Parser::Markdown for a file
+    # that is not markdown.
+    return if %w[tomdoc markdown].include? markup
 
     markup = Regexp.escape markup
 
--- a/lib/rdoc/parser.rb
+++ b/lib/rdoc/parser.rb
@@ -170,7 +170,7 @@ class RDoc::Parser
     file_name = top_level.absolute_name
     return if binary? file_name
 
-    parser = use_markup content
+    parser = use_markup content, file_name
 
     unless parser then
       parse_name = file_name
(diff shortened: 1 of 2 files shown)


lib/rdoc/parser.rb
```diff
@@ -228,14 +228,24 @@ def self.remove_modeline(content)
...
 
     return unless markup
 
-    # TODO Ruby should be returned only when the filename is correct
-    return RDoc::Parser::Ruby if %w[tomdoc markdown].include? markup
+    # tomdoc and markdown are comment formats, not parsers, so they imply
+    # the Ruby parser only when the file name does not select another parser
+    if %w[tomdoc markdown].include? markup then
+      return RDoc::Parser::Ruby if file_name.nil? or
+        can_parse_by_name(file_name) == RDoc::Parser::Ruby
```
tompng: If the filename is correct, and `can_parse_by_name(file_name) == RDoc::Parser::Ruby` is true, we only need to return nil. The correct parser will be selected by filename(`can_parse_by_name file_name`) in `Parser.for` (Line 187).
```ruby
return if %w[tomdoc markdown].include? markup
```
And we don't need file_name parameter
```diff
- def self.use_markup(content, file_name = nil)
+ def self.use_markup(content)
```
Comment needs to be updated. The main reason we need to skip tomdoc/markdown here, is not because that the content may be ruby or c, but to avoid false positive parser detection in the later `/^#{markup}$/i =~ parser.name.sub(/.*:/, '')` check.

sjh9714:
Thanks, that is simpler — pushed in aa1b1b2.

`use_markup` now just returns nil for `tomdoc`/`markdown` and no longer takes `file_name`, so `Parser.for` selects the parser by name as you described. The comment now says the real reason for the skip: keeping the `/^#{markup}$/i` search below from matching `RDoc::Parser::Markdown`.

One consequence worth flagging: `use_markup` returns nil instead of `RDoc::Parser::Ruby` for these two markups now, so `test_class_use_markup_markdown` and `test_class_use_markup_tomdoc` assert nil. I added `test_class_for_markup_markdown_ruby_file` so the "`.rb` with `:markup: markdown` still gets the Ruby parser" path stays covered through `Parser.for`.

Ran locally: `bundle exec rake normal_test` (2519 tests, 0 failures) and `bundle exec rubocop`.

tompng (approved):
:+1:

PR #1777, opened 2026-08-17
Deduplicate symlinked source files before parsing

### LINKED ISSUE #1776: RDoc follows symlinked source tree twice and crashes with SystemStackError
RDoc crashes with `SystemStackError: stack level too deep` when generating documentation for the Pagy source.

## Steps to reproduce
This is reproducible with rdoc 8.0.0 and with latest master branch.
```
gem install rdoc
rdoc --version

git clone https://github.com/ddnexus/pagy.git

cd pagy

rdoc
```

This returns:
```
Parsing sources...
100% [338/338]  /home/sk/Code/pagy/tasks/test.thor
uh-oh! RDoc had a problem:
stack level too deep

```

## Expected result
RDoc should generate documentation without throwing errors

## Probable cause
Pagy's release source contains two copies of its gem source:

```text
gem/lib/pagy/modules/i18n/i18n.rb
docs/gem/lib/pagy/modules/i18n/i18n.rb
```

Second file is symlinked.

RDoc traverses both paths, parses the same physical files twice, and creates a self-referencing superclass for Pagy::I18n::KeyError.

## Propose fix
Deduplicate discovered files by File.realpath before parsing, while preserving support for source trees accessible only through symlinks.

skatkov:
Fixes #1776 

The fix deduplicates discovered files using File.realpath while retaining the first logical path. The same filtering is applied to the live-server watcher, preventing the alternate path from being rediscovered later.

### DIFF
--- a/lib/rdoc/rdoc.rb
+++ b/lib/rdoc/rdoc.rb
@@ -124,7 +124,7 @@ class RDoc::RDoc
 
     file_list = normalized_file_list files, true, @options.exclude
 
-    file_list = remove_unparseable(file_list)
+    file_list = remove_duplicate_files(remove_unparseable(file_list))
 
     if file_list.count {|name, mtime|
          file_list[name] = @last_modified[name] unless mtime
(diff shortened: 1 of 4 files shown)


tompng (approved):
:+1:

PR #1787, opened 2026-08-23
Associate comments with their markup source

st0012:
`RDoc::Comment.new` now requires an explicit `markup_source:`, normally the shared `RDoc::TopLevel` of the parser. Comments parsed from text remain lazy and cached, and their documents derive file provenance from that source. The public `location=` and `document=` writers are removed.

Parser call sites reuse their existing `TopLevel`. `RDoc::Comment.from_document` provides the controlled path for restored or imported documents. RBS remains the only parser that uses this path to build synthetic merged comments.

### DIFF
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -39,7 +39,7 @@ class RDoc::ClassModule < RDoc::Context
   #
   # Before marshalling:
   # - +location+ is an RDoc::TopLevel
-  # - +comments+ are Strings
+  # - +comments+ are Strings or RDoc::Comment objects
   #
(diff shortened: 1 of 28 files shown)


lib/rdoc/code_object/class_module.rb
```diff
@@ -39,7 +39,7 @@ class RDoc::ClassModule < RDoc::Context
   #
   # Before marshalling:
   # - +location+ is an RDoc::TopLevel
-  # - +comments+ are Strings
+  # - +comments+ are Strings or RDoc::Comment objects
```
st0012, line 42: This reflects existing behaviour. It's not introduced in this PR

PR #1788, opened 2026-08-26
Reduce memory allocations for RI and POT generators

skatkov:
Each documented method currently retains an array of `RDoc::Parser::RubyColorizer::ColoredToken` objects and their fragmented source strings.

This data is unnecessary for:

- RI, whose serialized method representation excludes token_stream
- POT, which extracts translatable documentation comments
- Coverage reports, which generate no method-source pages

Large generated SDKs can contain millions of tokens. Avoiding those allocations substantially reduces parsing time and memory without changing generated output.

## Implementation

Generators can declare whether they require method source:
```ruby
class RDoc::Generator::RI
  def self.store_method_source? = false
end
```

```ruby
class RDoc::Generator::POT
  def self.store_method_source? = false
end
```

The Ruby parser checks that capability before syntax highlighting a method body.

Generators without the capability default to true, preserving compatibility with existing third-party generators.

This also allows third-party generators (like rdoc-markdown) to skip this costly process if they don't have need for `method.token_stream` or `method.markup_code`.  

If `store_method_source?` was disabled, but `token_stream` was still accessed, an error will be raised. 

## Benchmark
Here is a code that was used for benchmark this fix:
https://github.com/skatkov/rdoc-store-method-source-benchmark

RDoc 8.0.0 and `google-api-client` 0.53.0 on Ruby 4.0.6:

| Variant | Parse RSS | Peak RSS | Colored tokens | Live heap slots | Time |
| --- | ---: | ---: | ---: | ---: | ---: |
| Baseline | 1,069.4 MiB | 1,508.4 MiB | 3,900,930 | 10,031,484 | 61.72s |
| Optimized | 463.6 MiB | 730.0 MiB | 0 | 2,229,446 | 51.62s |

- Parse RSS reduction: 56.6%
- Peak RSS reduction: 51.6%
- Generated RI output: identical

### DIFF
--- a/lib/rdoc/options.rb
+++ b/lib/rdoc/options.rb
@@ -1325,6 +1325,16 @@ Usage: #{opt.program_name} [options] [names...]
   end
 
   ##
+  # Returns whether syntax-highlighted method source should be stored.
+
+  #: () -> bool
+  def store_method_source?
+    return false if @coverage_report
+
+    !@generator.respond_to?(:store_method_source?) || @generator.store_method_source?
+  end
+
+  ##
   # Finds the template dir for +template+
 
   def template_dir_for(template)
--- a/lib/rdoc/code_object/any_method.rb
+++ b/lib/rdoc/code_object/any_method.rb
@@ -114,6 +114,17 @@ class RDoc::AnyMethod < RDoc::MethodAttr
   end
 
   ##
+  # Current token stream.
+
+  def token_stream
+    unless options.store_method_source?
+      raise RDoc::Error, "method source for #{full_name} was not stored; set store_method_source? to true"
+    end
+
+    super
+  end
+
+  ##
   # Whether the method has a call-seq.
 
   def has_call_seq?
--- a/lib/rdoc/generator/markup.rb
+++ b/lib/rdoc/generator/markup.rb
@@ -138,9 +138,9 @@ class RDoc::MethodAttr
   # Prepends line numbers if +options.line_numbers+ is true.
 
   def markup_code
-    return '' if !@token_stream
+    return '' if !(tokens = token_stream)
 
-    src = RDoc::TokenStream.to_html @token_stream
+    src = RDoc::TokenStream.to_html tokens
 
     # dedent the source
     common_indent = src.length
--- a/lib/rdoc/generator/pot.rb
+++ b/lib/rdoc/generator/pot.rb
@@ -81,6 +81,8 @@ class RDoc::Generator::POT
     end
   end
 
+  def self.store_method_source? = false
+
   private
   def extract_messages
     extractor = MessageExtractor.new(@store)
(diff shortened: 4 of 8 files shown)


lib/rdoc/options.rb
```diff
@@ -1324,6 +1324,16 @@ def setup_generator(generator_name = @generator_name)
...
   end
 
+  ##
+  # Returns whether syntax-highlighted method source should be stored.
+
+  #: () -> bool
+  def store_method_source?
+    return false if @coverage_report
+
+    !@generator.respond_to?(:store_method_source?) || @generator.store_method_source?
```
tompng, line 1334: We need this `respond_to?` for custom generators :+1:
But can we define `store_method_source? = true` for all existing generators in this repository?

lib/rdoc/options.rb
```diff
@@ -1324,6 +1324,16 @@ def setup_generator(generator_name = @generator_name)
     end
   end
 
+  ##
+  # Returns whether syntax-highlighted method source should be stored.
+
+  #: () -> bool
+  def store_method_source?
```
tompng, line 1331: RDoc::RubyGemsHook#generate parses files once and then generates both formats from the same store, swapping the generator afterwards (simplified):
```ruby
class RDoc::RubyGemsHook
  def generate
    @rdoc.store = RDoc::Store.new(parse_options)
    @rdoc.parse_files parse_options.files

    document 'ri',    options, @ri_dir   if @generate_ri   and (@force or not File.exist? @ri_dir)
    document 'aliki', options, @rdoc_dir if @generate_rdoc and (@force or not File.exist? @rdoc_dir)
  end
end
```
parse_options takes its generator from the gem's spec.rdoc_options. So if a gem specifies --format=ri there, parsing skips method source tokens, and the subsequent aliki generation raises:
```ruby
method source for Foo#bar was not stored; set store_method_source? to true
```

Reproduction: https://gist.github.com/tompng/5fe25eaeb6cb53164a3c49374d330914

Since the generator can be swapped after parsing like this, deriving the decision from options.generator at parse time is not reliable. I think store_method_source should be an explicit option that RubyGemsHook sets before parsing, e.g.:
```ruby
parse_options.store_method_source = Generator::Aliki.store_method_source? || Generator::RI.store_method_source?
@rdoc.parse_files parse_options.files
```
(strictly, only the generators that will actually run need to be OR-ed)

```ruby
def store_method_source?
  # return true or false if it's explicitly set
  return @store_method_source unless store_method_source.nil?
  # fallback path (@coveragage_report, @generator.store_method_source?)`
end
```

skatkov: Good catch! 

I didn't know this was even possible. I'll take a closer look

skatkov: I'm closing this PR, suggested approach is not workable.

I'm leaning toward changing from eager code highlight approach to lazy one and working on a prototype in this PR:
https://github.com/ruby/rdoc/pull/1792

PR #1792, opened 2026-08-28
Lazy code colorization

skatkov:
Please see the prior PR https://github.com/ruby/rdoc/pull/1788

In that PR it was correctly established that eager generation of `RDoc::Parser::RubyColorizer::ColoredToken` objects creates a heavy performance/system burden for all generators, but not all generators need these objects.

My suggested approach in that PR was to establish a generator property that would stop this eager code colorization process.  This approach is not functional, because rdoc itself can switch generators, but use the same @store object. 

In this PR I want to suggest an alternative solution:
- Instead of doing this eagerly, do code colorization lazily (only if stream_token was actually called).
- This would also ensure, that no new properties or methods will be introduced. And this will help us keep backward compatibility, while improving performance.

## Solution
This change stores a copied method-source slice in a deferred token stream and materializes the existing mutable token array only when token_stream is accessed. It also switches the initial file scan from Prism.parse_lex to Prism.parse, preserves indentation and heredoc boundaries, and keeps the existing public token-stream behavior for HTML and third-party generators.

## Benchmark
I have done a benchmark on google-client-api gem (one of the biggest and popular gems in the ecosystem). This could be found here:
https://gist.github.com/skatkov/7e509742651586c1b993a8b15b10a8da

#### After Parse
| Variant | RSS | Peak RSS | Colored | Deferred | Raw strings | Raw source | Prism nodes | Prism tokens | Heap slots |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| released | 958.8 MiB | 965.6 MiB | 3900930 | 0 | 0 | 0.0 MiB | 0 | 0 | 9998568 |
| current | 478.2 MiB | 478.2 MiB | 0 | 46571 | 617 | 57.0 MiB | 0 | 0 | 2433708 |

#### After Generate
| Variant | RSS | Peak RSS | Colored | Deferred | Raw strings | Raw source | Prism nodes | Prism tokens | Heap slots |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| released | 1356.7 MiB | 1391.4 MiB | 3900930 | 0 | 0 | 0.0 MiB | 0 | 0 | 11901244 |
| current | 719.7 MiB | 755.3 MiB | 0 | 46571 | 617 | 57.0 MiB | 0 | 0 | 4336429 |

| Variant | Parse | Generate | Total | GNU peak RSS |
| --- | ---: | ---: | ---: | ---: |
| released | 24.31s | 30.40s | 58.18s | 1391.4 MiB |
| current | 14.62s | 28.59s | 45.43s | 755.3 MiB |

Parse RSS change: -50.1%
Generated RSS change: -47.0%
Peak RSS change: -45.7%
Parse time change: -39.8%
Generate time change: -5.9%
Total time change: -21.9%
Outputs identical: yes

### Aliki, 100 files

```sh
./benchmark.rb --format aliki --limit 100
```

#### After Parse

| Variant | RSS | Peak RSS | Colored | Deferred | Raw strings | Raw source | Prism nodes | Prism tokens | Heap slots |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| released | 96.3 MiB | 96.9 MiB | 253392 | 0 | 0 | 0.0 MiB | 0 | 0 | 708268 |
| current | 65.4 MiB | 65.4 MiB | 0 | 2701 | 2701 | 0.9 MiB | 0 | 0 | 206955 |

#### After Generate

| Variant | RSS | Peak RSS | Colored | Deferred | Raw strings | Raw source | Prism nodes | Prism tokens | Heap slots |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| released | 219.1 MiB | 220.4 MiB | 253392 | 0 | 0 | 0.0 MiB | 0 | 0 | 1037598 |
| current | 210.3 MiB | 210.3 MiB | 253392 | 0 | 0 | 0.0 MiB | 0 | 0 | 1038957 |

| Variant | Parse | Generate | Total | GNU peak RSS |
| --- | ---: | ---: | ---: | ---: |
| released | 1.40s | 44.09s | 46.10s | 220.4 MiB |
| current | 0.90s | 44.56s | 46.04s | 209.5 MiB |

Parse RSS changed by -32.1%, generated RSS by -4.0%, peak RSS by -4.9%,
parse time by -35.7%, generation time by +1.1%, and total time by -0.1%.
Outputs were identical.

## Results

According to the benchmark, we can see that the lazy approach brings significant memory improvements even for generators that actually output highlighted code. 

In cases of 'RI', generator that has no use for highlighted code, we can see significant memory improvements (~50%), but also ~20% speed improvements.

### DIFF
--- a/lib/rdoc/generator/markup.rb
+++ b/lib/rdoc/generator/markup.rb
@@ -138,8 +138,7 @@ class RDoc::MethodAttr
   # Prepends line numbers if +options.line_numbers+ is true.
 
   def markup_code
-    tokens = token_stream
-    return '' if !tokens
+    return '' if !(tokens = token_stream)
 
     src = RDoc::TokenStream.to_html tokens
 
--- a/lib/rdoc/generator/markup.rb
+++ b/lib/rdoc/generator/markup.rb
@@ -138,9 +138,10 @@ class RDoc::MethodAttr
   # Prepends line numbers if +options.line_numbers+ is true.
 
   def markup_code
-    return '' if !@token_stream
+    tokens = token_stream
+    return '' if !tokens
 
-    src = RDoc::TokenStream.to_html @token_stream
+    src = RDoc::TokenStream.to_html tokens
 
     # dedent the source
     common_indent = src.length
--- a/lib/rdoc/generator/markup.rb
+++ b/lib/rdoc/generator/markup.rb
@@ -138,9 +138,9 @@ class RDoc::MethodAttr
   # Prepends line numbers if +options.line_numbers+ is true.
 
   def markup_code
-    return '' if !(tokens = token_stream)
+    return '' unless token_stream
 
-    src = RDoc::TokenStream.to_html tokens
+    src = RDoc::TokenStream.to_html token_stream
 
     # dedent the source
     common_indent = src.length
--- a/lib/rdoc/parser/ruby_colorizer.rb
+++ b/lib/rdoc/parser/ruby_colorizer.rb
@@ -51,7 +51,7 @@ module RDoc::Parser::RubyColorizer
         nodes.concat(node.compact_child_nodes)
       end
 
-      @token_streams.replace(@token_streams.to_h { |node_id,| [node_id, staged_tokens.fetch(node_id)] })
+      @token_streams = staged_tokens
       @source = nil
     end
   end
--- a/lib/rdoc/parser/ruby_colorizer.rb
+++ b/lib/rdoc/parser/ruby_colorizer.rb
@@ -16,6 +16,7 @@ module RDoc::Parser::RubyColorizer
     def initialize(source)
       @source = source
       @streams = {}
+      @mutex = Mutex.new
     end
 
     #: (Integer) -> ^() -> Array[ColoredToken]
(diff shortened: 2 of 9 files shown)


skatkov:
@tompng this is available for review. What's your opinion on this approach?

lib/rdoc/token_stream.rb
```diff
@@ -75,14 +77,14 @@ def collect_tokens(language)
...
-    @token_stream.pop
+    token_stream.pop
   end
 
   ##
   # Current token stream
 
   def token_stream
-    @token_stream
+    @token_stream.is_a?(Proc) ? @token_stream.call : @token_stream
```
tompng: Multiple types in a single ivar is a style that RDoc frequently used, and that we struggled maintaining, and trying to remove.
(Example: `@comment` in a code object was `Comment|String|Document(only on marshal loaded CodeObject)|nil`)
How about splitting ivar like this?
```ruby
def collect_tokens(language, loader: nil)
  @token_stream_loader = loader
  ...
end

def token_stream
  if @token_stream_loader
    @token_stream = @token_stream_loader.call
    @token_stream_loader = nil
  end
  @token_stream
end
```

lib/rdoc/parser/ruby_colorizer.rb
```diff
@@ -10,6 +10,48 @@ module RDoc::Parser::RubyColorizer
 
   ColoredToken = Struct.new(:kind, :text)
 
+  # Defers colorization for all nodes in one source file until first access.
+  class DeferredContext
+    #: (String) -> void
+    def initialize(source)
+      @source = source
+      @streams = {}
+      @mutex = Mutex.new
```
tompng: I think we don't need mutex, until it is requested.
RDoc is not thread-safe. Using mutex here sends a wrong message that RDoc is thread-safe.
I think `token_stream_loader` is not thread safe too. `@streams[node_id]` may be overwritten by an empty array after it was initialized and proc is pushed in another thread.

lib/rdoc/parser/ruby_colorizer.rb
```diff
@@ -10,6 +10,48 @@ module RDoc::Parser::RubyColorizer
...
+    end
+
+    #: (Integer) -> ^() -> Array[ColoredToken]
+    def token_stream_loader(node_id)
+      tokens = nil
+      (@streams[node_id] ||= []) << ->(resolved_tokens) { tokens = resolved_tokens }
```
tompng: This is a pattern that register a listener proc, but the API of DeferredContext is not pub-sub style.
If so, the code below would be more simple
```ruby
def token_stream_loader(node_id)
  @token_streams[node_id] = nil
  -> { token_stream_for(node_id) }
end

def token_stream_for(node_id)
  materialize unless materialized?
  @token_streams[node_id]
end
```

skatkov: Indeed it looks easier on the eyes!

lib/rdoc/token_stream.rb
skatkov: Sounds good to me. Fixed

lib/rdoc/parser/ruby_colorizer.rb
```diff
@@ -10,6 +10,47 @@ module RDoc::Parser::RubyColorizer
...
+      nodes = [program_node]
+      until nodes.empty?
+        node = nodes.pop
+        if @token_streams.key?(node.node_id)
+          staged_tokens[node.node_id] = RDoc::Parser::RubyColorizer.partial_colorize(@source, node, prism_tokens)
+        end
+        nodes.concat(node.compact_child_nodes)
+      end
+
+      @token_streams.replace(@token_streams.to_h { |node_id,| [node_id, staged_tokens.fetch(node_id)] })
```
tompng: Can we use
```ruby
@token_streams = staged_tokens
```
or
```ruby
L43: if @token_streams.key?(node.node_id)
L44:   @token_streams[node.node_id] = RDoc::Parser::RubyColorizer.partial_colorize(@source, node, prism_tokens)
L45: end
```

lib/rdoc/generator/markup.rb
```diff
@@ -138,9 +138,9 @@ def add_location_comment(src)
   # Prepends line numbers if +options.line_numbers+ is true.
 
   def markup_code
-    return '' if !@token_stream
+    return '' if !(tokens = token_stream)
```
tompng: Can we use this?
```ruby
tokens = token_stream
return '' if !tokens
```
or simply
```ruby
return '' unless token_stream
src = RDoc::TokenStream.to_html token_stream
```

tompng (commented):
It basically looks good :smile:

lib/rdoc/generator/markup.rb
skatkov: We can!

lib/rdoc/parser/ruby_colorizer.rb
skatkov: I removed this object traversal logic.

test/rdoc/rdoc_token_stream_test.rb
```diff
@@ -126,4 +153,5 @@ def initialize
     end.new
     assert_equal "", foo.tokens_to_s
   end
+
```
tompng: Could you delete this empty line?

skatkov: done

tompng (approved):
Looks good. Thank you :+1:
I'll merge this after conflict is resolved

skatkov:
Rebased!

PR #1793, opened 2026-08-28
Allow less strict attribute argument parsing

### LINKED ISSUE #1790: Parser should be less strict about attribute arguments
The prism parser (the only parser since v8.0.0) is strict about arguments to `attr_reader`, `attr_writer`, `attr_accessor`, and `attr`.  It _only_ parses as an attribute when _all_ arguments are symbols.

As an example, the rdoc for `Net::IMAP::Config` intentionally took advantage of the looser parsing done by rdoc 7.2.  That class redefines `attr_reader`, `attr_writer` and `attr_accessor` to add keyword arguments for type validation/coercion and defaults:
```ruby
      # Seconds to wait until a connection is opened.
      #
      # Applied separately for establishing TCP connection and starting a TLS
      # connection.
      #
      # If the IMAP object cannot open a connection within this time,
      # it raises a Net::OpenTimeout exception.
      #
      # See Net::IMAP.new and Net::IMAP#starttls.
      #
      # The default value is +30+ seconds.
      attr_accessor :open_timeout, type: Integer, default: 30
```

rdoc 7.2 simply ignores the unknown keyword args, and parses this no differently from `attr_accessor :open_timeout`.

In my opinion, this is a bug in rdoc 8.0 which should be fixed.

nevans:
This updates argument parsing for the `attr`, `attr_reader`, `attr_writer`, and `attr_accessor` methods, so they behave more like rdoc 7.2's parser.

* Previously, the prism parser's `_visit_call_attr_reader_writer_accessor` only recognized attributes when _all_ arguments are symbols.
* Now, all symbol or string arguments are used and all others are ignored.

This also updates inferred attribute name parsing for the `:attr:`, `:attr_reader:`, `:attr_writer:`, and `:attr_accessor:` directives, to avoid creating unnamed attributes from unparsable arguments.

* Previously, non-string/non-symbol arguments would create unnamed (`nil`) attributes.
* Now, any non-string/non-symbol arguments are ignored.

As an example, the rdoc for `Net::IMAP::Config` intentionally took advantage of the looser parsing done by rdoc 7.2. That class redefines `attr_reader`, `attr_writer` and `attr_accessor` to add keyword arguments for type validation/coercion and defaults:

```ruby
      # Seconds to wait until a connection is opened.
      #
      # Applied separately for establishing TCP connection and starting a TLS
      # connection.
      #
      # If the IMAP object cannot open a connection within this time,
      # it raises a Net::OpenTimeout exception.
      #
      # See Net::IMAP.new and Net::IMAP#starttls.
      #
      # The default value is +30+ seconds.
      attr_accessor :open_timeout, type: Integer, default: 30
```

rdoc 7.2 simply ignored the unknown keyword args, and parses this no differently from `attr_accessor :open_timeout.`

Fixes #1790.

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -412,7 +412,7 @@ module RDoc
           case directive
           when 'attr', 'attr_reader', 'attr_writer', 'attr_accessor'
             attributes = [param] if param
-            attributes ||= call_node_name_arguments(node) if is_call_node
+            attributes ||= call_node_name_arguments(node).compact if is_call_node
             rw = directive == 'attr_writer' ? 'W' : directive == 'attr_accessor' ? 'RW' : 'R'
           when 'method'
             method_name = param if param
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -1215,6 +1215,11 @@ module RDoc
           names.all? ? names : nil
         end
 
+        def call_node_name_arguments(call_node)
+          names = @scanner.call_node_name_arguments(call_node).compact
+          names unless names.empty?
+        end
+
         def symbol_arguments(call_node)
           arguments_node = call_node.arguments
           return unless arguments_node && arguments_node.arguments.all? { |arg| arg.is_a?(Prism::SymbolNode)}
@@ -1329,8 +1334,8 @@ module RDoc
 
         def _visit_call_attr_reader_writer_accessor(call_node, rw)
           return if @scanner.in_proc_block
-          names = symbol_arguments(call_node)
-          @scanner.add_attributes(names.map(&:to_s), rw, call_node.location.start_line) if names
+          names = call_node_name_arguments(call_node)
+          @scanner.add_attributes(names, rw, call_node.location.start_line) if names
         end
 
         class MethodSignatureVisitor < Prism::Visitor # :nodoc:
--- a/test/rdoc/parser/ruby_test.rb
+++ b/test/rdoc/parser/ruby_test.rb
@@ -1780,19 +1780,19 @@ end
         ##
         # :attr:
         # attrs
-        add_my_method :attr1, :attr2
+        add_my_method :attr1, "attr2", (ignored)
         ##
         # :attr_reader:
         # readers
-        add_my_method :reader1, :reader2
+        add_my_method :reader1, ignored, "reader2"
         ##
         # :attr_writer:
         # writers
-        add_my_method :writer1, :writer2
+        add_my_method :writer1, :writer2, kwarg: ignored
         ##
         # :attr_accessor:
         # accessors
-        add_my_method :accessor1, :accessor2
+        add_my_method ignored, :accessor1, :accessor2
 
         # :stopdoc:
 
--- a/test/rdoc/parser/ruby_test.rb
+++ b/test/rdoc/parser/ruby_test.rb
@@ -1605,14 +1605,14 @@ end
         # attrs
         attr :attr1, :attr2
         # readers
-        attr_reader :reader1, :reader2
+        attr_reader :reader1, "reader2"
         # writers
-        attr_writer :writer1, :writer2
+        attr_writer "writer1", :writer2
         # accessors
         attr_accessor :accessor1, :accessor2
         # :stopdoc:
         attr :attr3, :attr4
-        attr_reader :reader3, :reader4
+        attr_reader :reader3, "reader4"
         attr_writer :write3, :writer4
         attr_accessor :accessor3, :accessor4
       end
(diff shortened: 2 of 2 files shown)


lib/rdoc/parser/ruby.rb
```diff
@@ -1348,7 +1354,7 @@ def _visit_call_private_constant(call_node)
 
     def _visit_call_attr_reader_writer_accessor(call_node, rw)
       return if @scanner.in_proc_block
-      names = symbol_arguments(call_node)
+      names = initial_symbol_arguments(call_node)
```
st0012: I feel `symbol_arguments` should already work like `initial_symbol_arguments`.

@tompng why was it designed to return `nil` when there's any non-symbol arguments? What's the case we're guarding against?

tompng: To handle `alias_method`.
We need to prevent `alias_method :foo, bar, :baz, blah: true` treated as `alias_method :foo, :baz`.
In all other case, we can just use the non-strict version.

lib/rdoc/parser/ruby.rb
```diff
@@ -1240,6 +1240,12 @@ def symbol_arguments(call_node)
       arguments_node.arguments.map { |arg| arg.value.to_sym }
     end
 
+    def initial_symbol_arguments(call_node)
+      return unless arguments = call_node.arguments&.arguments
+      symbol_args = arguments.take_while {|arg| arg.is_a?(Prism::SymbolNode) }
```
tompng: How about using `arguments.grep(Prism::SymbolNode)` instead of `take_while` and renaming the method?
I think it's worth adding a test like this, which is possible in the original `attr_accessor` spec
```ruby
# document :foo, :bar and :baz
attr_accessor :foo, *ignored1, :bar, (ignored2), :baz
```

nevans: Yeah, I was trying to be a little bit conservative, since I wasn't sure if the change was intentional.  But I'm even happier with that.

I feel like maybe _both_ methods should be renamed, to avoid ambiguity confusion...
* this grep-based variation => `any_symbol_arguments` or `every_symbol_argument`
  (or `grep_symbol_arguments`?  I prefer "any".)
* `symbol_arguments` => `exclusively_symbol_arguments`?

I _(originally)_ added "node" to the names to emphasize that this is only matching symbol literal nodes.
_(edited to add)_ But I dropped it in the version I pushed, both to be consistent with the original naming, and also because the methods are _returning_ symbols and only working with literal SymbolNode is an implementation detail that can be inferred from context.

nevans: Do we want to preserve _any_ form of `test_undocumentable_attributes`?  That is, what if the arguments aren't variables or expressions that're too dynamic or too complex to extract names from, but are instead integers or other literals which suggest an entirely different API is being used?

I'm updating the PR to just repurpose that test for your `grep` suggestion.  But that test case (added in fde99f1be) makes me think that you _maybe_ had a specific scenario you wanted to prohibit?

lib/rdoc/parser/ruby.rb
nevans: I feel like maybe expanding the scope to everything other than attributes either deserves separate PRs (automatically calling out in separate line items in generated release notes), or (at the very least) renaming this PR.  :)

nevans:
@st0012 @tompng As a separate question, do we want to also allow string literal arguments?  Because I'm pretty sure that _all_ of the Module methods in question (`attr`, `attr_reader`, `attr_writer`, `attr_accessor`, `alias_method`, `public`, `private`, `protected`, `module_function`, `private_class_function`, `public_class_function`, `private_constant`, `public_constant`) also support string arguments.  The only ones I was unsure about were the const visibility methods, and I just tested those (they work with strings).

_BUT_ it's pretty _weird_ to use string literals rather than symbol literals.  Sure, for convenience, maybe we sometimes pass some variables or expressions into them which might evaluate to strings, to avoid the extra `.to_sym` calls.  But using string _literals_ just isn't done (at least, I haven't seen it in long time)!

So if it _is_ done, _maybe_ there's a reason, and _maybe_ (for _some_ of these) that reason could be that it's a simple (if somewhat hacky) way to _avoid_ rdoc from picking up on them?

But I dunno... I can only really see that argument working for the attribute methods and `alias_method`.  For the visibility methods, I think we should err on the side of reading **every** visibility change we can reasonably parse.  So I'd want the visibility methods to support `StringNode` in addition to `SymbolNode`.

Anyway, after talking myself through it (and writing it down above), I'm leaning towards allowing both `SymbolNode` and `StringNode` for _all_ of these.  What do you think?

lib/rdoc/parser/ruby.rb
tompng: In my mind, there were two possible design:
1. Only accept `attr_accessor :a, :b` and reject `attr_accessor :a, opt:`
2. Accept `attr_accessor :a, 1, expr, :b, true` and also `attr_accessor :a, opt:`

And I selected the first one. `test_undocumentable_attributes` is testing this design decision.
IMO, if were going to be conservative, reject a usage that might be a completely different API, `attr_accessor :a, opt:` is one of the case that would be rejected. It's not ruby core attr_accessor usage. It's not a bug.
I think it's OK to change the design from 1 to 2, and allow any dynamic expression mixed with symbols.

nevans:
Related to that, I feel like maybe the following should _not_ be ignored, but should issue a warning instead?

```ruby
private def self.foo = bar
# => WARNING: "private def self.foo" makes "M#foo" private, not "M.foo"
private_class_method def foo = bar
# => WARNING: "private_class_method def foo" makes "M.foo" private, not "M#foo"
private def to_s.foo = bar
# => WARNING: "private def (expr).foo" makes "M#foo' private, not "(expr).foo"
```

Because `def (expr).foo` always returns `:foo`, all of those _will_ make a(n existing) method private.  Even if that isn't what the user expected, it's what they've done.

That's more scope than I want to push into this PR, but I'm curious about your thoughts?

tompng:
Accepting StringNode is the original behavior in the old Ripper based parser. The Prism based parser just ported that behavior.

I think there is no strong reason to keep that behavior because string isn't used.
Actual code found on GitHub:
https://github.com/search?q=%22attr_accessor%28%5C%22%22+language%3ARuby&type=code&l=Ruby
Other: Dynamic string, RuboCop BAD example comment, test code.

I'm fine either way - keeping or dropping the behavior accepting StringNode.

>`private def self.foo = bar`

I think it's OK to do that: Warn and make instance method `foo` private if it exist.

nevans:
> Accepting StringNode is the original behavior in the old Ripper based parser. The Prism based parser just ported that behavior.

@tompng Just to be clear: the Prism based parser does _not_ currently accept string arguments for these methods.  I think you're saying the old parser _did_ support string arguments.  So you're okay with me _adding_ support for them to the Prism parser?

nevans:
It looks like the current prism parser supports either StringNode or SymbolNode as the first argument for meta method comments (`call_node_name_arguments`) _(edited to add:) **or** for meta programmed attributes (with an attribute directive)_.

st0012:
I think we should accept string arguments in those calls.

nevans:
@st0012 @tompng I've updated the PR to use `@scanner.call_node_name_arguments`, which is doing almost exactly what we want.  I've also updated the parsing of metaprogrammed attribute name detection, so unparsed arguments don't generate unnamed attributes.

It might make sense to extract that method (and several others which process prism nodes and don't use any ivars) into a shared utilities module, but that's out of scope for this PR, IMO.

I'll post a second (stacked) PR to update the all of the visibility methods.

tompng (approved):
👍

PR #1794, opened 2026-08-29
Add memoized page lookup index to `RDoc::Store`

skatkov:
`page(name)` currently resolves page lookups by scanning every `TopLevel` in `@files_hash` each time it’s called (`each_value.each_with_object(...)`). In large documentation sets (or in server/live reload flows where lookups happen repeatedly), this results in repeated linear scans. 

This PR introduces `@page_index` as a memoized secondary index keyed by both `page_name` and `base_name` to avoid that repeated work and make page lookup O(1) on steady-state access.

## Benchmarks

Based on my findings, speed-up is noticeable in gems with many pages. This improvement doesn't affect smaller gems.

-> TODO: add exact numbers

### DIFF
--- a/lib/rdoc/store.rb
+++ b/lib/rdoc/store.rb
@@ -140,6 +140,8 @@ class RDoc::Store
     @modules_hash = {}
     @files_hash   = {}
     @text_files_hash = {}
+    # lazy cache, used for quick lookup by page name.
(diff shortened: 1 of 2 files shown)


skatkov:
So, after testing more... this speed up is barely <2% insignificant

PR #1795, opened 2026-08-29
Enforce nested class and module declarations

st0012:
And remove redundant namespace in constant references (e.g. `RDoc::Foo` can often just be `Foo`)


test/rdoc/support/test_case.rb
```diff
@@ -19,200 +19,202 @@
...
-  # Shortcut for RDoc::Markup::Document.new with +contents+
+    ##
+    # Creates an RDoc::Comment with +text+ which was defined on +top_level+.
+    # By default the comment has the 'rdoc' format.
 
-  def doc(*contents)
-    @RM::Document.new(*contents)
-  end
+    def comment(text, top_level = @top_level, language = nil)
+      comment = Comment.new text, top_level, language
```
st0012, line 116: This is pretty redundant. I'll open a separate PR to remove it.

PR #1799, opened 2026-08-30
Hide embedded Racc runtime from generated API docs

### LINKED ISSUE #1762: Embedded Racc runtime is exposed as part of RDoc's generated API documentation
#### Description

RDoc's generated RD parsers contain a complete embedded copy of the Racc
runtime:

```text
lib/rdoc/rd/block_parser.rb
lib/rdoc/rd/inline_parser.rb
```

In both files, the embedded section is delimited by:

```ruby
###### racc/parser.rb begin
```

and:

```ruby
###### racc/parser.rb end
```

RDoc parses this embedded implementation as if it were part of RDoc's public
API. As a result, generating RDoc's own documentation creates a `Racc` module
page.

The embedded comments also contain this reference:

```text
rdoc-ref:lib/racc/rdoc/grammar.en.rdoc
```

It produces two unresolved-reference warnings, one from each generated
parser:

```text
classes/Racc.html: `rdoc-ref:lib/racc/rdoc/grammar.en.rdoc` can't be resolved
classes/Racc.html: `rdoc-ref:lib/racc/rdoc/grammar.en.rdoc` can't be resolved
```

Relevant generated files:

https://github.com/ruby/rdoc/blob/v8.0.0/lib/rdoc/rd/block_parser.rb

https://github.com/ruby/rdoc/blob/v8.0.0/lib/rdoc/rd/inline_parser.rb

#### Expected behavior

The embedded Racc runtime is an implementation detail of the RD parsers. It
should not create `Racc`, `Racc::Parser`, or related pages in RDoc's API
documentation.

The actual RDoc classes defined after the embedded section, such as
`RDoc::RD`, must remain documentable.

#### Suggested fix

Wrap the embedded runtime in RDoc exclusion directives:

```ruby
# :stopdoc:
###### racc/parser.rb begin

# Embedded Racc runtime

###### racc/parser.rb end
# :startdoc:
```

This must be integrated into parser generation or its post-processing, since
the generated `.rb` files are committed and regenerated from the `.ry`
sources.

I tested these directive positions:

- `Racc` is no longer added to the documentation store;
- `RDoc::RD` remains documented;
- the two unresolved references disappear.

A regression test could parse both generated files and assert:

```ruby
refute store.find_class_or_module("Racc")
assert store.find_class_or_module("RDoc::RD")
```

OskarEichler:
## Summary

Exclude the two generated RD parser implementations from RDoc input through `.rdoc_options`.

This keeps the generated parser files and generator untouched while preventing their embedded Racc runtime internals from appearing in the public API documentation.

Fixes #1762.

## Verification

- full RDoc dry-run completes successfully;
- the final diff contains only the two `.rdoc_options` exclusions;
- parser runtime code and generated files are unchanged.

### DIFF
--- a/.rdoc_options
+++ b/.rdoc_options
@@ -8,6 +8,8 @@ exclude:
 - AGENTS.md
 - CLAUDE.md
 - lib/rdoc/markdown.kpeg
+- lib/rdoc/rd/block_parser.rb
+- lib/rdoc/rd/inline_parser.rb
 
 footer_content:
   DOCUMENTATION:
--- a/Rakefile
+++ b/Rakefile
@@ -61,7 +61,18 @@ def generate_parser_file(parser_file)
   when '.ry' # need racc
     sh "bundle", "exec", "racc", "-l", "-E", "-o", parsed_file, parser_file
     File.open(parsed_file, 'r+') do |f|
-      newtext = "# frozen_string_literal: true\n#{f.read}"
+      generated = f.read
+      runtime_begin = "###### racc/parser.rb begin\n"
+      runtime_end = "###### racc/parser.rb end\n"
+
+      unless generated.include?(runtime_begin) && generated.include?(runtime_end)
+        raise "Racc runtime markers are missing from #{parsed_file}"
+      end
+
+      generated.sub!(runtime_begin, "# :stopdoc:\n#{runtime_begin}")
+      generated.sub!(runtime_end, "#{runtime_end}# :startdoc:\n")
+
+      newtext = "# frozen_string_literal: true\n#{generated}"
       f.rewind
       f.write newtext
     end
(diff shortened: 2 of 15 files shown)


st0012:
We shouldn't update generated docs this way. Can we exclude them in `.rdoc_options` instead?

OskarEichler:
Updated as requested: the generated parser files and generator are untouched now, and both generated RD parsers are excluded in `.rdoc_options`. I also merged current master; the PR diff is now only those two exclude entries. A full dry-run documentation build completes successfully.

st0012 (approved):

st0012:
Can you please update the PR description too?

OskarEichler:
Updated the PR description so it now reflects the final `.rdoc_options`-only implementation and its verification. Thanks for the reminder.

st0012:
Thanks for the PR 👍

thyresias:
If I understand well, the .rdoc_options file is not copied in the gem installation.
As a result, the issue will remain when building the documentation from the installed gem.
Maybe these should be in `s.rdoc_options = [...]` in `rdoc.gemspec`?

PR #1800, opened 2026-08-30
Remove `Context#visibility` and `Context#current_line_visibility`

### LINKED ISSUE #346: RDoc ri store update corrupts visibility information

tompng:
Fixes #346

Context doesn't need visibility. It's a parser's lexical state.
Visibility of added method won't be modified in `add_method`, so we can remove all workarounds related to method/attribute/alias and its visibility: Set/modify visibility after adding method/attribute/alias.
    
AnyMethod#marshal_load and Attr#marshal_load fall back to :public when a cache written by an older RDoc has nil visibility, the same as Constant#marshal_load.
    
rdoc_stats_test passed a NormalClass to add_constant, which only worked because Context responded to `visibility`.

Added test: from #1736

### DIFF
--- a/lib/rdoc/code_object/class_module.rb
+++ b/lib/rdoc/code_object/class_module.rb
@@ -84,7 +84,6 @@ class RDoc::ClassModule < RDoc::Context
     klass.sections.concat mod.sections
     klass.unmatched_alias_lists = mod.unmatched_alias_lists
     klass.current_section = mod.current_section
-    klass.visibility = mod.visibility
 
     klass.classes_hash.update mod.classes_hash
     klass.modules_hash.update mod.modules_hash
@@ -397,7 +396,6 @@ class RDoc::ClassModule < RDoc::Context
     @done_documenting  = false
     @parent            = nil
     @temporary_section = nil
-    @visibility        = nil
     @classes           = {}
     @modules           = {}
 
@@ -442,11 +440,10 @@ class RDoc::ClassModule < RDoc::Context
 
     array[8].each do |type, visibilities|
       visibilities.each do |visibility, methods|
-        @visibility = visibility
-
         methods.each do |name, file|
           method = RDoc::AnyMethod.new name, singleton: type == 'class'
           method.record_location RDoc::TopLevel.new file
+          method.visibility = visibility
           add_method method
         end
       end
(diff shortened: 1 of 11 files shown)


test/rdoc/rdoc_store_test.rb
```diff
@@ -951,6 +951,14 @@ def test_save_class_merge_constant
     assert_empty result.constants
   end
 
+  def test_save_class_merge_method_visibility
+    @s.save_class @klass
+    assert_equal :public, @meth.visibility
+
+    @s.save_class @klass # save again to trigger ClassModule#merge
+    assert_equal :public, @meth.visibility
```
st0012, line 964: Hmm I'm not sure if I understand what this test checks against 🤔

test/rdoc/rdoc_context_test.rb
```diff
@@ -194,7 +193,7 @@ def test_add_include
 
   def test_add_method
     meth = RDoc::AnyMethod.new 'old_name'
-    meth.visibility = nil
+    meth.visibility = :public
```
st0012: Why are these necessary? Should method objects instantiate with default public visibility?

st0012 (approved):
Cleanup looks good, just have some questions

test/rdoc/rdoc_context_test.rb
tompng: This is not needed.
I added this line when I mistakenly thought the default is `nil`. 
removed

lib/rdoc/code_object/class_module.rb
```diff
@@ -442,8 +440,6 @@ def marshal_load(array) # :nodoc:
 
     array[8].each do |type, visibilities|
       visibilities.each do |visibility, methods|
```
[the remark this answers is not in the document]
tompng: Add `method.visibility = visibility` and a test for it

test/rdoc/rdoc_store_test.rb
tompng: :private case and description added

Issue #1735, opened 2026-06-17
RDoc emits duplicate method-c-new anchors when a class defines both self.new and initialize


skatkov:
I'm working on a `rdoc-markdown` gem and stumbled into unexpected behavior that seems like a bug to me.

RDoc appears to produce duplicate method anchors when a class defines an explicit singleton new method and also defines initialize.

```ruby
class DuplicateNewExample
  def self.new(*args, **options, &block)
    super
  end

  def initialize(name)
  end
end

class AliasThenNewExample
  class << self
    alias_method :create, :new

    def new(name)
    end
  end

  def initialize(name, extra = nil)
  end
end
```

Run:
```bash
rdoc duplicate_new.rb
grep -n 'id="method-c-new"' doc/DuplicateNewExample.html
grep -n 'id="method-c-new"' doc/AliasThenNewExample.html
```

This results in RDoc emitting two id="method-c-new" anchors for each class.

In the "wild" you can find DeprecatedConstantProxy having two "new" methods.
https://api.rubyonrails.org/v8.1.3/classes/ActiveSupport/Deprecation/DeprecatedConstantProxy.html

Or ActiveSupport::TimeZone having two different new methods
https://api.rubyonrails.org/v7.0.0/classes/ActiveSupport/TimeZone.html#method-c-new

## Expected result
I'm not 100% sure how to resolve this. 

Maybe it's not a good idea to rename #initialize into #new?

tompng:
Fixed by #1802

PR #1802, opened 2026-08-30
Rename `#initialize` to `::new` before registering it to the container

tompng:
## Background

The Ruby parser renames an instance method `initialize` to `::new` for documentation purposes. This rename happened *after* `container.add_method`, with a comment claiming the ordering is intentional: "Rename after add_method to register duplicated 'new' and 'initialize' defined in c and ruby".

The actual reason for this placement is older. In the Ripper-based streaming parser, documentation modifiers such as `:notnew:` were read *after* the method line, so at `add_method` time the parser simply did not know yet whether the method should be renamed:

```ruby
# Having now read the method parameters and documentation modifiers, we
# now know whether we have to rename #initialize to ::new
```
(lib/rdoc/parser/ripper_ruby.rb, removed in #1690)

The Prism parser processes directives and modifier lines before `add_method`, so this constraint is gone. The post-add placement was a consequence of the streaming parser's information ordering, not a design goal — which is why it is safe to retire the post-add mutation pattern now. The comment in the current code was a port-time rationalization of an observable side effect.

## What the old placement actually did

Registering the method under the `#initialize` key and renaming it afterwards had two effects:

1. `Context#methods_hash` was left keyed by a stale name (`#initialize` pointing at a method now called `::new`). This is one of the obstacles to making `Context#find_method` hash-based (see the discussion in #1796).
2. Duplicate detection in `Context#add_method` was bypassed. A class documenting both `::new` (an explicit `def self.new`, or a C-defined `new`) and `#initialize` ended up with two `::new` entries on its page.

## Change

Move the rename block before `container.add_method`. All information it uses (the method name, `singleton`, and `dont_rename_initialize` set by the `:notnew:` directive) is already available at that point.

With the rename in place before registration, the normal deduplication applies: the first registration wins, and the duplicate is reported by the existing "Duplicate method" warning (visible with `--verbose`).

## Corpus diff

Compared per-class method lists (name, singleton, visibility, file) for whole corpora, before vs after:

| Corpus | Changes |
| --- | --- |
| ruby/ruby | 4 classes lose a duplicated `::new` entry: `Gem::Package::TarReader`, `Gem::Package::TarWriter`, `Gem::Resolver::APISpecification`, `JSON::Ext::Generator::State` |
| activesupport 8.1.3 | 1 class: `ActiveSupport::Deprecation::DeprecatedConstantProxy` |
| rdoc itself | no change |

All other entries are identical. Each changed class defines both `def self.new` and `def initialize` (or, for `JSON::Ext::Generator::State`, a C-defined `new` and a Ruby `initialize` — the exact "c and ruby" case the old comment referred to). On master these pages show `new` twice, with a duplicated `id="method-c-new"` anchor (invalid HTML; the index can only link to the first entry); with this change they show it once.

## Cleanup

The second commit removes the `dont_rename_initialize` keyword argument of `internal_add_method`. It has never been passed a truthy value since its introduction: one call site passes an explicit `false` and the other relies on the `false` default. It is unrelated to `AnyMethod#dont_rename_initialize` (set by the `:notnew:` directive), which remains the live mechanism; removing the constant-false parameter also removes the confusion of two same-named flags in one method.

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -451,7 +451,6 @@ class RDoc::Parser::Ruby < RDoc::Parser
         @container,
         comment: comment,
         directives: directives,
-        dont_rename_initialize: false,
         line_no: line_no,
         visibility: visibility,
         singleton: @singleton || singleton_method,
@@ -722,7 +721,7 @@ class RDoc::Parser::Ruby < RDoc::Parser
     )
   end
 
-  private def internal_add_method(method_name, container, comment:, dont_rename_initialize: false, directives:, modifier_comment_lines: nil, line_no:, visibility:, singleton:, params:, calls_super:, block_params:, tokens:, type_signature_lines: nil) # :nodoc:
+  private def internal_add_method(method_name, container, comment:, directives:, modifier_comment_lines: nil, line_no:, visibility:, singleton:, params:, calls_super:, block_params:, tokens:, type_signature_lines: nil) # :nodoc:
     meth = RDoc::AnyMethod.new(method_name, singleton: singleton)
     meth.comment = comment
     handle_code_object_directives(meth, directives) if directives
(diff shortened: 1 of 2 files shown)


st0012 (approved):

skatkov:
It would be awesome to add a test containing both `def self.new` and `def initialize`, asserting one `::new` entry.

skatkov:
This solves #1735?

PR #1803, opened 2026-08-31
Less strict parsing of visibility method arguments

nevans:
NOTE: this PR stacks on #1793.  Only [the last three commits](https://github.com/ruby/rdoc/pull/1803/changes/f4607d26d8db4ca8d96b819028bf083e8ad5f5c4..242a1e9a28d2c5abdc261e7141bd26ba2999e202) are specific to this PR.

This updates method name argument parsing for the visibility methods:
* `private`, `public`, `protected`
* `private_class_method`, `public_class_method`
* `private_constant`, `public_constant`
* `module_function`

Prior to this PR, the parser is stricter about visibility method name arguments than it should be: it only parses method names when _all_ arguments are symbols.

This updates the prism parser to behave more like the rdoc 7.2, so every (non-interpolated) string and symbol in the argument list is used (`call_node_name_arguments` is used to parse the arguments list).

Additionally:
* Fix `Parser::Ruby` docs for `private_class_method` and `public_class_method`.
  There aren't any `private_class_function` and `public_class_function` methods.  😉
* Refactor `Parser::Ruby#call_node_name_arguments` into plural and singular versions.
  Since meta-programmed method names are only ever inferred from the first argument, the singular version only reads the first argument.  The plural version is now free to use `filter_map` and convert `[]` to `nil`, enabling the short-circuiting pattern used in the visitor.
  Arguably, these methods belong more to the visitor than the "scanner".  Since they simply process prism nodes without any ivar references, they should probably be converted into module functions on a utility module.  But I'm considering that out of scope for this PR.

### DIFF
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -388,15 +388,23 @@ module RDoc
       end
 
       def call_node_name_arguments(call_node) # :nodoc:
-        return [] unless call_node.arguments
-        call_node.arguments.arguments.map do |arg|
-          case arg
-          when Prism::SymbolNode
-            arg.value
-          when Prism::StringNode
-            arg.unescaped
-          end
-        end || []
+        return unless arguments_node = call_node.arguments
+        names = arguments_node.arguments.filter_map { |arg| argument_name(arg) }
+        names unless names.empty?
+      end
+
+      def call_node_name_argument(call_node) # :nodoc:
+        return unless call_node.arguments
+        argument_name(call_node.arguments.arguments.first)
+      end
+
+      def argument_name(argument_node) # :nodoc:
+        case argument_node
+        when Prism::SymbolNode
+          argument_node.value
+        when Prism::StringNode
+          argument_node.unescaped
+        end
       end
 
       # Handles meta method comments
@@ -412,7 +420,7 @@ module RDoc
           case directive
           when 'attr', 'attr_reader', 'attr_writer', 'attr_accessor'
             attributes = [param] if param
-            attributes ||= call_node_name_arguments(node).compact if is_call_node
+            attributes ||= call_node_name_arguments(node) if is_call_node
             rw = directive == 'attr_writer' ? 'W' : directive == 'attr_accessor' ? 'RW' : 'R'
           when 'method'
             method_name = param if param
@@ -438,7 +446,7 @@ module RDoc
             mark_container_documentable(@container)
           end
         elsif line_no || node
-          method_name ||= call_node_name_arguments(node).first if is_call_node
+          method_name ||= call_node_name_argument(node) if is_call_node
           line_no = node.location.start_line if node
           internal_add_method(
             method_name,
@@ -1216,8 +1224,7 @@ module RDoc
         end
 
         def call_node_name_arguments(call_node)
-          names = @scanner.call_node_name_arguments(call_node).compact
-          names unless names.empty?
+          @scanner.call_node_name_arguments(call_node)
         end
 
         def symbol_arguments(call_node)
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -420,7 +420,7 @@ module RDoc
           case directive
           when 'attr', 'attr_reader', 'attr_writer', 'attr_accessor'
             attributes = [param] if param
-            attributes ||= call_node_name_arguments(node) if is_call_node
+            attributes ||= call_node_name_arguments(node) || [] if is_call_node
             rw = directive == 'attr_writer' ? 'W' : directive == 'attr_accessor' ? 'RW' : 'R'
           when 'method'
             method_name = param if param
--- a/lib/rdoc/parser/ruby.rb
+++ b/lib/rdoc/parser/ruby.rb
@@ -18,7 +18,7 @@ module RDoc
     # * constants
     # * aliases
     # * private, public, protected
-    # * private_class_function, public_class_function
+    # * private_class_method, public_class_method
     # * private_constant, public_constant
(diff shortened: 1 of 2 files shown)


nevans:
NOTE: This PR stacks on #1793.  But if I make that the PR's branch the base branch, GitHub will convert this PR into a PR on _my_ repo.  Only [the last three commits](https://github.com/ruby/rdoc/pull/1803/changes/f4607d26d8db4ca8d96b819028bf083e8ad5f5c4..242a1e9a28d2c5abdc261e7141bd26ba2999e202) are specific to this PR.

lib/rdoc/parser/ruby.rb
```diff
@@ -1334,20 +1345,20 @@ def _visit_call_extend(call_node)
 
     def _visit_call_public_constant(call_node)
       return if @scanner.in_proc_block || @scanner.singleton
-      names = symbol_arguments(call_node)
-      @scanner.container.set_constant_visibility_for(names.map(&:to_s), :public) if names
+      return unless names = call_node_name_arguments(call_node)
+      @scanner.container.set_constant_visibility_for(names, :public)
```
st0012, line 1349: nit: I think the original condition shape can be kept and is clearer

```suggestion
      names = call_node_name_arguments(call_node)
      @scanner.container.set_constant_visibility_for(names, :public) if names
```

lib/rdoc/parser/ruby.rb
```diff
@@ -1334,20 +1345,20 @@ def _visit_call_extend(call_node)
...
     end
 
     def _visit_call_private_constant(call_node)
       return if @scanner.in_proc_block || @scanner.singleton
-      names = symbol_arguments(call_node)
-      @scanner.container.set_constant_visibility_for(names.map(&:to_s), :private) if names
+      return unless names = call_node_name_arguments(call_node)
+      @scanner.container.set_constant_visibility_for(names, :private)
```
st0012, line 1355: Same as above

lib/rdoc/parser/ruby.rb
```diff
@@ -1334,20 +1345,20 @@ def _visit_call_extend(call_node)
...
     end
 
     def _visit_call_attr_reader_writer_accessor(call_node, rw)
       return if @scanner.in_proc_block
-      names = symbol_arguments(call_node)
-      @scanner.add_attributes(names.map(&:to_s), rw, call_node.location.start_line) if names
+      return unless names = call_node_name_arguments(call_node)
+      @scanner.add_attributes(names, rw, call_node.location.start_line)
```
st0012, line 1361: Same as above

st0012 (approved):
LGTM, just some nits
Thanks for fixing it 👍

nevans:
@st0012 I only just got back around to addressing your nits.  And I rebased it after #1795, which made every line a conflict.  But the changes here were small and simple to replay.

lib/rdoc/parser/ruby.rb
nevans: I disagree that it's clearer here, but that's somewhat subjective and a very tiny difference... when the lines aren't too long.

lib/rdoc/parser/ruby.rb
```diff
@@ -388,15 +388,23 @@ def handle_modifier_directive(code_object, line_no) # :nodoc:
...
-          case arg
-          when Prism::SymbolNode
-            arg.value
-          when Prism::StringNode
-            arg.unescaped
-          end
-        end || []
+        return unless arguments_node = call_node.arguments
+        names = arguments_node.arguments.filter_map { |arg| argument_name(arg) }
+        names unless names.empty?
```
tompng, line 393: Return value was always an array.
It'll be changed to nil-able, so code around L415 (updated in #1793) needs change.
```ruby
# Around L415
when 'attr', 'attr_reader', 'attr_writer', 'attr_accessor'
  # attributes ||= call_node_name_arguments(node).compact if is_call_node
  # ↓
  attributes ||= call_node_name_arguments(node) || [] if is_call_node
  ...
end

if attributes
  # handles attr, attr_reader, attr_writer, attr_accessor (so attributes must be non-nil)
elsif line_no || node
  # handles non-attr ghost method
end
```

nevans: While it is true that it was previously always an array:
* it was only _used_ inside an `if attributes` clause
* the `elsif line_no || node` clause is triggered if and only if there's _also_ a `method` or `singleton-method` directive.

Although it's a weird edge case (which explains why there's no tests for it), allowing `method` or `singleton-method` to trigger when `attributes` failed to collect any names seems like maybe a good thing to me?  Anyway, that was my reasoning when I made the change.  Is there some way this could cause an issue?

Mostly I changed it for the consistency with the similar `Visitor` methods (`constant_arguments_names`, `symbol_arguments`, `visibility_method_arguments`, etc).  While `call_node_name_arguments` lives on the scanner, like those visitor methods it processes an AST node and doesn't reference any ivars.  It feels to me like it ought to be a module_function in a shared module together with those other methods, so I figured it made sense to be consistent with them.

But I'm happy to change it back to always returning an array: unless there's some important distinction between empty vs non-existent, I generally prefer to return an empty collection rather than `nil`.  It just feels odd for it to behave differently from the other related methods.

nevans: Nope, I reasoned incorrectly.  While _`line_no`_ is only set when those other directives are visible, `node` could be there for any `##` comments that are matched to a subsequent node.

nevans: @tompng Thanks for catching that.  I've added some examples to `test_invalid_meta_method` that capture the bug I'd created.

tompng (approved):
:+1:

PR #1804, opened 2026-08-31
Use one responsive search surface in Aliki

st0012:
Aliki had two separate search problems on compact layouts:

- Pressing `/` focused the hidden desktop field instead of opening the visible search surface.
- Search result cards looked clickable, but only the title text was a link. Clicking the badge, snippet, or card padding did nothing.

The first problem came from rendering separate desktop and compact search forms. Aliki now renders one search form and one result list. CSS presents the same markup inline on larger screens and as an overlay on smaller screens, so the search logic no longer needs separate paths.

Each result anchor now covers its full card. This makes the entire result clickable while preserving standard link actions.

https://github.com/user-attachments/assets/18355b04-b734-4efe-a736-66019d965b99

### DIFF
--- a/lib/rdoc/generator/template/aliki/js/aliki.js
+++ b/lib/rdoc/generator/template/aliki/js/aliki.js
@@ -47,6 +47,8 @@ function createSearchInstance(input, result) {
       html += `<span class="search-type search-type-${this.escapeHTML(typeClass)}">${typeLabel}</span>`;
     }
 
+    html += '</p>';
+
     if (result.snippet)
       html += `<div class="search-snippet">${result.snippet}</div>`;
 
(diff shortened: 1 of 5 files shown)


st0012:
@codex review

test/rdoc/generator/aliki_test.rb
```diff
@@ -105,6 +105,16 @@ def test_asset_version_query_strings
     assert_match %r{js/theme-toggle\.js\?v=#{Regexp.escape(RDoc::VERSION)}}, content
   end
 
+  def test_generated_header_uses_one_search_surface
```
st0012: This is redundant. Remove it

lib/rdoc/generator/template/aliki/js/aliki.js
```diff
@@ -47,6 +47,8 @@ function createSearchInstance(input, result) {
       html += `<span class="search-type search-type-${this.escapeHTML(typeClass)}">${typeLabel}</span>`;
     }
 
+    html += '</p>';
```
st0012, line 50: This is an unrelated fix that doesn't affect user behaviour.

tompng (approved):
Looks nice!

PR #1805, opened 2026-08-31
Keep selected Aliki search result visible

st0012:
On compact layouts, the search body owned scrolling while keyboard navigation continued to scroll the result list. Arrow-key selection moved, but the selected row could leave the visible panel.

Keep the result list as the scroll container in the compact layout, matching desktop. The existing search navigation now keeps the selected row visible without breakpoint-specific JavaScript or moving the page.

### DIFF
--- a/lib/rdoc/generator/template/aliki/css/rdoc.css
+++ b/lib/rdoc/generator/template/aliki/css/rdoc.css
@@ -2116,24 +2116,29 @@ header.top-navbar #search-results[aria-expanded="false"] {
   }
 
   header.top-navbar .navbar-search.is-open .search-body {
+    display: flex;
     flex: 1;
-    padding: var(--space-6);
-    overflow-y: auto;
+    flex-direction: column;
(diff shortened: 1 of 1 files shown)


lib/rdoc/generator/template/aliki/css/rdoc.css
```diff
@@ -2116,9 +2116,12 @@ header.top-navbar #search-results[aria-expanded="false"] {
   }
 
   header.top-navbar .navbar-search.is-open .search-body {
+    display: flex;
     flex: 1;
+    flex-direction: column;
+    min-height: 0;
     padding: var(--space-6);
```
tompng: I like the original scrollbar position.
left: before, right: after
Image

How about moving this padding to `#search-results` ?

st0012: Fixed 👍

tompng (approved):
:+1:

PR #1813, opened 2026-09-06
Remove unused method `RDoc::Require#top_level` that always raises NoMethodError

tompng:
It has been broken since 2d884240 (Move global parse state to RDoc::Store) removed TopLevel.all_files_hash and TopLevel.all_files.

Result of `RDoc::Require.new('a','b').top_level` is basically unchanged.
Before: `NoMethodError(undefined method 'all_files_hash')`
After: `NoMethodError(undefined method 'top_level')`

### DIFF
--- a/lib/rdoc/code_object/require.rb
+++ b/lib/rdoc/code_object/require.rb
@@ -15,8 +15,7 @@ module RDoc
 
     def initialize(name, comment)
       super()
-      @name = name.gsub(/'|"/, "") #'
-      @top_level = nil
+      @name = name.gsub(/'|"/, "")
       self.comment = comment
     end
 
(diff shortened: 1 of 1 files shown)


lib/rdoc/code_object/require.rb
```diff
@@ -16,7 +16,6 @@ class Require < CodeObject
     def initialize(name, comment)
       super()
       @name = name.gsub(/'|"/, "") #'
```
st0012: Not related, but does the `#'` here does anything?

st0012 (approved):

lib/rdoc/code_object/require.rb
tompng: Removed.
It's strange. There's two more `#'` comments in `lib/rdoc/context.rb` 🤔

kou: It's for broken syntax highlighting implementations. (e.g. Regular expression based implementation.)

If developers don't use these implementations, we can remove it.

PR #1824, opened 2026-09-17
Cache regexp in `Markup::ToHtml`

Earlopain:
Especially the first one is not good for performance.

In ruby/prism, docs generation goes from 7s to 6.75s for me, which is about a 4% improvement (including bundler startup etc, so for pure rdoc timing it should be even better)


skatkov:
I already have benchmarks developed for my previous work - took this PR for a benchmark spin. 

I could not reproduce improvements in Prism, but Rails docs generation definitely improved (not only in speed, but more notably in memory allocations)

| Variant | Total Time | Peak RSS | Live heap after generation |
| --- | ---: | ---: | ---: |
| Baseline | 22.83s | 415.4 MiB | 1,908,566 |
| PR #1824 | 21.87s | 375.2 MiB | 1,838,027 |
| Change | **-4.2%** | **-9.7%** | **-3.7%** |

It would be lovely if this PR could be merged.

st0012 (approved):
Thanks!

Issue #1823, opened 2026-09-16
module_function without arguments is not applied to subsequent method definitions


thyresias:
RDoc 8 appears not to handle correctly the no-argument form of `Module#module_function`.

In Ruby, this:

```ruby
module M
  module_function

  def foo
  end
end
```

defines both:

- `M.foo`, as a public singleton method;
- `M#foo`, as a private instance method.

However, RDoc appears to document `foo` only as an instance method and does not create/document the corresponding module method `M.foo`.

After generating the documentation, I would expect `M.foo` to appear as a module method.

It does not.

### Real-world example

This occurs in `json` 3.0.2.

In `lib/json/common.rb`, `JSON.parse` is defined using the no-argument form of `module_function`:

```ruby
module JSON
  # ...

  module_function

  # ...

  def parse(source, on_load: nil, object_class: nil, array_class: nil, **options)
    # ...
  end
end
```

Ruby therefore exposes this method as:

```ruby
JSON.parse(...)
```

but RDoc 8 does not appear to document `JSON.parse`.

### Likely cause

RDoc already seems to handle the explicit form:

```ruby
module_function :foo
```

by converting an existing method to a module function.

The no-argument form has different semantics: it changes the state of the module so that methods defined *subsequently* become module functions.

RDoc therefore needs to remember that `module_function` was called without arguments and apply that state to later `def` nodes, until the relevant scope ends or the state is otherwise changed.

Conceptually:

```ruby
module_function

def foo
end
```

should result in RDoc recording both:

```ruby
# private instance method
M#foo

# public singleton/module method
M.foo
```

### Notes

This does not appear to be a Prism parsing issue itself: the syntax is parsed normally. The missing part seems to be in RDoc's tracking of the semantic state introduced by a no-argument `module_function`.

I encountered this while generating documentation for `json` 3.0.2 with RDoc 8.

skatkov:
By default, rdoc documents only `public` and `protected` methods. Did you try changing visibility?

`rdoc --visibility=private`?

thyresias:
> By default, rdoc documents only `public` and `protected` methods. Did you try changing visibility?
> 
> `rdoc --visibility=private`?

The module method is public. And it is not documented.

tompng:
Currently, only `module_function` with argument is supported.
```ruby
module_function :name1, :name2
```
`module_function` without argument is not supported nor implemented yet.
I think it wasn't implemented in the old ripper-based parser too.

[...]

PR #1825, opened 2026-09-20
Module function without argument

thyresias:
This supports module_function without argument (#1823).
I had initially the intention to use a special value for `@visibibility`, but I think it is better that it has actual visibility values, so I introduced `@module_function_mode`.


lib/rdoc/parser/ruby.rb
```diff
@@ -1276,6 +1282,11 @@ def _visit_call_require(call_node)
         end
 
         def _visit_call_module_function(call_node)
+          if !call_node.arguments || call_node.arguments.arguments.empty?
+            @scanner.visibility = :private
```
st0012, line 1286: This condition doesn't check `@scanner.in_proc_block` and it means in cases like

```rb
module M
  Module.new do
    module_function
  end

  def outer; end
end
```

The private visibility leaks outside of the anonymous module.

test/rdoc/parser/ruby_test.rb
```diff
@@ -1403,6 +1403,32 @@ def m3; end
...
+    instance_methods = mod.method_list.reject(&:singleton)
+    singleton_methods = mod.method_list.select(&:singleton)
+    assert_equal ['a1', 'a2'], attributes.map(&:name)
+    assert_equal [:public, :private], attributes.map(&:visibility)
+    assert_equal ['m1', 'm2', 'm3', 'm4'], instance_methods.map(&:name)
+    assert_equal [:private, :private, :public, :private], instance_methods.map(&:visibility)
+    assert_equal ['m1', 'm2'], singleton_methods.map(&:name)
+    assert_equal [:public, :public], singleton_methods.map(&:visibility)
+  end
+
```
st0012, line 1431: Let's also add this test:

```rb
def test_module_function_no_arg_does_not_leak_from_block
  util_parser <<~RUBY
    module M
      Module.new do
        module_function
      end
      def outer; end
    end
  RUBY

  mod = @store.find_module_named 'M'
  methods = mod.method_list.map { |method| [method.name, method.singleton, method.visibility] }
  assert_equal [['outer', false, :public]], methods
end
```

lib/rdoc/parser/ruby.rb
```diff
@@ -133,7 +133,7 @@ class Ruby < Parser
       # Matches an RBS inline type annotation line: #: followed by whitespace
       RBS_SIG_LINE = /\A#:\s/ # :nodoc:
 
-      attr_accessor :visibility
+      attr_accessor :visibility, :module_function_mode
```
tompng, line 136: Visibility and module_function_mode are exclusive.
We don't want to introduce instance values with complex dependencies. Simple internal representation is better.
How about changing `attr_accessor :visibility` to `attr_accessor :visibility_mode`
and define `def visibility` that computes actual visibility from exclusive state like this?

```ruby
attr_accessor :visibility_mode   # :public, :private, :protected, :module_function

def visibility
  @visibility_mode == :module_function ? :private : @visibility_mode
end
```

lib/rdoc/parser/ruby.rb
```diff
@@ -1159,6 +1163,7 @@ def visit_def_node(node)
             return unless receiver_name
           when nil
             visibility = @scanner.visibility
+            mod_function = @scanner.module_function_mode && !singleton
```
tompng, line 1166: singleton is always `nil` because the local variable is not assigned yet.
I think it's a mistake of `@scanner.singleton`, but `module_function_mode` is only set on non-singleton module scope. Simply removing `&& (singleton_check_expression)` is enough.

tompng: I think it's better to check singleton in `_visit_call_module_function`.
```ruby
module A
  class << A
    module_function rescue p($!) # NameError, RDoc should just ignore it
    def f; end # public method
  end
end
```

PR #1829, opened 2026-09-21
Fix stack overflow in MethodAttr#documented? on cyclic includes

### LINKED ISSUE #1088: Including modules creates an infinite loop between methods
Hi! I think RDoc sometimes wrongly infers module ancestry and then methods can have cyclical `see` references.

I encountered this error in the docs for GrpahQL-Ruby (fix: https://github.com/rmosolgo/graphql-ruby/pull/4819) and I created this script to replicate it: 

```ruby
# rdoc_bug.rb
module DoSomething
  def self.generate_module
    Module.new do
      include DoSomethingExtra
    end
  end

  def do_something
    1 + 1
  end

  module DoSomethingExtra
    include DoSomething

    def do_something
      super + 1
    end
  end
end
```

For me, that script ends in "stack level too deep": 

```
$ rdoc rdoc_bug.rb
Parsing sources...
100% [ 1/ 1]  rdoc_bug.rb

Generating Darkfish format into /Users/rmosolgo/code/graphql-ruby/doc...
uh-oh! RDoc had a problem:
stack level too deep

```

I was able to work around it, so this bug is not blocking me, but I thought I'd open it here in case its helpful to anyone else.

SulimanAbdulrazzaq:
`MethodAttr#documented?` also treats a method as documented when its alias or its `#see` method is documented, and `#see` is looked up in the ancestors of the method's owner. With the reproduction from https://github.com/ruby/rdoc/issues/1088#issuecomment-3863703871 (two modules that include each other), `A#m.see` is `B#m` and `B#m.see` is `A#m`, so `documented?` recursed until rdoc stopped with "stack level too deep". A superclass cycle, which #1785 made `ancestors` handle, ends in the same loop.

The guard stays inside `documented?`: a method or attribute that is reached again while it is still being checked counts as undocumented, since the cycle adds no documentation, and every other alias and `#see` path is still followed. The flag is only set around the check and is reset in an `ensure`, so a later call gives the same answer.

The new tests in `method_attr_test.rb` cover cyclic includes with and without a documented method, and a superclass cycle.

Fixes #1088


SulimanAbdulrazzaq:
@st0012 when you have a moment, could you approve the workflow runs and take a look? With two modules that include each other (the reproduction from #1088), `MethodAttr#documented?` follows `#see` back and forth until rdoc stops with "stack level too deep"; this adds a cycle guard. The workflows are waiting for approval, and the branch has no conflicts.

